PR #32, opened 2014-06-02
Window.URL

zapletnev:


swhgoon:
use `js.Dynamic.global.URL.asInstanceOf[dom.URL]` now???

sjrd:
@swhgoon Hum yes, something like that.

PR #59, opened 2014-10-13
Add touch events

tindzk:


tindzk:
@lihaoyi I'm not convinced of the ``chromefirefox`` and ``internetexplorer`` sub-packages. We'd also need to keep track of the version numbers to make it actually useful. I hardly doubt it's worth the effort.

@sjrd I changed my commit. Is it how you envisioned the implicits?

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -245,3 +248,50 @@ sealed class Storage(domStorage: dom.Storage) {
...
+  var ontouchcancel: js.Function1[raw.TouchEvent, _] = js.native
+}
+
+object TouchEvents {
+  implicit def HTMLDocumentToTouchEvents(html: raw.HTMLDocument): TouchEvents =
+    html.asInstanceOf[TouchEvents]
+
+  implicit def WindowToTouchEvents(window: raw.Window): TouchEvents =
+    window.asInstanceOf[TouchEvents]
+}
```
sjrd: EOL at EOF.

sjrd:
That's all. And yes, that's exactly what I meant about the implicits.

tindzk:
Added the EOL.

lihaoyi:
@tindzk yeah i'm not wedded to the idea I just thought it would be nice to be able to keep track of who depends on what browser version. Currently the rule has basically been "must be present version in latest version of IE/FF/Chrome", which has been nice for me because I mostly never need to worry about compatibility

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -245,3 +248,51 @@ sealed class Storage(domStorage: dom.Storage) {
...
+
```
sjrd: Ah ah. Now You have an empty line, which is 2 eols instead of 1 ... :-s

tindzk:
I see. Using traits for experimental features seems like a good trade-off to me because this makes it explicit to the developer that a feature must be taken with a grain of salt.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
tindzk: This happens when using different editors to edit the same file... Should be fixed now.

lihaoyi:
Yeah. That object-with-implicits looks like any object-with-implicits though; could we add a doc-string or annotation saying "this is not as cross platform as the other things and the implicits enable cross-platform stuff?"?

tindzk:
Sure. See updated commit.

lihaoyi:
Cool, looks good to me

sjrd:
LGTM

PR #61, opened 2014-10-29
Keyboard polyfill

l15k4:
Based on [this] (https://groups.google.com/forum/#!topic/scala-js/5sWDOyu-OJM)  conversation I'd like to discuss how to include this PR into scala-js-dom... It starts to be a little messy in there and we should think about a system that would allow for adding new extensions, polyfills and other cool things more easily... Meaning that it would be clear how and where to add it.

For instance here the `object KeyboardPolyfill` is a polyfill thingy, the `KCode` is a keyboard thingy and `ChCode` rather Character thingy :-) Anyway all this stuff is quite essential if you want to do something really keyboard interactive...

The polyfill's `normalize` method will probably `grow` as people will notice differences on different devices/browsers, which is mainly mobile devices and ios.

### DIFF
--- a/src/main/scala/org/scalajs/dom/extensions/Keyboard.scala
+++ b/src/main/scala/org/scalajs/dom/extensions/Keyboard.scala
@@ -1,0 +1,274 @@
+package org.scalajs.dom.extensions
+
+import org.scalajs.dom
+import org.scalajs.dom.KeyboardEvent
+
+import scala.scalajs.js
+import scala.scalajs.js.UndefOr
+
+object KeyboardPolyfill {
+
+  /**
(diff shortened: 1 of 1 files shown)


Voltir:
I would use this - I am using this, but I just duplicated the file and am using it locally.

l15k4:
I made it available as a library [scalajs-keyboard-polyfil](https://github.com/viagraphs/scalajs-keyboard-polyfill). Note it is not released or published yet until scalajs 6.0.0 is out.

mseddon:
@sjrd we can close this one- scala-js-dom does not implement polyfills.

l15k4:
Sure, oss library is more suitable here...

PR #83, opened 2015-01-04
IDBDatabase#version should be typed as Integer

l15k4:
Hi, from what I observed this is the correct typing for `IDBDatabase#version`. 

I haven't tested whether firefox still returns empty string when database is first created. It is not recommended though because within version upgrade callback one should be calling only `IDBVersionChangeEvent#newVersion` and `IDBVersionChangeEvent#oldVersion`.

Anyway webkits always return Integer, imho they have a fallback to `IDBVersionChangeEvent#newVersion` whereas firefox has/had fallback to empty string.

Either way `String` type doesn't make sense and I think that using `js.Any` just because Firefox might fall back to empty string is wrong. User of the API will know from scaladoc that he should use  `IDBVersionChangeEvent#newVersion` instead in that case.

### DIFF
--- a/src/main/scala/org/scalajs/dom/Idb.scala
+++ b/src/main/scala/org/scalajs/dom/Idb.scala
@@ -544,12 +544,13 @@ object IDBTransaction extends js.Object {
  */
 class IDBDatabase extends EventTarget {
   /**
-   * A 64-bit integer that contains the version of the connected database. When a
-   * database is first created, this attribute is the empty string.
-   *
-   * MDN
+   * A 64-bit integer that contains the version of the connected database.
+   * When a database is first created or upgraded you should use
+   * [[org.scalajs.dom.IDBVersionChangeEvent#newVersion]] instead.
+   * Webkit returns always integer and the value is 1 when
+   * database is first created.
    */
-  def version: String = js.native
+  def version: Int = js.native
 
   /**
    * A DOMString that contains the name of the connected database.


sjrd:
What's the possible range of that value? Can it exceed the range of an `Int`? I'm concerned about the bit saying 64-bit integer.

l15k4:
I don't think it may exceed it, IndexedDb version upgrades serve for schema changes/versioning (adding a new store or new store index + a few other things, not much can be done) so I can't imagine a use case where db version would exceed number 1000. I can imagine a service that would create a new document store for some application entity, but imho it would be rather misuse if it exceeded numbers like 1000...  Imho the overall ~ value is like 1 - 20 :-)

sjrd:
It's there some guarantee that it always starts at 1 and then increases? It could be an arbitrary value. In that case, even if only 1000 different values are used, some of them could be too big.

l15k4:
Afaik there used to be `setVersion` method that has been [deprecated] (http://stackoverflow.com/a/9542393/306488), you can't even "find" it in the [spec](http://www.w3.org/TR/IndexedDB/) or on MDN now. And imho they forced the incremental way only 1,2,3,4,5,... So that now version increases in onversionupgrade event/callback only and it starts at value 1. And everytime when onversionupgrade occurs there is `IDBVersionChangeEvent` available as a callback parameter that contains those preferred `newVersion` and `oldVersion` values.

sjrd:
OK LGTM

PR #85, opened 2015-01-07
Reorganize the org.scalajs.dom namespace into something saner

lihaoyi:
- The raw "everything goes" namespace is now org.scalajs.dom.raw
- org.scalajs.dom.{css, html, idb, svg, webgl} namespaces contain un-prefixed versions of the things that start with their prefix. The `Element` suffix has also been dropped where possible
- org.scalajs.dom.extensions has been renamed to org.scalajs.dom.ext
- org.scalajs.dom itself contains all things in org.scalajs.dom.raw that aren't in any of the un-prefixed namespaces

Overall, this should accomplish a few things:

- Massively clean up org.scalajs.dom.*, since we've moved ~270 names into proper namespaces
- Make dealing with these names much easier, e.g. now `dom.HTMLTableCellElement` is just `html.TableCell`
- Make a clear separation between "all un-modified dom facades" and "re-organized nice-to-use dom facades", so hopefully there won't be any confusion over things missing or where things should go
- Reasonable backwards compat just by changing imports: org.scalajs.dom -> org.scalajs.dom.raw and things should keep working

Some Caveats:

- The mapping isn't *quite* perfect: we still have `svg.RectElement` instead of just `svg.Rect`, sinct `svg.Rect` is taken by something else
- You don't get doc-dropdown in IntelliJ because of the aliases. I assume this is something we can bug the IntelliJ people to fix since it affects lots of other things (e.g. the std lib) and not just Scala.js

### DIFF
--- a/example/src/main/scala/example/Example.scala
+++ b/example/src/main/scala/example/Example.scala
@@ -1,9 +1,7 @@
 package example
 
-import scala.scalajs.js.JSApp
-
 import org.scalajs.dom
-
+import dom.html
 import scala.scalajs.js.annotation.JSExport
 
 @JSExport
@@ -18,7 +16,7 @@ object Alert {
 @JSExport
 object NodeAppendChild {
   @JSExport
-  def main(div: dom.HTMLDivElement) = {
+  def main(div: html.Div) = {
     val child = dom.document
                    .createElement("div")
 
@@ -32,7 +30,7 @@ object NodeAppendChild {
 @JSExport
 object ElementStyle {
   @JSExport
-  def main(div: dom.HTMLDivElement) = {
+  def main(div: html.Div) = {
     val colors = Seq(
       "red", "green", "blue"
     )
@@ -47,8 +45,7 @@ object ElementStyle {
 @JSExport
 object LocalStorage {
   @JSExport
-  def main(in: dom.HTMLInputElement,
-           box: dom.HTMLDivElement) = {
+  def main(in: html.Input, box: html.Div) = {
     val key = "my-key"
 
     in.value =
@@ -67,7 +64,7 @@ object LocalStorage {
 @JSExport
 object Canvas {
   @JSExport
-  def main(c: dom.HTMLCanvasElement) = {
+  def main(c: html.Canvas) = {
     type Ctx2D =
       dom.CanvasRenderingContext2D
     val ctx = c.getContext("2d")
@@ -93,8 +90,8 @@ object Canvas {
 @JSExport
 object Base64 {
   @JSExport
-  def main(in: dom.HTMLInputElement,
-           out: dom.HTMLDivElement) = {
+  def main(in: html.Input,
+           out: html.Div) = {
     in.onkeyup = { (e: dom.Event) =>
       out.textContent =
         dom.btoa(in.value)
@@ -105,7 +102,7 @@ object Base64 {
 @JSExport
 object EventHandler{
   @JSExport
-  def main(pre: dom.HTMLPreElement) = {
+  def main(pre: html.Pre) = {
     pre.onmousemove = {
       (e: dom.MouseEvent) =>
         pre.textContent =
@@ -123,7 +120,7 @@ object EventHandler{
 @JSExport
 object XMLHttpRequest{
   @JSExport
-  def main(pre: dom.HTMLPreElement) = {
+  def main(pre: html.Pre) = {
     val xhr = new dom.XMLHttpRequest()
     xhr.open("GET",
       "http://api.openweathermap.org/" +
@@ -142,8 +139,8 @@ object XMLHttpRequest{
 @JSExport
 object Websocket {
   @JSExport
-  def main(in: dom.HTMLInputElement,
-           pre: dom.HTMLPreElement) = {
+  def main(in: html.Input,
+           pre: html.Pre) = {
     val echo = "ws://echo.websocket.org"
     val socket = new dom.WebSocket(echo)
     socket.onmessage = {
@@ -162,8 +159,8 @@ object Websocket {
 @JSExport
 object AjaxExtension {
   @JSExport
-  def main(pre: dom.HTMLPreElement) = {
-    import dom.extensions.Ajax
+  def main(pre: html.Pre) = {
+    import dom.ext.Ajax
     import scalajs.concurrent
                   .JSExecutionContext
                   .Implicits
--- a/src/main/scala/org/scalajs/dom/css.scala
+++ b/src/main/scala/org/scalajs/dom/css.scala
@@ -1,0 +1,19 @@
+package org.scalajs.dom
+
+/**
+ * Short aliasas of all the dom.CSSThing classes
+ */
+object css {
+  type FontFaceRule = raw.CSSFontFaceRule
+  type ImportRule = raw.CSSImportRule
+  type KeyframeRule = raw.CSSKeyframeRule
+  type MediaRule = raw.CSSMediaRule
+  type NamespaceRule = raw.CSSNamespaceRule
+  type PageRule = raw.CSSPageRule
+  type Rule = raw.CSSRule
+  @inline def Rule = raw.CSSRule
+  type RuleList = raw.CSSRuleList
+  type StyleDeclaration = raw.CSSStyleDeclaration
+  type StyleSheet = raw.CSSStyleSheet
+  type StyleRule = raw.CSSStyleRule
+}
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -1,4 +1,6 @@
-package org.scalajs.dom.extensions
+package org.scalajs.dom.ext
+
+import org.scalajs.dom.html
 
 import scala.scalajs.js
 import org.scalajs.dom
@@ -80,7 +82,7 @@ object Image {
     val img =
       dom.document
         .createElement("img")
-        .asInstanceOf[dom.HTMLImageElement]
+        .asInstanceOf[html.Image]
 
     img.src = "data:image/svg+xml;base64," + s
     img
@@ -199,7 +201,7 @@ object Ajax{
             withCredentials: Boolean, 
             responseType: String): Future[dom.XMLHttpRequest] = {
     val req = new dom.XMLHttpRequest()
-    val promise = Promise[dom.XMLHttpRequest]
+    val promise = Promise[dom.XMLHttpRequest]()
 
     req.onreadystatechange = {(e: dom.Event) =>
       if (req.readyState.toInt == 4){
--- a/src/main/scala/org/scalajs/dom/ext/package.scala
+++ b/src/main/scala/org/scalajs/dom/ext/package.scala
@@ -2,31 +2,29 @@ package org.scalajs.dom
 
 import scala.language.implicitConversions
 
-import scala.scalajs.js
-import org.scalajs.dom
 import scala.collection.mutable
 
-package object extensions {
+package object ext {
 
-  implicit class PimpedNodeList(nodes: dom.NodeList) extends EasySeq[dom.Node](nodes.length, nodes.apply)
+  implicit class PimpedNodeList(nodes: NodeList) extends EasySeq[Node](nodes.length, nodes.apply)
 
-  implicit class PimpedTouchList(nodes: dom.TouchList) extends EasySeq[dom.Touch](nodes.length, nodes.apply)
+  implicit class PimpedTouchList(nodes: TouchList) extends EasySeq[Touch](nodes.length, nodes.apply)
 
-  implicit class PimpedHtmlCollection(coll: dom.HTMLCollection) extends EasySeq[dom.Element](coll.length, coll.apply)
+  implicit class PimpedHtmlCollection(coll: html.Collection) extends EasySeq[Element](coll.length, coll.apply)
 
-  implicit class PimpedSVGTransformList(coll: dom.SVGTransformList) extends EasySeq[dom.SVGTransform](coll.numberOfItems, coll.getItem)
+  implicit class PimpedSVGTransformList(coll: svg.TransformList) extends EasySeq[svg.Transform](coll.numberOfItems, coll.getItem)
 
   implicit class Castable(x: Any) {
     def cast[T] = x.asInstanceOf[T]
   }
 
-  implicit def pimpAnimatedNumber(x: dom.SVGAnimatedNumber) = x.baseVal
+  implicit def pimpAnimatedNumber(x: svg.AnimatedNumber) = x.baseVal
 
-  implicit def pimpRichAnimatedNumber(x: dom.SVGAnimatedNumber) = x.baseVal: runtime.RichDouble
+  implicit def pimpRichAnimatedNumber(x: svg.AnimatedNumber) = x.baseVal: runtime.RichDouble
 
-  implicit def pimpAnimatedLength(x: dom.SVGAnimatedLength) = x.baseVal.value
+  implicit def pimpAnimatedLength(x: svg.AnimatedLength) = x.baseVal.value
 
-  implicit def pimpRichAnimatedLength(x: dom.SVGAnimatedLength) = x.baseVal.value: runtime.RichDouble
+  implicit def pimpRichAnimatedLength(x: svg.AnimatedLength) = x.baseVal.value: runtime.RichDouble
 
   implicit def color2String(c: Color) = c.toString
   implicit class pimpedContext(val ctx: CanvasRenderingContext2D) {
@@ -69,7 +67,6 @@ package object extensions {
     }
   }
 
-
   /**
    * Implicit class to deal with attributes as with normal mutable Map
    * @param attributes
@@ -80,7 +77,6 @@ package object extensions {
     override def iterator: Iterator[(String, Attr)] = new Iterator[(String, Attr)] {
       var index = 0
 
-
       override def next(): (String, Attr) = {
         val n: Attr = attributes.item(index)
         this.index = this.index + 1
@@ -90,14 +86,12 @@ package object extensions {
       override def hasNext: Boolean = index < self.length
     }
 
-
     override def get(key: String): Option[Attr] = attributes.getNamedItem(key) match {
       case null => None
       case attr => Some(attr)
     }
 
     def length: Int = attributes.length.toInt
-
 
     override def -=(key: String) = {
       attributes.removeNamedItem(key)
@@ -108,6 +102,4 @@ package object extensions {
       attributes.setNamedItem(kv._2)
       this}
   }
-
-
 }
--- a/src/main/scala/org/scalajs/dom/html.scala
+++ b/src/main/scala/org/scalajs/dom/html.scala
@@ -1,0 +1,85 @@
+package org.scalajs.dom
+
+/**
+ * Short aliasas of all the dom.HTMLThing classes
+ */
+object html {
+  type Anchor = raw.HTMLAnchorElement
+  type Applet = raw.HTMLAppletElement
+  type Audio = raw.HTMLAudioElement
+  type Area = raw.HTMLAreaElement
+  type AreasCollection = raw.HTMLAreasCollection
+  type Base = raw.HTMLBaseElement
+  type BaseFont = raw.HTMLBaseFontElement
+  type BGSound = raw.HTMLBGSoundElement
+  type BlockElement = raw.HTMLBlockElement
+  type Body = raw.HTMLBodyElement
+  type Button = raw.HTMLButtonElement
+  type BR = raw.HTMLBRElement
+  type Canvas = raw.HTMLCanvasElement
+  type Collection = raw.HTMLCollection
+  type DataList = raw.HTMLDataListElement
+  type DD = raw.HTMLDDElement
+  type Directory = raw.HTMLDirectoryElement
+  type Div = raw.HTMLDivElement
+  type DList = raw.HTMLDListElement
+  type DT = raw.HTMLDTElement
+  type Document = raw.HTMLDocument
+  type Element = raw.HTMLElement
+  type Embed = raw.HTMLEmbedElement
+  type FieldSet = raw.HTMLFieldSetElement
+  type Font = raw.HTMLFontElement
+  type Form = raw.HTMLFormElement
+  type Frame = raw.HTMLFrameElement
+  type FrameSet = raw.HTMLFrameSetElement
+  type Head = raw.HTMLHeadElement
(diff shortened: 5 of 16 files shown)


src/main/scala/org/scalajs/dom/ShortAliases.scala
```diff
@@ -0,0 +1,279 @@
...
+
+object css{
+  type StyleDeclaration = CSSStyleDeclaration
+  type StyleSheet = CSSStyleSheet
+  type StyleRule = CSSStyleRule
+  type MediaRule = CSSMediaRule
+  type NamespaceRule = CSSNamespaceRule
+  type ImportRule = CSSImportRule
+  type Rule = CSSRule
+  val Rule = raw.CSSRule
```
sjrd: I would recommend `@inline def Rule = raw.CSSRule` for these cases instead.
`val` could cause problems if you have more than one in one namespace. If one exists and you access it, but another doesn't in that browser, you'll get a boom, just because the object is entirely initialized to give you the `val` you asked for, but in doing so it also tries to fetch all the other `val`s in that object.

sjrd: An alternative is `lazy val`, but it has a footprint in terms of generated code. The `@inline def` completely disappears after optimizations.

src/main/scala/org/scalajs/dom/ShortAliases.scala
```diff
@@ -0,0 +1,279 @@
...
+  type Form = HTMLFormElement
+}
+object idb{
+  type ObjectStore = IDBObjectStore
+  type VersionChangeEvent = IDBVersionChangeEvent
+  type Index = IDBIndex
+  type Cursor = IDBCursor
+  val Cursor = raw.IDBCursor
+  type CursorWithValue = IDBCursorWithValue
+  trait IDBEnvironment
```
sjrd: What is this?

src/main/scala/org/scalajs/dom/ShortAliases.scala
```diff
@@ -0,0 +1,279 @@
+package org.scalajs.dom
+
+import org.scalajs.dom.raw._
+
+object css{
```
sjrd: I'd rather have each of these namespaces in their own files.

src/main/scala/org/scalajs/dom/ShortAliases.scala
lihaoyi: Probably a mistake in my regexes

src/main/scala/org/scalajs/dom/ShortAliases.scala
lihaoyi: Sure

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -80,7 +83,7 @@ object Image {
     val img =
       dom.document
         .createElement("img")
-        .asInstanceOf[dom.HTMLImageElement]
+        .asInstanceOf[HTMLImageElement]
```
sjrd: Why not use `html.Image` here, then?

sjrd:
What about things like `Event`. It's not in any of the namespaces, but I don't want to write `raw.Event`. It doesn't feel right.

lihaoyi:
Those guys get aliased in dom.*, inside the package object

src/main/scala/org/scalajs/dom/raw/Idb.scala
```diff
@@ -546,7 +547,7 @@ class IDBDatabase extends EventTarget {
   /**
    * A 64-bit integer that contains the version of the connected database.
    * When a database is first created or upgraded you should use
-   * [[org.scalajs.dom.IDBVersionChangeEvent#newVersion]] instead.
+   * `dom.IDBVersionChangeEvent#newVersion` instead.
```
sjrd: That was a ScalaDoc link. It should be kept.

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,164 @@
...
+  type Text = raw.Text
+  type PositionError = raw.PositionError
+  val PositionError = raw.PositionError
+  type StyleSheetList = raw.StyleSheetList
+  type CustomEvent = raw.CustomEvent
+  type Geolocation = raw.Geolocation
+  type History = raw.History
+  type TimeRanges = raw.TimeRanges
+  type BeforeUnloadEvent = raw.BeforeUnloadEvent
+  type Event = raw.Event
```
lihaoyi: @sjrd here

src/main/scala/org/scalajs/dom/ext/package.scala
```diff
@@ -1,32 +1,34 @@
 package org.scalajs.dom
 
+import org.scalajs.dom.raw._
```
sjrd: Why does this file use `raw._` and not the aliases?

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -1,4 +1,7 @@
-package org.scalajs.dom.extensions
+package org.scalajs.dom.ext
+
+import org.scalajs.dom.raw
+import org.scalajs.dom.raw.{Event, Storage, XMLHttpRequest, HTMLImageElement}
```
sjrd: Same here. Why not use the aliases?

src/main/scala/org/scalajs/dom/ShortAliases.scala
```diff
@@ -0,0 +1,279 @@
...
+  type Program = WebGLProgram
+  type Renderbuffer = WebGLRenderbuffer
+  type Shader = WebGLShader
+  type Texture = WebGLTexture
+  type UniformLocation = WebGLUniformLocation
+  type ActiveInfo = WebGLActiveInfo
+  type ShaderPrecisionFormat = WebGLShaderPrecisionFormat
+  val RenderingContext =  raw.WebGLRenderingContext
+  type RenderingContext = WebGLRenderingContext
+}
```
sjrd: Missing EOL at EOF

example/src/main/scala/example/Example.scala
```diff
@@ -1,9 +1,11 @@
 package example
 
+import org.scalajs.dom.raw._
```
sjrd: And again, we probably don't want to promote `raw` in the example, do we?

benjaminjackman:
Maybe at a top level this should be called org.scalajs.webapi or (just org.scalajs.web) since it superceeds the DOM.  And then all of the methods here could be implemented:
https://developer.mozilla.org/en-US/docs/Web/API

Then a standard namespacing rule could be created, items that have prefixes are lumped together under a package so
org.scalajs.web.html.* holds all the html elements in the shortened form.

Further moz items could be put under a moz namespace as an optional extension, but still coherenet with this.

One last issue is what constitutes a namespace vs what doesn't should all the audio elements be placed into a namespace (e.g. audio.Buffer, audio.Node, audio.Param) or left at the top level (e.g. AudioBuffer AudioNode AudioParam) etc?

lihaoyi:
Yeah, I'm open to renaming things.

> One last issue is what constitutes a namespace vs what doesn't should all the audio elements be placed into a namespace (e.g. audio.Buffer, audio.Node, audio.Param) or left at the top level (e.g. AudioBuffer AudioNode AudioParam) etc?

 Maybe =P I just grabbed things which looked like there was a lot of prefix-repetition and pulled them out. We could pull out more.

benjaminjackman:
There is some implicit namespacing (hierarchies) on this page:
https://developer.mozilla.org/en-US/docs/WebAPI

Basically it breaks down to be all the WebApis, then the DOM get's it's own section (it's near the bottom of the page)

So in that case you would have things like:
org.scalajs.web.dom.html
org.scalajs.web.dom.svg
org.scalajs.web.dom.Event
org.scalajs.web.dom.Document

org.scalajs.web.hardware.

//indexed db
org.scalajs.web.idb.
--or--
org.scalajs.web.data.idb.
org.scalajs.web.data.contacts.

benjaminjackman:
I'm just throwing this out there because if we are breaking stuff and getting the house in order for 1.0 anyhow it might be worth considering how to logically wrap all of the WebAPIs into a coherent hierarchy along with the dom stuff.

Do you guys know/remember how does typescript structures this? I think it's one big flat namespace but I haven't used it for a while.

lihaoyi:
Typescript dumps them all in one huge namespace; when I generated these it was in that form, and I had to split it up into "Html.scala" "Css.scala" etc. manually

benjaminjackman:
Yeah I found it: https://github.com/c9/typescript/blob/master/bin/lib.d.ts

Your right It's a big blob.

 However it as a total aside it reminded of some nice features they have over there, take for example an anchor element:
```typescript
    getElementsByTagName(name: "a"): NodeListOf<HTMLAnchorElement>;
    createElement(tagName: "a"): HTMLAnchorElement;
```

Pretty documentation
```typescript
interface HTMLAnchorElement extends HTMLElement, MSDataBindingExtensions {
    /**
      * Sets or retrieves the relationship between the object and the destination of the link.
      */
    rel: string;
    /**
      * Contains the protocol of the URL.
      */
    protocol: string;
    /**
      * Sets or retrieves the substring of the href property that follows the question mark.
      */
    search: string;
    /**
      * Sets or retrieves the coordinates of the object.
      */
    coords: string;
    /**
      * Contains the hostname of a URL.
      */
    hostname: string;
    /**
      * Contains the pathname of the URL.
      */
    pathname: string;
    Methods: string;
    /**
      * Sets or retrieves the window or frame at which to target content.
      */
    target: string;
    protocolLong: string;
    /**
      * Sets or retrieves a destination URL or an anchor point.
      */
    href: string;
    /**
      * Sets or retrieves the shape of the object.
      */
    name: string;
    /**
      * Sets or retrieves the character set used to encode the object.
      */
    charset: string;
    /**
      * Sets or retrieves the language code of the object.
      */
    hreflang: string;
    /**
      * Sets or retrieves the port number associated with a URL.
      */
    port: string;
    /**
      * Contains the hostname and port values of the URL.
      */
    host: string;
    /**
      * Contains the anchor portion of the URL including the hash sign (#).
      */
    hash: string;
    nameProp: string;
    urn: string;
    /**
      * Sets or retrieves the relationship between the object and the destination of the link.
      */
    rev: string;
    /**
      * Sets or retrieves the shape of the object.
      */
    shape: string;
    type: string;
    mimeType: string;
    /** 
      * Returns a string representation of an object.
      */
    toString(): string;
}
```

And then they can call `new` on their facades
```typescript
declare var HTMLAnchorElement: {
    prototype: HTMLAnchorElement;
    new (): HTMLAnchorElement;
}
```

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,164 @@
 package org.scalajs
-
-package object dom extends dom.Window with scalajs.js.GlobalScope
+import org.scalajs.dom.raw.Window
+package object dom extends Window with scalajs.js.GlobalScope{
+  val XPathResult = raw.XPathResult
```
sjrd: You can't have custom `val`s in a `js.GlobalScope`. That's an accessor to raw JS stuff. The [Travis build](https://travis-ci.org/scala-js/scala-js-dom/jobs/46207850) reports tons of warnings about this.

lihaoyi:
Fixed everything brought up. Also took the opportunity to arrange all the aliases nicely in alphabetical order. Not going to bother doing the same with the original definitions unless someone has an automated tool to do so

lihaoyi:
@benjaminjackman if you feel that the various other faux-namespaces are a pain feel free to send a PR to break them out. There are lots of things that can be broken out (e.g. svg.fe.*) but not gonna do so for now unless someone really cares

src/main/scala/org/scalajs/dom/css.scala
```diff
@@ -0,0 +1,19 @@
+package org.scalajs.dom
+
+/**
+ * Created by haoyi on 1/8/15.
```
sjrd: This ScalaDoc is superfluous.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -1,4 +1,7 @@
-package org.scalajs.dom.extensions
+package org.scalajs.dom.ext
+
+
```
sjrd: Spurious blank line.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -220,7 +223,7 @@ object Ajax{
 }
 
 /**
- * Wraps [[dom.Storage]] replacing null-returning methods with option-returning ones
+ * Wraps [[Storage]] replacing null-returning methods with option-returning ones
```
sjrd: This will now link to itself, not very useful. The `dom.` prefix must be kept here.

src/main/scala/org/scalajs/dom/ext/package.scala
sjrd: My question here still holds.

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,187 @@
 package org.scalajs
+import org.scalajs.dom.raw.Window
+package object dom extends Window with scalajs.js.GlobalScope{
 
-package object dom extends dom.Window with scalajs.js.GlobalScope
+  type AbstractWorker = raw.AbstractWorker
+  type AnimationEvent = raw.AnimationEvent
+  type ApplicationCache = raw.ApplicationCache
+  @inline def ApplicationCache = raw.ApplicationCache
```
sjrd: This still doesn't work (in this file) because `dom` is still a `js.GlobalScope`. To the rhs here is completely ignored (it's basically `js.native`).

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,187 @@
 package org.scalajs
+import org.scalajs.dom.raw.Window
+package object dom extends Window with scalajs.js.GlobalScope{
```
sjrd: Missing blank lines. One before the `import` line and one before the `package object` line.

src/main/scala/org/scalajs/dom/raw/Html.scala
```diff
@@ -8,7 +8,9 @@
  * http://opensource.org/licenses/MIT
  */
 
-package org.scalajs.dom
+package org.scalajs.dom.raw
+
+import org.scalajs.dom._
```
sjrd: Why do you import `dom._` in this file, whereas you import `dom` in `Css.scala` and `Idb.scala` and others?

src/main/scala/org/scalajs/dom/raw/Idb.scala
```diff
@@ -8,7 +8,8 @@
  * http://opensource.org/licenses/MIT
  */
 
-package org.scalajs.dom
+package org.scalajs.dom.raw
+import org.scalajs.dom
```
sjrd: Missing blank line before the `import` line.

src/main/scala/org/scalajs/dom/webgl.scala
```diff
@@ -0,0 +1,16 @@
+package org.scalajs.dom
+
+object webgl{
```
sjrd: Missing space before `{`

src/main/scala/org/scalajs/dom/svg.scala
```diff
@@ -0,0 +1,169 @@
+package org.scalajs.dom
+
+/**
+ * Created by haoyi on 1/8/15.
```
sjrd: Same thing about the ScalaDoc.

src/main/scala/org/scalajs/dom/package.scala
sjrd: And missing space before `{`

src/main/scala/org/scalajs/dom/idb.scala
```diff
@@ -0,0 +1,22 @@
+package org.scalajs.dom
+
+/**
+ * Created by haoyi on 1/8/15.
+ */
+object idb{
```
sjrd: Missing space before `{`.

src/main/scala/org/scalajs/dom/html.scala
```diff
@@ -0,0 +1,82 @@
+package org.scalajs.dom
+
+object html{
```
sjrd: Missing space before `{`.

src/main/scala/org/scalajs/dom/css.scala
```diff
@@ -0,0 +1,19 @@
...
+ */
+object css{
```
sjrd: Missing space before `{`.

src/main/scala/org/scalajs/dom/package.scala
lihaoyi: Would the solution to make `dom` *not* a `js.GlobalScope`? Or would it still be a js.native because of `dom.Window`? What would be a solution here?

sjrd: It would still be native because of `dom.Window`, since `dom.Window` inherits from `js.Object`. There's no real solution here, unfortunately. Other than *not* inheriting from `dom.Window`.

Well, there is one, but it's a bit hacky: just leave it as native ^^ But write it as
```scala
val ApplicatiionCache: raw.ApplicationCache.type = js.native
```
Since it *is* a global scope after all, selecting this field will fetch `global["ApplicationCache"]`, which is precisely what `raw.ApplicationCache` does anyway. And it will have the same type as `raw.ApplicationCache`, so you can select its members normally. It's the best I can come up with.

lihaoyi: I'll go with the singleton type I think

lihaoyi:
Updated

src/main/scala/org/scalajs/dom/raw/Css.scala
```diff
@@ -8,7 +8,9 @@
  * http://opensource.org/licenses/MIT
  */
 
-package org.scalajs.dom
+package org.scalajs.dom.raw
+
+import org.scalajs.dom
```
sjrd: Useless import.

src/main/scala/org/scalajs/dom/raw/Svg.scala
```diff
@@ -8,7 +8,10 @@
  * http://opensource.org/licenses/MIT
  */
 
-package org.scalajs.dom
+package org.scalajs.dom.raw
+
+import org.scalajs.dom
+import org.scalajs.dom.raw.Document
```
sjrd: Useless imports (both of them).

src/main/scala/org/scalajs/dom/raw/WebGL.scala
```diff
@@ -4,11 +4,12 @@
  * Based on https://www.khronos.org/registry/webgl/specs/1.0/
  */
 
-package org.scalajs.dom
+package org.scalajs.dom.raw
+
+import org.scalajs.dom
```
sjrd: Useless import.

src/main/scala/org/scalajs/dom/webgl.scala
```diff
@@ -0,0 +1,18 @@
+package org.scalajs.dom
+/**
```
sjrd: Missing blank line above this line.

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,189 @@
 package org.scalajs
+import org.scalajs.dom.raw.Window
```
sjrd: Missing blank line above this line.

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,189 @@
...
 
-package object dom extends dom.Window with scalajs.js.GlobalScope
+import scala.scalajs.js
+
+package object dom extends Window with scalajs.js.GlobalScope{
```
sjrd: Missing space before `{`.

sjrd:
That's all.

I was a bit skeptical before. But it looks nice, I think.

lihaoyi:
Updated. @sjrd do you have a lint tool or something that finds all these for you? Could save time in future reviews =P

sjrd:
> do you have a lint tool or something that finds all these for you?

I have my eyes ^^ I'm quite pedantic about these things, so I immediately spot these things. If I had a tool, I'd run it as part of CI.

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -1,3 +1,189 @@
...
+  type MessageChannel = raw.MessageChannel
+  type MessageEvent = raw.MessageEvent
+  type MessagePort = raw.MessagePort
+  type ModifierKeyEvent = raw.ModifierKeyEvent
+  type MouseEvent = raw.MouseEvent
+  type MutationEvent = raw.MutationEvent
+  val MutationEvent: raw.MutationEvent.type = js.native
+  type MutationObserver = raw.MutationObserver
+  type MutationObserverInit = raw.MutationObserverInit
+  val MutationObserverInit = raw.MutationObserverInit
```
sjrd: You forgot this one (compiler warning told me so: https://travis-ci.org/scala-js/scala-js-dom/jobs/46440575)

lihaoyi:
Are these going to turn into errors at some point? Or maybe I should just turn on fatal warnings...

sjrd:
> Are these going to turn into errors at some point? Or maybe I should just turn on fatal warnings...

In 1.0.0, yes. See https://github.com/scala-js/scala-js/issues/1111
But you can also turn on fatal warnings, yes, good idea.

That's all. Is the plan to merge this one right away, or are we waiting for some particular event? I remember reading something along those lines but couldn't find any reference anymore.

lihaoyi:
Good question. I was thinking to publish the current version in parallel with the new version for 0.6.0, so at least people can upgrade to 0.6.0 without being blocked on changing all their code. 

On the other hand, upgrading (without modifying too much code) is easy because you just need to import the {raw => dom} and you'll be on your way. So maybe it's not a huge deal asking people to upgrade?

Merging it now and only publishing the new version for 0.6.0 would inconvenience everyone (bad) but force them to upgrade (good). WDYT?

sjrd:
If it's just a matter of publishing, we can always publish 0.7.0 for Scala.js 0.6.0 when it comes out: checkout tag v0.7.0, edit project/build.sbt to change Scala.js' version, and republish. That's what I did for 0.6.0-M3.

lihaoyi:
Cool, in that case I don't mind merging it now. Let's call it 0.8.0?

sjrd:
LGTM

sjrd:
We'll have to call it 0.8.0, yes, given how much is changed in here.

PR #89, opened 2015-02-04
Update docs for 0.8.0

lihaoyi:
- Update sbt snippet
- Added a few words about the various `dom.xxx` namespaces that we now have, including `dom.raw`
- `extensions` -> `ext`
- Re-publish with updated examples (using `html.*`)

@sjrd this is for you to merge whenever 0.6.0 goes out


sjrd:
OK. Could you please squash the two commits in 1 with, as commit message, the title and body of this PR?

sjrd:
Ah, apparently *all* commits in gh-pages have that "." commit message :-s Well, LGTM then.

lihaoyi:
Yeah, if we want to worry about the commit history for gh-pages we can go clean it up later =P 

I wouldn't worry about it; the "real" source version is in master. gh-pages is just a convenient place to publish things and holds no state of its own

PR #95, opened 2015-02-21
Added a method to AjaxException to check for timeout status.

ochrons:
As XHR doesn't offer a clear status for indicating timeout, this method provides an easier way to check for timeout. Tried also using `ontimeout` but it gets called after `onreadystatechange` making it rather useless in this case.

Change-Id: Ia0d66bd4ca27f2ce0157258bcc885817efe21a2f

### DIFF
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -157,7 +157,7 @@ import scala.concurrent.{Promise, Future}
  * Contains the XMLHttpRequest that resulted in that response
  */
 case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
-  def timeout = xhr.status == 0 && xhr.readyState == 4
+  def isTimeout = xhr.status == 0 && xhr.readyState == 4
 }
 
 /**
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -154,9 +154,12 @@ object KeyCode {
 import scala.concurrent.{Promise, Future}
 /**
  * Thrown when `Ajax.get` or `Ajax.post` receives a non-20X response code.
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -154,9 +154,12 @@ object KeyCode {
 import scala.concurrent.{Promise, Future}
 /**
  * Thrown when `Ajax.get` or `Ajax.post` receives a non-20X response code.
- * Contains the XMLHttpRequest that resulted in that respons
+ * Contains the XMLHttpRequest that resulted in that response
  */
-case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception
+case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
+  def timeout = xhr.status == 0 && xhr.readyState == 4
```
sjrd: I'd rather this method be called `isTimeout`.

sjrd:
Is there some spec somewhere backing the assumption about these status and readyState?

sjrd:
Also, wouldn't it be more appropriate to identity timeouts right where the `AjaxException` is created, and throw a `scala.concurrent.TimeoutException` instead?

ochrons:
`scala.concurrent.TimeoutException` would be ok, but breaks backwards compatibility with code that expects to receive an `AjaxException`.

Couldn't find any explicit spec on how to detect timeout (except with ontimeout), but that combination of status values seemed to be used elsewhere (for example https://github.com/scala-js/scala-js-dom/pull/37)

ochrons:
Tested this on Chrome, Firefox and IE11 and the timeout detection works ok on all of them.

sjrd:
OK, I guess that makes sense this way, then. I'd still like it to be called `isTimeout`, to follow Scala's sort-of conventions.

ochrons:
Since I messed up my fork branching and cannot squash these commits, gonna close this PR and redo it The Right Way

PR #96, opened 2015-02-21
Added a method to AjaxException to check for timeout status.

ochrons:
Replaces #95


sjrd:
OK, but what's the "Change-Id" in the commit message?

ochrons:
Ah, that's something we use for Gerrit and it gets automatically added to commits.

sjrd:
Er, could you remove it?

ochrons:
Doh, it of course keeps creating it again... gotta remove the script :)

sjrd:
LGTM Thanks.

PR #98, opened 2015-02-28
GeoLocation Position -> timestamp should be double

chandu0101:
i got the following error with current code

```scala
Uncaught scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: 1425095692856 is not an instance of java.lang.Integer ...
```
when i cast Position to js.Dynamic and then timestamp to Double it worked fine.


sjrd:
LGTM

PR #100, opened 2015-03-08
Node: Fix calls to insertBefore() with one argument

tindzk:


sjrd:
Actually, [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore) that *both* arguments are required. The default value should therefore be removed altogether.

sjrd:
Note that the value put after `=` doesn't matter to the compiler in a JS facade. It only serves documentation purposes. So putting `null` there will *not* make the compiler add a `null` if you don't specify the second argument.

tindzk:
Ok, thanks for clearing this up. I updated the commit accordingly.

sjrd:
LGTM, thanks.

PR #101, opened 2015-03-11
Add ServiceWorker

chandu0101:
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/serviceWorker

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -11,6 +11,7 @@
 package org.scalajs.dom.raw
 
 import scala.scalajs.js
+import scala.scalajs.js.annotation.JSName
 
 object XPathResult extends js.Object {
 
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
 
 /**
+ * The Promise object is used for deferred and asynchronous computations. A Promise is in one of these states:
+ *pending: initial state, not fulfilled or rejected.
+ *fulfilled: successful operation
+ *rejected: failed operation.
+ *settled: the Promise is either fulfilled or rejected, but not pending.
+
+  MDN
+ */
```
sjrd: The style of the ScalaDoc is off, here. As are most of the other ones.
It should be:
```scala
  /**
   * First line.
   * Second line.
   *
   * MDN
   */
```

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
+trait Promise[+A]  extends js.Object{
```
sjrd: Space before `{`. And there is one space too many before `extends`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
+
+  /**
+   * The catch() method returns a Promise and deals with rejected cases only.
+   *  It behaves the same as calling Promise.prototype.then(undefined, onRejected).
+   *
+   *  MDN
+   */
+  @JSName("catch")
+  def catchErr[B](onRejected : js.Function1[Any,B]) : Promise[B] = js.native
```
sjrd: No space before `:`, but space after `,`.

sjrd: Since you're changing the name, you as well use something consistent with Scala's future. I think the corresponding concept would be `recover`, but please check.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
+   */
+  var scriptURL: String = js.native
+
+  /**
+   * The state read-only property of the ServiceWorker interface returns a string representing the current state of the service worker.
+   *  It can be one of the following values: installing, installed, activating, activated, or redundant.
+   *  MDN
+   *
+   */
+  var state: String = js.native
```
sjrd: `def` without `()`, since this is read-only

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
+  var active: ServiceWorker = js.native
+
+  /**
+   * The scope read-only property of the ServiceWorkerRegistration interface returns a unique identifier for a service worker registration.
+   *  The service worker must be on the same origin as the document that registers the ServiceWorker.
+   *  MDN
+   */
+  var scope: String = js.native
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
+  /**
+   * Gets a ServiceWorkerRegistration object whose scope URL matches the document URL.
+   *  If the method can't return a ServiceWorkerRegistration, it returns a Promise.
+   *
+   *  MDN
+   */
+  def getRegistration(scope : js.UndefOr[String] = js.undefined) : Promise[ServiceWorkerRegistration] = js.native
+}
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,196 @@ trait WindowTimers extends WindowTimersExtension {
...
+   */
+  def register(scriptURL : String,options : js.UndefOr[js.Object] = js.undefined) : Promise[ServiceWorkerRegistration] = js.native
+
+  /**
+   * The ServiceWorkerContainer.controller read-only property of the ServiceWorkerContainer interface returns the ServiceWorker whose state is activated (the same object returned by ServiceWorkerRegistration.active).
+   * This property returns null if the request is a force refresh (Shift + refresh) or if there is no active worker.
+   *
+   * MDN
+   */
+  var controller: ServiceWorker = js.native
```
sjrd: `def`

sjrd:
I did not realize you had updated this. Hint: next time you push new commits, post a comment. Just pushing doesn't notify me.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+trait Promise[+A] extends js.Object {
+
+  /**
+   * The catch() method returns a Promise and deals with rejected cases only.
+   * It behaves the same as calling Promise.prototype.then(undefined, onRejected).
+   *
+   * MDN
+   */
+  @JSName("catch")
+  def recover[B](onRejected: js.Function1[Any,B]): Promise[B] = js.native
```
sjrd: I reread the spec more carefully, and did some experiments. The `catch` method, if the promise it is applied to is *successful*, will return a promise that is also successful with that same value. In that sense, it indeed has the same semantics as `recover` in Scala. But then that means the signature is not correct. It should be:
```scala
def recover[B >: A](onRejected: js.Function1[Any, B]): Promise[B] = js.native
```
i.e., `B >: A`.

sjrd: Also, I reiterate my comment about having a space after `,`.

sjrd:
Hum, actually, there is **very serious** typing issue with these `Promises`. The problem is that a `Promise` resolved with another `Promise` will be *flatMapped*, in Scala terms. But this distinction happens at runtime. Grrr. This is really really bad. Basically it means `Promise` is untypable (or results are also `Promise[Any]`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+
+  /**
+   * The then() method returns a Promise.
+   * It takes two arguments, both are callback functions for the success and failure cases of the Promise.
+   *
+   * MDN
+   */
+  @JSName("then")
+  def andThen[B](onFulfilled: js.Function1[A,B]): Promise[B] = js.native
```
sjrd: For example, this basically needs to return a `Primise[Any]`, not a `Promise[B]`, because in case the returned value happens to be a `Promise`, it will be flat-mapped. But we can't know that at compile-time, so it's basically `Any`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+
+  @JSName("then")
+  def andThen[B](onFulfilled: js.Function1[A,B], onRejected: js.Function1[Any,B]): Promise[B] = js.native
```
sjrd: Same here.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+   * MDN
+   */
+  def race[B](iterable: js.Array[Promise[B]]): Promise[B] = js.native
+
+  /**
+   * The Promise.reject(reason) method returns a Promise object that is rejected with the given reason.
+   *
+   * MDN
+   */
+  def reject(reason: String): Promise[Any] = js.native
```
sjrd: `Promise[Nothing]`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+}
+
+object Promise extends js.Object {
+
+  /**
+   * The Promise.all(iterable) method returns a promise that resolves when all of the promises in the iterable argument have resolved.
+   *
+   * MDN
+   */
+  def all[B](iterable: js.Array[Promise[B]]): Promise[B] = js.native
```
sjrd: And here, likely the same `Any` issue.

src/main/scala/org/scalajs/dom/raw/lib.scala
vendethiel: Can't this be an Either ...or `B with Promise[Any]` maybe? Or maybe 2 different methods should exist?

sjrd: It's more problematic than that: it could be any level of nesting of `Promise`s, and they would all be flattened.
Two methods doesn't solve the problem, because you cannot express in the Scala type system that you expect something that is probably *not* a given type (but anything else).

vendethiel: thanks for the explanation

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: for all three methods ? and could you please tell me why it's Nothing instead of Any ?

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: >Hint: next time you push new commits, post a comment. Just pushing doesn't notify me.

ah i thought PR will notify.. , ok next time :)

>Also, I reiterate my comment about having a space after ,.

I thought i fixed all formatting issues , i missed this one apparently . ( I didnt' know that  world out there is serious about formatting! :( )

src/main/scala/org/scalajs/dom/raw/lib.scala
sjrd: No that's just for `reject`. When you create a rejected `Promise`, it is a valid value for *any* kind of `Promise`. This is similar to `Future.failed`, which returns a `Future[Nothing]`.

chandu0101:
@sjrd  i made changes , could you please review them .

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: thanks for explanation :)

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+  /**
+   * Creates or updates a ServiceWorkerRegistration for the given scriptURL.
+   * If successful, a service worker registration ties the provided script URL to a scope, which is subsequently used for navigation matching.
+   * If the method can't return a ServiceWorkerRegistration, it returns a Promise.
+   * You can call this method unconditionally from the controlled page, i.e.,
+   * you don't need to first check whether there's an active registration.
+   *
+   *  MDN
+   */
+  def register(scriptURL: String, options: js.UndefOr[js.Object] = js.undefined): Promise[ServiceWorkerRegistration] = js.native
```
sjrd: `options` should be:
```scala
options: js.Object = ???
```

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -461,6 +462,202 @@ trait WindowTimers extends WindowTimersExtension {
...
+   */
+  def controller: ServiceWorker = js.native
+
+  /**
+   * Gets a ServiceWorkerRegistration object whose scope URL matches the document URL.
+   * If the method can't return a ServiceWorkerRegistration, it returns a Promise.
+   *
+   *  MDN
+   */
+  def getRegistration(scope: js.UndefOr[String] = js.undefined): Promise[ServiceWorkerRegistration] = js.native
```
sjrd:
```
scope: String = ???
```

sjrd:
That's all.

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: >options Optional MDN !

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: > scope Optional MDN !

src/main/scala/org/scalajs/dom/raw/lib.scala
sjrd: Yes, I know. And writing `options: js.Object = ???` is precisely what makes it optional in Scala.js. The `???` is irrelevant and used only to "typecheck". You could also put `= new js.Object()`, probably more self-documenting.

src/main/scala/org/scalajs/dom/raw/lib.scala
sjrd: Same. Although here I'm not sure there's a valid default value for documentation purposes.

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: ah ok thank you. I think i have to opt this in my facades!.

src/main/scala/org/scalajs/dom/raw/lib.scala
chandu0101: scope is relative url 

example  MDN
>navigator.serviceWorker.getRegistration('/app').then(function(registration) {
  document.querySelector('#status').textContent = 'ServiceWorkerRegistration found.';
});

i set default value to ``"/"``  please let me know if you disagree on this ,I'll change to ``???``

sjrd: The main criteria is: if you call it with `"/"`, will it (always) do the same thing as if you call it without the argument? I don't know, you tell me. The documentation isn't clear enough to me to figure it out.

chandu0101: my guess is wrong :( , With ``"/"``  i am getting undefined,  but for ``""`` it doing the same thing as ``getRegistration()``

sjrd:
OK, could you just squash all the commits, and then it'll be good to merge :)

chandu0101:
done! , learned many things . Thank you for answering all my entry level questions :)

sjrd:
LGTM, thanks. :)

PR #102, opened 2015-03-26
WebAudio API

soc:


soc:
Hi, can I get some feedback on this? Is this going into the right direction?

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,228 @@
...
+  * happens inside a context.
+  *
+  * An AudioContext can be a target of events, therefore it implements the EventTarget interface.
+  */
+class AudioContext extends EventTarget {
+  /// Properties ///
+  /** Returns a double representing an ever-increasing hardware time in seconds used for scheduling.
+    *  It starts at 0 and cannot be stopped, paused or reset.
+    */
+  val currentTime: Double = js.native
```
sjrd: This should be a `def`, as it's not constant.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,228 @@
...
+    * The sample-rate of an AudioContext cannot be changed.
+    */
+  val sampleRate: Double = js.native
+
+  /// Methods ///
+
+  /** Creates an AnalyserNode, which can be used to expose audio time and frequency data and for example to create data
+    * visualisations.
+    */
+  def createAnalyser: AnalyserNode = js.native
```
sjrd: This should have `()`, I guess. I assume this is a method that must be called with `()` in JavaScript.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,228 @@
...
+  val pan: AudioParam = js.native
+}
+
+trait WaveShaperNode extends AudioNode {
+  // TODO
+}
+
+//// MISC ////
+
+trait AudioBuffer {
```
sjrd: I guess you need `extends js.Object` here.

sjrd:
Don't forget to read http://www.scala-js.org/doc/calling-javascript.html at least once entirely. It will prevent misconceptions about facades.

src/main/scala/org/scalajs/dom/raw/Audio.scala
soc: Is there some rule-of-the-thumb on what needs `()` and what doesn't?

sjrd: That's explained in the document I was referring to:

> `var`, `val` and `def` definitions without parentheses all map to field access in JavaScript, whereas `def` definitions with parentheses (even empty) map to method calls in JavaScript.

soc:
Thanks, looks like I forgot most of the important stuff since I read it the last time.

So basically all `def`s which refer to JavaScript methods need to receive `()`s, if I'm reading the guide  correctly?

sjrd:
Yes, exactly.

soc:
Done! Nice, this works now in Scala.js: https://github.com/mdn/violent-theremin/blob/gh-pages/scripts/app.js

soc:
Btw, regarding the documentation ... wouldn't it make more sense to have an URL per class to link to the right MDN page and let the IDEs handle the rest? It feels like pretty pointless work to copy this stuff around and has the danger of being outdated.

soc:
@lihaoyi Could you have a look, especially at `object audio`? Because not much started with "Audio" to begin with, and dropping "Audio" could have caused confusion (e. g. `AudioNode` -> `Node`), I just left all names as-is. So we have no abbreviations here, but all the niceties of namespacing.

lihaoyi:
@soc I'd just drop the `audio` object then and put them in the `dom` namespace, as according to our convention

soc:
Mhhh, I really liked having at least some notion of sensible namespaces ... (I was just doing the same for the web worker API).

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1020 @@
...
+
+/** The AudioContext interface represents an audio-processing graph built from audio modules linked together, each
+  * represented by an AudioNode. An audio context controls both the creation of the nodes it contains and the execution
+  * of the audio processing, or decoding. You need to create an AudioContext before you do anything else, as everything
+  * happens inside a context.
+  *
+  * An AudioContext can be a target of events, therefore it implements the EventTarget interface.
+  */
+class AudioContext extends EventTarget {
+  /// Properties ///
```
sjrd: Typically we don't use that kind of headers.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1020 @@
...
+  *  - Number of outputs:      1 (but may be left unconnected)
+  *  - Channel count mode:     "explicit"
+  *  - Channel count:          1
+  *  - Channel interpretation: "speakers"
+  */
+trait AnalyserNode extends AudioNode {
+  /// Properties ///
+  /** Is an unsigned long value representing the size of the FFT (Fast Fourier Transform) to be used to determine the
+    * frequency domain.
+    */
```
sjrd: The recommended ScalaDoc style is:
```scala
/** First line.
 *  Second line.
 */
```
Also, the soft right margin is at 80 characters. When there is value in not cutting the line, the hard margin is at 120. But for running text in comments, for example, it should definitely stay under 80.

The reason is that GitHub mobile displays 80 characters, and GitHub desktop displays 120 characters.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1020 @@
...
+  /** Schedules the start of the playback of the audio asset.
+    *
+    * @param when     The when parameter defines when the play will start. If when represents a time in the past, the
+    *                 play will start immediately. If the method is called more than one time, or after a call to
+    *                 AudioBufferSourceNode.stop(), an exception is raised.
+    * @param offset   The offset parameter, which defaults to 0, defines where the playback will start.
+    * @param duration The duration parameter, which defaults to the length of the asset minus the value of offset,
+    *                 defines the length of the portion of the asset to be played.
+    */
+  def start(when: Double = 0.0, offset: Double = 0.0, duration: Double /* = length - offset */): Unit = js.native
```
sjrd: The duration default value should then be `= ???` (or `= js.native`, I cannot decide anymore ^^). This tells the Scala.js compiler that there *is* a default value. The actual value that is put there is irrelevant as far as compilation is concerned. It is only relevant for documentation.

sjrd:
That's all from me. Maybe @joelross would like to comment on this PR?

lihaoyi:
@soc yeah sensible namespaces would be nice, but so would consistency with the rest of the API. Currently the logic is "either it's top-level, or we pull out a prefix". Having a third "... or we arbitrarily prefixed it" would definitely make things less sensible in the short term.

Maybe in future we'll have another breaking version where we change the convention and reshuffle things, but having one odd-one-out definitely seems like a regression in understandability.

Remember all this stuff is customarily imported under the `dom` namespace anyway, so it's not like any of it will clash with user-defined names, ever.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1020 @@
...
+//// CONTEXTS ////
+
+/** The AudioContext interface represents an audio-processing graph built from audio modules linked together, each
+  * represented by an AudioNode. An audio context controls both the creation of the nodes it contains and the execution
+  * of the audio processing, or decoding. You need to create an AudioContext before you do anything else, as everything
+  * happens inside a context.
+  *
+  * An AudioContext can be a target of events, therefore it implements the EventTarget interface.
+  */
+class AudioContext extends EventTarget {
```
joelross: The parameters are missing for some methods: createBuffer, createChannelMerger, createChannelSplitter, createDelay, createMediaElementSource, createMediaStreamSource, createPeriodicWave and decodeAudioData

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1020 @@
...
+  def connect(audioNode: AudioNode): Unit = js.native
+
+  /** Allows us to connect one output of this node to one input of an audio parameter. */
+  def connect(audioParam: AudioParam): Unit = js.native
+
+  /** Allows us to disconnect the current node from another one it is already connected to.
+    *
+    * @param output The index describing which output of the AudioNode is going to be disconnected.
+    */
+  def disconnect(output: AudioNode): Unit = js.native
```
joelross: It works, but the parameter is optional (none = disconnect all)

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1020 @@
...
+
+  /// Methods ///
+  /** Returns a Float32Array containing the PCM data associated with the channel, defined by the channel parameter
+    * (with 0 representing the first channel).
+    *
+    * @param channel The channel property is an index representing the particular channel to get data for. An index
+    *                value of 0 represents the first channel. If the channel index value is greater than of equal to
+    *                AudioBuffer.numberOfChannels, an INDEX_SIZE_ERR exception will be thrown.
+    */
+  def getChannelData(channel: Int): js.Dynamic = js.native
```
joelross: Scala.js now has wrappers for the typed arrays. We could use them here (js.typedarray.Float32Array)

src/main/scala/org/scalajs/dom/raw/Audio.scala
soc: Is this an issue? I added those headers so that future maintainers can more easily crosscheck the API with the available documentation online.

I can remove it if you want.

src/main/scala/org/scalajs/dom/raw/Audio.scala
soc: Thanks!

src/main/scala/org/scalajs/dom/raw/Audio.scala
sjrd: Yes, please remove them.

src/main/scala/org/scalajs/dom/raw/Audio.scala
soc: Thanks, didn't see that.

src/main/scala/org/scalajs/dom/raw/Audio.scala
soc: Done.

src/main/scala/org/scalajs/dom/raw/Audio.scala
soc: > The recommended ScalaDoc style is:

I find that new-old ScalaDoc style visually jarring, but I guess it's better to be consistent ... I'm on it.

> But for running text in comments, for example, it should definitely stay under 80.

Will do!

soc: Argh... is there a tool to reformat comments? Looks like neither Eclipse nor IntelliJ support this.

soc:
Ok, after manually reflowing 1000 lines, everything should be done now.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1252 @@
...
+  /** Creates a WaveShaperNode, which is used to implement non-linear
+   *  distortion effects. */
+  def createWaveShaper(): WaveShaperNode = js.native
+
+  /** Asynchronously decodes audio file data contained in an ArrayBuffer. In
+   *  this case, the ArrayBuffer is usually loaded from an XMLHttpRequest's
+   *  response attribute after setting the responseType to arraybuffer.
+   *  This method only works on complete files, not fragments of audio files.
+   */
+  def decodeAudioData(audioData: js.typedarray.ArrayBuffer): Promise[AudioBuffer] = js.native
```
joelross: To be honest, I'm not sure what is the actual state of the Promise version of this method in the current browsers. From what I tested, I have the feeling that Firefox is the only one supporting it (Chrome throwed an "2 arguments required, but only 1 present" error when I tried). So it could be safer to have the other version taking the callbacks as parameters too.

The second point I have some doubts, is the current capacity of Scala.js to translate the JavaScript Promises to Scala Promises. I don't think it supports that yet.

By the way, as I understood it, a JavaScript Promise is not strictly the equivalent to a Scala Promise, but:
JavaScript Deferred = Scala Promise (being the ones to "inject" the value/error)
JavaScript Promise = Scala Future (being the ones to "retrieve" the value/error)
(with the new version of JavaScript Promise actually merging the previous JavaScript Deferred & Promise objects... Yeah, simple)

So I think this would actually be a Future[AudioBuffer] instead of a Promise[AudioBuffer]

Maybe @sjrd could confirm or reject what I just said, I'm really not sure about all that.

joelross: New JavaScript version: http://jsfiddle.net/sf4qjaej/5/
```javascript
var myDeferred = {};

var myPromise = new Promise(function (resolve, reject) {
    myDeferred.resolve = resolve;
    myDeferred.reject = reject;
});

myPromise.then(function (v) {
    console.log("Yeah, it's a success: " + v);
});
myPromise.catch(function (t) {
    console.log("Bouh, it's a failure: " + t);
});

myDeferred.resolve(42); // for success
//myDeferred.reject("argh"); // for failure
```
Old JavaScript version (doesn't work on Firefox): http://jsfiddle.net/1w1fn48v/1/
```javascript
var myDeferred = Promise.defer();

var myPromise = myDeferred.promise;

myPromise.then(function (v) {
    console.log("Yeah, it's a success: " + v);
});
myPromise.catch(function (t) {
    console.log("Bouh, it's a failure: " + t);
});

myDeferred.resolve(42); // for success
//myDeferred.reject("argh"); // for failure
```
Scala version
```scala
import scala.concurrent._
import ExecutionContext.Implicits.global

val myPromise = Promise[Int]

val myFuture = myPromise.future

myFuture.onSuccess { case v =>
  println("Yeah, it's a success: " + v)
}
myFuture.onFailure { case t =>
  println("Bouh, it's a failure: " + t)
}

myPromise.success(42) // for success
//myPromise.failure(new RuntimeException("arfgh")) // for failure
```

soc: > So it could be safer to have the other version taking the callbacks as parameters too.

Yes, I can adopt what the WebIDL says, but I'm not quite sure how the callbacks need to be translated for Scala.js.

IDL:
```
callback DecodeSuccessCallback = void (AudioBuffer decodedData);
callback DecodeErrorCallback = void ();

    [Throws]
    Promise<AudioBuffer> decodeAudioData(ArrayBuffer audioData,
                                         optional DecodeSuccessCallback successCallback,
                                         optional DecodeErrorCallback errorCallback);
```

Are they just standard `js.FunctionX`s?

> The second point I have some doubts, is the current capacity of Scala.js to translate the JavaScript Promises to Scala Promises. I don't think it supports that yet.

I understand, but how is it relevant here? I'm returning a `dom.raw.Promise` and I have no real alternative ... I can't pretend it returns a scala.concurrent.Future, because then the method names and the semantics are all different from what the JavaScript API actually supports.

There might be future work to better integrate the JavaScript Promises with Scala's Futures&Promises, but I think that's out of scope for this PR ...

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -0,0 +1,1268 @@
...
+   *                         representing the decoded PCM audio data. Usually
+   *                         you'll want to put the decoded data into an
+   *                         AudioBufferSourceNode, from which it can be played
+   *                         and manipulated how you want.
+   *  @param errorCallback   An optional error callback, to be invoked if an
+   *                         error occurs when the audio data is being decoded.
+   */
+  def decodeAudioData(audioData: js.typedarray.ArrayBuffer,
+                      successCallback: js.Function1[AudioBuffer, _] = js.native,
+                      errorCallback: js.Function0[_] = js.native): Promise[AudioBuffer] = js.native
```
soc: @joelross Is this correct this way?

joelross: I think it is, yes.

> I understand, but how is it relevant here? I'm returning a dom.raw.Promise and I have no real alternative ... I can't pretend it returns a scala.concurrent.Future, because then the method names and the semantics are all different from what the JavaScript API actually supports.

Actually, you are right, it's not relevant. I just thought you were using a scala.concurrent.Promise here... My fault, sorry!

soc: No problem, I was just confused, too. Thanks!

soc:
Is there anything left to do here?

sjrd:
LGTM, thanks :)

PR #103, opened 2015-03-29
Add key codes for F1 to F12

japgolly:


sjrd:
Hum, uppercase `F1`? Isn't that more common?

japgolly:
There are two conventions that motivated me to use lowercase.

1. All the other vals in that `object` are lowercase.
2. Scala variables are often lowercase with uppercase representing types.

Happy to change it if you'd still prefer I do.

sjrd:
Ah, you're right about 1. I believe that's a mistake we should fix globally. (with deprecated aliases)
Scala variables are lowercase, but *constants* are typically written with an uppercase.

japgolly:
np, replaced with #105

Issue #104, opened 2015-03-29
vals for KeyCodes?


japgolly:
@sjrd Question about efficiency here. We have:
```scala
object KeyCode {
  val backspace = 8
  val tab = 9
  // etc
```

Is there effeciency to be gained by changing to `final val`s?

I was thinking `@inline def` would be best but then they can be used in pattern matching.

sjrd:
`final val` is best, indeed.

japgolly:
Cool, I'll make another PR for this after we finish with #103.

japgolly:
Lumped into #105

PR #105, opened 2015-03-31
Update Extensions.scala

japgolly:
* Added key codes F1~F12
* Made KeyCodes start with uppercase
* Changed `val` to `final val`

Addresses #103 and #104.


src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -96,62 +96,131 @@ object Image {
...
-  val s = 83
-  val t = 84
-  val u = 85
-  val v = 86
-  val w = 87
-  val x = 88
-  val y = 89
-  val z = 90
+  final val Backspace = 8
+  final val Tab       = 9
```
sjrd: Do not align vertically. Doing it means you need to modify all lines when you add something.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -60,15 +60,15 @@ object Color {
...
-  val all = Seq(
+  final val White   = Color(255, 255, 255)
+  final val Red     = Color(255, 0, 0)
+  final val Green   = Color(0, 255, 0)
+  final val Blue    = Color(0, 0, 255)
+  final val Cyan    = Color(0, 255, 255)
+  final val Magenta = Color(255, 0, 255)
+  final val Yellow  = Color(255, 255, 0)
+  final val Black   = Color(0, 0, 0)
+  final val all = Seq(
```
sjrd: There's no point for this set to be final vals. It's only meaningful for constant expressions of primitive types. So don't change these.

sjrd:
The three bullets should be 3 commits.

sjrd:
That's all.

japgolly:
Feedback adhered to and resubmitted in #109

PR #106, opened 2015-04-09
Narrow ownerDocument for HTML elements

jducoeur:
If we have an HTMLElement, then we know that the ownerDocument is
specifically an HTMLDocument, not a plain Document.  That's useful type
information, since you often want to use HTML-specific methods on it.


sjrd:
LGTM

PR #108, opened 2015-04-22
Added support for sending ByteBuffers, typed arrays and Blobs in Ajax calls.

ochrons:


project/build.sbt
```diff
@@ -1,3 +1,3 @@
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.0")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.2")
```
sjrd: Please put this is in a separate, first commit.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -167,42 +173,74 @@ case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
  * Wraps an XMLHttpRequest to provide an easy one-line way of making 
  * an Ajax call, returning a Future.
  */
-object Ajax{
+object Ajax {
+  /**
+   * Supported data formats for Ajax are implicitly converted to Ajaxable
+   */
+  sealed trait Ajaxable extends js.Any
```
sjrd: `Ajaxable` was a fine word to discuss the "proposal", but it seems like a very weird name for real code.

What about `InputData` (therefore with qualified name `Ajax.InputData`)?

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -167,42 +173,74 @@ case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
...
+        data.typedArray().subarray(data.position, data.limit)
+      } else {
+        // fall back to copying the data
+        val tempBuffer = ByteBuffer.allocateDirect(data.remaining)
+        val origPosition = data.position
+        tempBuffer.put(data)
+        data.position(origPosition)
+        tempBuffer.typedArray()
+      }
+      b.asInstanceOf[Ajaxable]
```
sjrd: This last line should not be needed. If you just leave the `if` as the result, then the `implicit arraybuffer2ajax` should kick in to cast the `typedArray` into `Ajaxable` already.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -167,42 +173,74 @@ case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
...
+object Ajax {
+  /**
+   * Supported data formats for Ajax are implicitly converted to Ajaxable
+   */
+  sealed trait Ajaxable extends js.Any
+
+  object Ajaxable {
+    implicit def str2ajax(s: String): Ajaxable = s.asInstanceOf[Ajaxable]
+
+    implicit def arraybuffer2ajax(b: ArrayBufferView): Ajaxable = b.asInstanceOf[Ajaxable]
```
sjrd: I would call this `arrayBufferView2ajax`

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -167,42 +173,74 @@ case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
...
+
+    implicit def blob2ajax(b: Blob): Ajaxable = b.asInstanceOf[Ajaxable]
+
+    implicit def bytebuffer2ajax(data: ByteBuffer): Ajaxable = {
```
sjrd: And, similarly, `byteBuffer2ajax`

sjrd:
The commit message should use present tense (Add instead of Added).

sjrd:
So I guess we drop backwards binary compat?

sjrd:
That's all.

sjrd:
LGTM

PR #110, opened 2015-04-23
change websocket MessageEvent.data to an Any from a js.Any, add overload...

codingismy11to7:
...s to WebSocket.send to specify the types of data that can be sent


sjrd:
Hum, why didn't I see this earlier?

The code is good. The commit message is way too long for a first line. I suggest this:

    Give better types to MessageEvent.data and WebSocket.send.

    * MessageEvent.data is now of type Any, because it can be a String.
    * WebSocket.send has overloads with precise types for valid arguments.

codingismy11to7:
Rebased onto latest master and changed the commit message

sjrd:
LGTM, thanks :)

PR #112, opened 2015-05-06
Add ClipboardEvent

mseddon:
Add support for ClipboardEvent.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -3763,6 +3763,21 @@ trait DataTransfer extends js.Object {
   def files: FileList = js.native
 }
 
+/**
+ * The ClipboardEvent interface represents events providing information related to
+ * modification of the clipboard, that is cut, copy, and paste events.
+ *
+ * MDN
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3763,6 +3763,21 @@ trait DataTransfer extends js.Object {
   def files: FileList = js.native
 }
 
+/**
+ * The ClipboardEvent interface represents events providing information related to
+ * modification of the clipboard, that is cut, copy, and paste events.
+ *
+ * MDN
+ */
+class ClipboardEvent extends Event {
```
sjrd: The constructor is not well-typed. See its documentation at https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent/ClipboardEvent

mseddon:
Oops! Just added that one.  I've provided a js.Dynamic parameter for the second parameter for now, modelled after the EventSource, but if there's a better definition I'm happy to change it.

sjrd:
LGTM I guess `js.Dynamic` is good enough for now.

PR #113, opened 2015-05-06
Expose ClipboardEvent in dom package

mseddon:
We should also expose ClipboardEvent as dom.ClipboardEvent.


sjrd:
LGTM.

Note-to-self: duh, next time I should really catch that during review ...

mseddon:
:) I could do better sanity checking my PRs sometimes.

PR #115, opened 2015-05-13
Constructor for BlobPropertyBag. Fixes #114

### LINKED ISSUE #114: BlobPropertyBag needs a constructor
Not terribly high priority, but since [it seems to be causing confusion](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/scala-js/IHvjcmNhFdk/48et0T-2l0QJ): BlobPropertyBag is, if I understand it correctly, basically an options object.  As such, the facade isn't actually useful on its own -- there's no good way to use the facade of BlobPropertyBag to create an object to pass into the Blob constructor.  It really needs a BlobPropertyBag constructor somewhere to make it functional.

mseddon:


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -6773,6 +6773,15 @@ trait BlobPropertyBag extends js.Object {
   def endings: String = js.native
 }
 
+object BlobPropertyBag {
+  @inline
+  def apply(`type`: js.UndefOr[String] = js.undefined): BlobPropertyBag = {
+    val result = js.Dynamic.literal()
+    `type`.foreach(result.`type`= _)
```
sjrd: There's just a missing space before the `=` sign. That's all I have to say ;)

sjrd:
Oh also please format the commit message as

> Fix #114: Add a constructor for BlobPropertyBag.

Notice the "Fix #114: " in the beginning, the imperative style of "Add", and the "." at the end ;)

mseddon:
Guess what I forgot at the end? :) There.

sjrd:
LGTM, thanks.

PR #116, opened 2015-05-13
Add CSS property 'resize'

chrisghost:
MDN : https://developer.mozilla.org/en-US/docs/Web/CSS/resize

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/Css.scala
+++ b/src/main/scala/org/scalajs/dom/raw/Css.scala
@@ -171,6 +171,7 @@ class CSSStyleDeclaration extends js.Object {
   var overflowX: String = js.native
   var textAlign: String = js.native
   var margin: String = js.native
+  var resize: String = js.native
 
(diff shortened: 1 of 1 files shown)


sjrd:
This seems to be quite experimental, at this point. It's not supported by IE nor Opera. @lihaoyi I don't remember what's our policy in such cases?

lihaoyi:
The policy was "only stuff which is present in all major browsers", i.e. latest(-ish) Chrome/Firefox/IE. The words to that effect seems to have disappeared in the port of the readme to scalatex, I'll put it back. As such, I don't think we should merge this. Having all the included facades be solid & dependable is I think a nice property to have.

lihaoyi:
We now have a place to put all these experimental things, as in https://github.com/scala-js/scala-js-dom/pull/135

e.g. this would be an implicit conversion on the relevant type, that would need to be imported from `import org.scalajs.dom.experimental.CSSResize._`

PR #117, opened 2015-05-14
Improve ClipboardEvent type saftey.

mseddon:


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3763,13 +3763,42 @@ trait DataTransfer extends js.Object {
   def files: FileList = js.native
 }
 
+trait ClipboardEventInit extends js.Object {
+  /**
+   * The data for this clipboard event.
+   */
+  val data: String
```
sjrd: Needs `= js.native`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3763,13 +3763,42 @@ trait DataTransfer extends js.Object {
...
+
+  /**
+   * The MIME type of the data.
+   */
+  val dataType: String
```
sjrd: Same here.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3763,13 +3763,42 @@ trait DataTransfer extends js.Object {
...
+object ClipboardEventInit {
+  /**
+   * Construct a new ClipboardEventInit
+   * 
+   * @param data       The data for this clipboard event
+   * @param dataType   The MIME type of the data.
+   * @return a new ClipBoardEventInit
+   */
+  @inline
+  def apply(data: js.UndefOr[String] = js.undefined, dataType: js.UndefOr[String]): ClipboardEventInit = {
```
sjrd: `dataType` is missing `= js.undefined`.

sjrd: Also, break the line before `dataType`, with 4 spaces of indentation, to keep the line length under 80 characters.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3763,13 +3763,42 @@ trait DataTransfer extends js.Object {
...
+
 /**
  * The ClipboardEvent interface represents events providing information related to
  * modification of the clipboard, that is cut, copy, and paste events.
  *
  * MDN
  */
-class ClipboardEvent(`type`: String, settings: js.Dynamic) extends Event {
+class ClipboardEvent(`type`: String, settings: ClipboardEventInit) extends Event {
   /**
```
sjrd: We can easily avoid breaking existing code by keeping the `js.Dynamic` constructor as `@deprecated`:
```scala
@deprecated("Use the overload with a ClipboardEventInit instead.", "0.8.1")
def this(`type`: String, settings: js.Dynamic) = this(??? : String, ??? : ClipboardEventInit)
```

sjrd:
There's also a typo in the commit message (saftey).
That's all.

sjrd:
LGTM

PR #124, opened 2015-06-05
Implement Web Cryptography API

mseddon:
Implements http://www.w3.org/TR/WebCryptoAPI/

### DIFF
--- a/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
+++ b/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
@@ -1,0 +1,672 @@
+package org.scalajs.dom.crypto
+
+import scalajs.js
+import org.scalajs.dom.raw._
+import scala.scalajs.js.typedarray.ArrayBufferView
+
+object GlobalCrypto extends js.GlobalScope {
+  val crypto: Crypto = js.native
+}
+
+/**
+ * The Crypto interface represents basic cryptography features available in the
+ * current context. It allows access to a cryptographically strong random
+ * number generator and to cryptographic primitives.
+ *
+ * MDN
+ */
+trait Crypto extends js.Object {
+  /**
+   * Returns a SubtleCrypto object providing access to common cryptographic
+   * primitives, like hashing, signing, encryption or decryption.
+   *
+   * MDN
+   */
+  val subtle: SubtleCrypto = js.native
+
+  /**
+   * Fills the passed TypedArray with cryptographically sound random values.
+   *
+   * MDN
+   */
+  def getRandomValues(array: ArrayBufferView): ArrayBufferView = js.native
+}
+
+trait Algorithm extends js.Object {
+  var name: String = js.native
+}
+
+trait KeyAlgorithm extends Algorithm
+
+trait CryptoKey extends js.Object {
+  val `type`: String = js.native
+
+  val extractable: Boolean = js.native
+
+  val algorithm: js.Object = js.native
+
+  val usages: js.Object = js.native
+}
+
+trait RsaOtherPrimesInfo extends js.Object {
+  var r: String = js.native
+
+  var d: String = js.native
+
+  var t: String = js.native
+}
+
+trait JsonWebKey extends js.Object {
+  var kty: String = js.native
+
+  var use: String = js.native
+
+  var key_ops: js.Array[String] = js.native
+
+  var alg: js.Array[String] = js.native
+
+  var ext: Boolean = js.native
+
+  var crv: String = js.native
+
+  var x: String = js.native
+
+  var y: String = js.native
+
+  var d: String = js.native
+
+  var n: String = js.native
+
+  var e: String = js.native
+
+  var p: String = js.native
+
+  var q: String = js.native
+
+  var dp: String = js.native
+
+  var dq: String = js.native
+
+  var qi: String = js.native
+
+  var oth: js.Array[String] = js.native
+
+  var k: String = js.native
+}
+
+/**
+ * The SubtleCrypto interface represents a set of cryptographic primitives.
+ * It is available via the Crypto.subtle properties available in a window
+ * context (via Window.crypto).
+ *
+ * MDN
+ */
+trait SubtleCrypto extends js.Object {
+  /**
+   * Returns a Promise of the encrypted data corresponding to the clear text,
+   * algorithm and key given as parameters.
+   *
+   * MDN
+   */
+  def encrypt(algorithm: String, key: CryptoKey,
+      data: BufferSource): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of the clear data corresponding to the encrypted text,
+   * algorithm and key given as parameters.
+   *
+   * MDN
+   */
+  def decrypt(algorithm: String, key: CryptoKey,
+      data: BufferSource): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of the signature corresponding to the text, algorithm
+   * and key given as parameters.
+   *
+   * MDN
+   */
+  def sign(algorithm: String, key: CryptoKey,
+      data: BufferSource): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a Boolean value indicating if the signature given as
+   * parameter matches the text, algorithm and key also given as parameters.
+   *
+   * MDN
+   */
+  def verify(algorithm: String, key: CryptoKey, signature: BufferSource,
+      data: BufferSource): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a digest generated from the algorithm and text given
+   * as parameters.
+   *
+   * MDN
+   */
+  def digest(algorithm: String, data: BufferSource): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a newly generated CryptoKey, for symmetrical
+   * algorithms, or a CryptoKeyPair, containing two newly generated keys, for
+   * asymmetrical algorithm, that matches the algorithm, the usages and the
+   * extractability given as parameters.
+   *
+   * MDN
+   */
+  def generateKey(algorithm: String, extractable: Boolean,
+      keyUsages: js.Array[String]): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a newly generated CryptoKey derivated from a master
+   * key and a specific algorithm given as parameters.
+   *
+   * MDN
+   */
+  def deriveKey(algorithm: String, baseKey: CryptoKey,
+      derivedKeyType: String, extractable: Boolean,
+      keyUsages: js.Array[String]): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a newly generated buffer of pseudo-random bits
+   * derivated from a master key and a specific algorithm given as parameters.
+   *
+   * MDN
+   */
+  def deriveBits(algorithm: String, baseKey: CryptoKey,
+      length: Double): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a CryptoKey corresponding to the format, the
+   * algorithm, the raw key data, the usages and the extractability given
+   * as parameters.
+   *
+   * MDN
+   */
+  def importKey(format: String, keyData: BufferSource,
+      algorithm: String, extractable: Boolean,
+      keyUsages: js.Array[String]): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a CryptoKey corresponding to the format, the
+   * algorithm, the raw key data, the usages and the extractability given as
+   * parameters.
+   *
+   * MDN
+   */
+  def importKey(format: String, keyData: JsonWebKey,
+      algorithm: String, extractable: Boolean,
+      keyUsages: js.Array[String]): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a buffer containing the key in the format requested.
+   *
+   * MDN
+   */
+  def exportKey(format: String, key: CryptoKey): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a wrapped symmetric key for usage (transfer, storage)
+   * in unsecure environments. The wrapped buffer returned is in the format
+   * given in parameters, and contained the key wrapped by the give wrapping
+   * key with the given algorithm.
+   *
+   * MDN
+   */
+  def wrapKey(format: String, key: CryptoKey, wrappingKey: CryptoKey,
+      wrapAlgorithm: String): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of a CryptoKey corresponding to the wrapped key given in
+   * parameter.
+   *
+   * MDN
+   */
+  def unwrapKey(format: String, wrappedKey: BufferSource,
+      unwrappingKey: CryptoKey, unwrapAlgorithm: String,
+      unwrappedKeyAlgorithm: String, extractable: Boolean,
+      keyUsages: js.Array[String]): Promise[js.Any] = js.native
+}
+
+// RSASSA-PKCS1-v1_5
+
+trait RsaKeyGenParams extends Algorithm {
+  var modulusLength: Double = js.native
+
+  var publicExponent: BigInteger = js.native
+}
+
+object RsaKeyGenParams {
+  @inline
+  def apply(name: String, modulusLength: Long,
+      publicExponent: BigInteger): RsaKeyGenParams = {
+    js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
+        publicExponent = publicExponent).asInstanceOf[RsaKeyGenParams]
+  }
+}
+
+trait RsaHashedKeyGenParams extends RsaKeyGenParams {
+  var hash: String = js.native
+}
+
+object RsaHashedKeyGenParams {
+  @inline
+  def apply(name: String, modulusLength: Long,
+      publicExponent: BigInteger, hash: String): RsaHashedKeyGenParams = {
+    js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
+        publicExponent = publicExponent,
+        hash = hash).asInstanceOf[RsaHashedKeyGenParams]
+  }
+}
+
+trait RsaKeyAlgorithm extends KeyAlgorithm {
+  var modulusLength: Double = js.native
+
+  var publicExponent: BigInteger = js.native
+}
+
+object RsaKeyAlgorithm {
+  @inline
+  def apply(name: String, modulusLength: Long,
+      publicExponent: BigInteger): RsaKeyAlgorithm = {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,755 @@
+package org.scalajs.dom.raw
+
+import scalajs.js
+import scala.scalajs.js.typedarray.{Uint8Array, ArrayBuffer, ArrayBufferView}
+
+private[dom] object CryptoTypes {
+  type BigInteger = Uint8Array
+  sealed trait BufferSource extends js.Any
+  implicit def arrayBuffer2BufferedSource(b: ArrayBuffer): BufferSource =
```
sjrd: Add blank lines between members.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,755 @@
...
+    b.asInstanceOf[BufferSource]
+}
+
+trait GlobalCrypto extends js.Any {
+  val crypto: Crypto = js.native
+}
+
+/**
+ * The Crypto interface represents basic cryptography features available in the current context. It allows access to a
+ * cryptographically strong random number generator and to cryptographic primitives.
```
sjrd: These lines are too long. Break the lines so that they stay under 80 characters.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,755 @@
...
+ *
+ * MDN
+ */
+trait Crypto extends js.Any {
+  /**
+   * Returns a SubtleCrypto object providing access to common cryptographic primitives, like hashing, signing,
+   * encryption or decryption.
```
sjrd: Same here.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,755 @@
+package org.scalajs.dom.raw
```
sjrd: Would it make sense to put this in `org.scalajs.dom.crypto` instead?

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,755 @@
...
+trait RsaKeyGenParams extends Algorithm {
+  import CryptoTypes._
+  var modulusLength: Double = js.native
+
+  var publicExponent: BigInteger = js.native
+}
+
+object RsaKeyGenParams {
+  import CryptoTypes._
+  @inline
```
sjrd: Add a blank line above `@inline`

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,755 @@
...
+  def apply(name: String, modulusLength: Long, publicExponent: BigInteger): RsaKeyGenParams = {
```
sjrd: What's the point of taking a `Long` if you're going to convert it to a `Double` anyway. Why not take a `Double`?

sjrd: Seeing `BigInteger` like this is disturbing. I keep thinking it's a Java/Scala BigInteger, not an `Uint8Array` as was aliased at the top of the file. Why not writing `Uint8Array`?

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+}
+
+/**
+ * The Crypto interface represents basic cryptography features available in the
+ * current context. It allows access to a cryptographically strong random
+ * number generator and to cryptographic primitives.
+ *
+ * MDN
+ */
+trait Crypto extends js.Any {
```
sjrd: Extend from `js.Object` instead (same for all other classes here, except `BufferSource`).

src/main/scala/org/scalajs/dom/raw/Crypto.scala
mseddon: I strongly agree.  org.scalajs.dom.raw is rather cluttered.  I'm unsure why pretty much *everything* is defined in here.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+
+  implicit def arrayBuffer2BufferSource(b: ArrayBuffer): BufferSource =
+    b.asInstanceOf[BufferSource]
+
+  implicit def arrayBufferView2BufferSource(b: ArrayBufferView): BufferSource =
+    b.asInstanceOf[BufferSource]
+}
+
+trait GlobalCrypto extends js.Any {
+  val crypto: Crypto = js.native
```
sjrd: I think this would have a better place in the `crypto` package object.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+trait SubtleCrypto extends js.Any {
+  import CryptoTypes._
+  /**
+   * Returns a Promise of the encrypted data corresponding to the clear text,
+   * algorithm and key given as parameters.
+   *
+   * MDN
+   */
+  def encrypt(algorithm: String, key: CryptoKey,
+    data: BufferSource): Promise[js.Any] = js.native
```
sjrd: Use 4 spaces of indentation when wrapping inside `()`.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+
+  /**
+   * Returns a Promise of the encrypted data corresponding to the clear text,
+   * algorithm and key given as parameters.
+   *
+   * MDN
+   */
+  def encrypt(algorithm: Algorithm, key: CryptoKey,
+    data: BufferSource): Promise[js.Any] = js.native
```
sjrd: I cannot find evidence of the existence of this overload. [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/encrypt) `algorithm` must be a `String`.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+  /**
+   * Returns a Promise of the clear data corresponding to the encrypted text,
+   * algorithm and key given as parameters.
+   *
+   * MDN
+   */
+  def decrypt(algorithm: String, key: CryptoKey,
+    data: BufferSource): Promise[js.Any] = js.native
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
sjrd: The result is a `Promise[ArrayBuffer]`, actually.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+    data: BufferSource): Promise[js.Any] = js.native
+
+  /**
+   * Returns a Promise of the clear data corresponding to the encrypted text,
+   * algorithm and key given as parameters.
+   *
+   * MDN
+   */
+  def decrypt(algorithm: String, key: CryptoKey,
+    data: BufferSource): Promise[js.Any] = js.native
```
sjrd: The result is a `Promise[ArrayBuffer]`, actually.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+  /**
+   * Returns a Promise of the signature corresponding to the text, algorithm
+   * and key given as parameters.
+   *
+   * MDN
+   */
+  def sign(algorithm: String, key: CryptoKey,
+    data: BufferSource): Promise[js.Any] = js.native
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+  /**
+   * Returns a Promise of a digest generated from the algorithm and text given
+   * as parameters.
+   *
+   * MDN
+   */
+  def digest(algorithm: String, data: BufferSource): Promise[js.Any] = js.native
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+   * Returns a Promise of a CryptoKey corresponding to the format, the
+   * algorithm, the raw key data, the usages and the extractability given as
+   * parameters.
+   *
+   * MDN
+   */
+  def importKey(format: String, keyData: BufferSource,
+    algorithm: Algorithm, extractable: Boolean,
+    keyUsages: js.Array[String]): Promise[js.Any] = js.native
+  /**
```
sjrd: Missing blank line.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+object RsaKeyGenParams {
+  import CryptoTypes._
+  @inline
+  def apply(name: String, modulusLength: Long,
+    publicExponent: BigInteger): RsaKeyGenParams = {
+    val result = js.Dynamic.literal()
+    result.name = name
+    result.modulusLength = modulusLength.toDouble
+    result.publicExponent = publicExponent
+    result.asInstanceOf[RsaKeyGenParams]
```
sjrd: Why not
```scala
js.Dynamic.literal(
    name = name,
    modulusLength = modulusLength.toDouble,
    publicExponent = publicExponent
).asInstanceOf[RsaKeyGenParams]
```
?

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+    result.modulusLength = modulusLength.toDouble
+    result.publicExponent = publicExponent
+    result.hash = hash
+    result.asInstanceOf[RsaHashedKeyGenParams]
+  }
+}
+
+trait RsaKeyAlgorithm extends KeyAlgorithm {
+  import CryptoTypes._
+  var modulusLength: Double = js.native
```
sjrd: Missing blank line.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+    result.hash = hash
+    result.asInstanceOf[RsaHashedKeyAlgorithm]
+  }
+}
+
+trait RsaHashedImportParams extends js.Any {
+  var hash: String = js.native
+}
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+object RsaHashedKeyAlgorithm {
+  @inline
+  def apply(name: String, modulusLength: Long,
+    publicExponent: Uint8Array, hash: String): RsaHashedKeyAlgorithm = {
+    val result = js.Dynamic.literal()
+    result.name = name
+    result.modulusLength = modulusLength.toDouble
+    result.publicExponent = publicExponent
+    result.hash = hash
+    result.asInstanceOf[RsaHashedKeyAlgorithm]
```
sjrd: Same about `js.Dynamic.literal`.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
mseddon: Hmm, where did you get that? I based that signature from http://www.w3.org/TR/WebCryptoAPI/#subtlecrypto-interface

That said, it's been returning me ArrayBuffers in my tests.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+
+trait RsaHashedImportParams extends js.Any {
+  var hash: String = js.native
+}
+
+
+object RsaHashedImportParams {
+  @inline
+  def apply(hash: String): RsaHashedImportParams =
+    js.Dynamic.literal("hash" -> hash).asInstanceOf[RsaHashedImportParams]
```
sjrd: With `literal`, the named parameter notation is more canonical when the names are known: `literal(hash = hash)`

src/main/scala/org/scalajs/dom/raw/Crypto.scala
sjrd: Hum, I got that from https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#rsa-oaep-registration and other similar entries.

But then apparently this depends on the algorithm being used, in theory. Although all algorithms mentioned in the spec say `ArrayBuffer`.

Let's be safe and keep `js.Any` for now.

src/main/scala/org/scalajs/dom/raw/Crypto.scala
```diff
@@ -0,0 +1,831 @@
...
+
+trait RsaPssParams extends Algorithm {
+  var saltLength: Double = js.native
+}
+
+object RsaPssParams {
+  @inline
+  def apply(name: String, saltLength: Long): RsaPssParams =
+    js.Dynamic.literal("name" -> name,
+      "saltLength" -> saltLength.toDouble).asInstanceOf[RsaPssParams]
```
sjrd: A method which is not a one-liner should have `{}`

src/main/scala/org/scalajs/dom/raw/Crypto.scala
mseddon: Alright.  It's a pity that the effort I am going to help type parameters, how everything gets absorbed by Any, but I suppose that's just ECMAScript for you...

sjrd: Well I guess we could make the type of result path-dependent on the algorithm being used :-p

mseddon: Hah- I thought about that. But then I figured I'd botch it horribly :)

src/main/scala/org/scalajs/dom/raw/Crypto.scala
mseddon: I... don't know.  It was early :)

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1840,7 +1840,9 @@ trait WindowSessionStorage extends js.Object {
  *
  * MDN
  */
-class Window extends EventTarget with WindowLocalStorage with WindowSessionStorage with WindowTimers with WindowBase64 with IDBEnvironment with WindowConsole {
+class Window extends EventTarget with WindowLocalStorage
+  with WindowSessionStorage with WindowTimers with WindowBase64
+  with IDBEnvironment with WindowConsole with GlobalCrypto {
```
sjrd: I don't think this is needed. I would encourage people to use `org.scalajs.dom.crypto.crypto` instead of `org.scalajs.dom.window.crypto`.

mseddon:
Okay, I figured out my bizarre crash- the Window.crypto member in dom (when I previously mixed GlobalCrypto into Window) was conflicting with my crypto package object, and the compiler wasn't handling it particularly well.  That should be the lot, hopefully!  All operations now happen through crypto.crypto as suggested.

mseddon:
Whoops, forgot to remove those spurious Algorithm parameter overloads on the Crypto object.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -0,0 +1,673 @@
...
+
+  var publicExponent: BigInteger = js.native
+}
+
+object RsaKeyGenParams {
+  @inline
+  def apply(name: String, modulusLength: Long,
+      publicExponent: BigInteger): RsaKeyGenParams = {
+    js.Dynamic.literal(name = name, modulusLength = modulusLength,
+        publicExponent = publicExponent).asInstanceOf[RsaKeyGenParams]
```
sjrd: Ah ah now you have the worst both worlds: `modulusLength` of type `Long`, and *not* `.toDouble`. You need either of those. As I told earlier, I would advocate declaring `modulusLength` as a `Double`.

Possibly with an overload taking a `Long` and calling the `Double` version with `.toDouble`.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -0,0 +1,673 @@
...
+        publicExponent = publicExponent,
+        hash = hash).asInstanceOf[RsaHashedKeyAlgorithm]
+  }
+}
+
+trait RsaHashedImportParams extends js.Object {
+  var hash: String = js.native
+}
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -0,0 +1,673 @@
...
+  /**
+   * Fills the passed TypedArray with cryptographically sound random values.
+   *
+   * MDN
+   */
+  def getRandomValues(array: ArrayBufferView): ArrayBufferView = js.native
+}
+
+trait Algorithm extends js.Object {
+  var name: String = js.native
```
sjrd: Is `name` (and other fields in subclasses of `Algorithm`) really a `var`? Shouldn't it be a `val`?

src/main/scala/org/scalajs/dom/crypto/package.scala
```diff
@@ -0,0 +1,20 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.typedarray.{ArrayBufferView, ArrayBuffer, Uint8Array}
+
+package object crypto {
+  import scala.language.implicitConversions
```
sjrd: Put language imports at the top of the file, just after `package` (before any other import).

sjrd:
That's all.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: Algorithm is just a user provided dictionary: http://www.w3.org/TR/WebCryptoAPI/#algorithm-dictionary, so I believe `name` should be a `var`.

mseddon:
Thanks again for your patience :)  It's been nearly a year since I did any serious commits, so I'm a little rusty.

sjrd:
LGTM

PR #126, opened 2015-06-25
Fix WebGL signatures.

mseddon:
Fixes a few issues with the type signatures of some WebGL calls:

`polygonOffset` was `(Int, Int) => Unit`, rather than `(Double, Double) => Unit`
`lineWidth` was `(Int) => Unit`, rather than `(Double) => Unit`
`texParameterf` was `(Int, Int, Int) => Unit`, rather than `(Int, Int, Double) => Unit`

Also cleaned up some incorrect parameter documentation, and further clarified internalformat/format.

I also tried (uncommitted) to prevent some scaladoc warnings in this file of the form "Could not find any member to link for "Float32Array", but for some reason (possibly because it's external to scala-js-dom?) even fully qualified names still issue a warning.  If anyone has any idea how to handle this correctly, I'll combine that into this PR, too, since this file got a little shouty since the introduction of `scalajs.js.typedarray`.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/WebGL.scala
+++ b/src/main/scala/org/scalajs/dom/raw/WebGL.scala
@@ -1054,11 +1054,11 @@ class WebGLRenderingContext extends js.Object {
    *                 [[WebGLRenderingContext.TEXTURE_CUBE_MAP_NEGATIVE_X]], [[WebGLRenderingContext.TEXTURE_CUBE_MAP_POSITIVE_Y]], [[WebGLRenderingContext.TEXTURE_CUBE_MAP_NEGATIVE_Y]],
    *                 [[WebGLRenderingContext.TEXTURE_CUBE_MAP_POSITIVE_Z]], or [[WebGLRenderingContext.TEXTURE_CUBE_MAP_NEGATIVE_Z]]
    *  @param level  the mipmap level of detail.  0 is the base image.
-   *  @param internalformat  the format of the compressed data.
    *  @param xoffset the x texel offset into the texture image.
    *  @param yoffset the y texel offset into the texture image.
    *  @param width  the width of the texture image.
    *  @param height the height of the texture image.
+   *  @param format the format of the compressed image data
    *  @param data  the compressed image data.
    */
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/WebGL.scala
```diff
@@ -1054,7 +1054,6 @@ class WebGLRenderingContext extends js.Object {
    *                 [[WebGLRenderingContext.TEXTURE_CUBE_MAP_NEGATIVE_X]], [[WebGLRenderingContext.TEXTURE_CUBE_MAP_POSITIVE_Y]], [[WebGLRenderingContext.TEXTURE_CUBE_MAP_NEGATIVE_Y]],
    *                 [[WebGLRenderingContext.TEXTURE_CUBE_MAP_POSITIVE_Z]], or [[WebGLRenderingContext.TEXTURE_CUBE_MAP_NEGATIVE_Z]]
    *  @param level  the mipmap level of detail.  0 is the base image.
-   *  @param internalformat  the format of the compressed data.
```
sjrd: Shouldn't this be renamed to `format` instead of removing it?

sjrd: And in general, there seems to be many removals that should have been renamings instead.

sjrd:
OK for the actual signature changes. These should be in a different commit than the documentation fixes, since they have nothing to do with each other.
You can easily recreate your commits first with `git reset HEAD^` (removes the top commit but keeps its changes, in the working copy) then `git add -p` (allows you to stage parts of the working copy to create a commit).

src/main/scala/org/scalajs/dom/raw/WebGL.scala
mseddon: Whoops, I got a little carried away there...

sjrd:
LGTM

mseddon:
Any ideas on the scaladoc warnings, or should we just ignore them for now?

sjrd:
Ignore them for now. I think there is something that needs to be done about linking different ScalaDocs with each other, but I never really understood how that worked. Anyway it's worth a separate PR.

mseddon:
Alrighty. Ta!

PR #128, opened 2015-06-26
Fix WebGL clearDepth signature

mseddon:
One more silly signature issue.  Integer depths for a depth buffer clamped from 0..1 aren't very useful :)


sjrd:
LGTM

PR #129, opened 2015-06-26
Fix #127 - HTMLElements should be abstract classes

### LINKED ISSUE #127: Constructors don't work for DOM elements
I don't see a reason why DOM elements like `HTMLElement` are represented by classes and not traits or abstract classes.

Currently the compiler does not protect us from trying to create HTML elements with plain constructor but it won't work in runtime.

mseddon:
Fixes the HTML DOM elements.

Note- this needs to be done for SVG too, but will take a little longer, since I am less familiar with SVG and will need to consult the spec.  I can address that in this PR if necessary.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/Html.scala
+++ b/src/main/scala/org/scalajs/dom/raw/Html.scala
@@ -12,7 +12,7 @@ package org.scalajs.dom.raw
 
 import scala.scalajs.js
 
-class HTMLDocument extends Document {
+abstract class HTMLDocument extends Document {
 
   /**
    * Returns the title of the current document.
@@ -545,7 +545,7 @@ class HTMLDocument extends Document {
  *
  * MDN
  */
-class HTMLTableElement extends HTMLElement {
+abstract class HTMLTableElement extends HTMLElement {
   /**
    * Is a DOMString containing the length in pixels or in percentage of the desired width
    * fo the entire table. It reflects the obsolete width attribute.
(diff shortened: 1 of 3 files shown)


mseddon:
Scratch that, SVG appears to work the same, will update with SVG momentarily.  I also forgot to make Element abstract...

sjrd:
Just a confirmation: are you quite sure that none of these can be instantiated with `new`?

Oh and our commit message format when fixing an issue is "Fix #127: Make HTMLElements abstract classes."
So with a `: `, and with a message that tells: Do this on the codebase. Not what it should be (that's an appropriate title for an Issue, but not for the fix for that issue).

mseddon:
Ah, I didn't realise I needed the :. Regarding the new constructor- I believe so, but I'll perform a test before I update this PR (just a throwaway javascript gist that will attempt to construct each of the classes I'm labelling `abstract`), to be sure.

mseddon:
https://jsfiddle.net/g7tzd7L7/  - passing under firefox, chrome, opera and IE (win64).

mseddon:
https://jsfiddle.net/g7tzd7L7/2/ has my final test, covering DOM features in general.  I've elected to make anything that can not be constructed by any browser not-abstract (e.g. DocumentFragment can be constructed in Chrome and Opera, but I would discourage this, since it's non-portable, you should use `Document.createFragment`).

sjrd:
LGTM

Issue #121, opened 2015-05-31
Package org.scalajs.dom can't contain vanilla scala definitions


mseddon:
As revealed by #94.

`package object dom` extends `js.Any`, and as such it can't export the constructors for various javascript dictionaries.

As it stands, only MutationObserverInit and BlobPropertyBag have constructors, and with BlobPropertyBag only the *type* is being exported, so it's not being a problem. (though #117 would add a ClipboardEventInit constructor)

Currently my thoughts are either to:

 - Just accept that constructors can't go in the dom package, (keeping them in their defining package  for now) with a note in the documentation.
 - Rename this package object to org.scalajs.dom.globals or similar, and create a *new* package that copies in these definitions along with the constructors.

Sadly @sjrd's implicit method trick can only work if dom were a singleton object, not a package.

PR #130, opened 2015-06-26
Fix #94: Remove erroneous MutationObserverInit export

### LINKED ISSUE #94: MutationObserverInit object is not js.native
The companion object of `MutationObserver` is [a plain scala object](https://github.com/scala-js/scala-js-dom/blob/2d7c1e26687151a4ddab97dd27d4301831938a04/src/main/scala/org/scalajs/dom/raw/lib.scala#L3984), so it should not be `js.native` [here](https://github.com/scala-js/scala-js-dom/blob/2d7c1e26687151a4ddab97dd27d4301831938a04/src/main/scala/org/scalajs/dom/package.scala#L96)

I thought I create PR that simply changing to 
```scala
val MutationObserverInit: raw.MutationObserverInit.type = raw.MutationObserverInit
```
, but produces warning message
```
Members of traits, classes and objects extending js.Any may only contain members that call js.native. This will be enforced in 1.0.
```

mseddon:
Unfortuntely, due to #121 we can't sensibly export this object without a *massive* reworking of the entire `org.scalajs.dom` package.

Might as well at least remove it for now, given anyone using this export will get surprising runtime errors.

### DIFF
--- a/src/main/scala/org/scalajs/dom/package.scala
+++ b/src/main/scala/org/scalajs/dom/package.scala
@@ -113,7 +113,6 @@ package object dom extends Window with scalajs.js.GlobalScope {
   val MutationEvent: raw.MutationEvent.type = js.native
   type MutationObserver = raw.MutationObserver
   type MutationObserverInit = raw.MutationObserverInit
-  val MutationObserverInit: raw.MutationObserverInit.type = js.native
   type MutationRecord = raw.MutationRecord
 
   type NamedNodeMap = raw.NamedNodeMap


sjrd:
Are there any other such objects in that package object? If there are, they should be removed as well, preferably in the same commit so we can later track which should be added back, should we find a solution.

mseddon:
Good question.  I'll do a thorough check tomorrow and report back.  Watch this space :)

mseddon:
Confirmed- this appears to be the only one in here.  All other native exports are below js.Any.

sjrd:
LGTM then

PR #131, opened 2015-06-26
Fix #48: HTMLScriptElement.onload, mutable Document.body

### LINKED ISSUE #48: Need to add HTMLScriptElement.onload and make Document.body mutable
As (poorly) documented [here](https://developer.mozilla.org/en/docs/Web/API/HTMLScriptElement) and [here](https://developer.mozilla.org/en-US/docs/Web/API/Document.body). The port of `workbench` to scala-js uses it, and currently works around it via casts to `js.Dynamic`

If anyone wants to do it that'd be great, otherwise I'll leave this here so I don't forget for later.

mseddon:
Some miscellaneous fixes that have been hanging around for longer than necessary.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/Html.scala
+++ b/src/main/scala/org/scalajs/dom/raw/Html.scala
@@ -153,7 +153,7 @@ abstract class HTMLDocument extends Document {
    *
    * MDN
    */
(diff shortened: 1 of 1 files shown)


sjrd:
LGTM

PR #136, opened 2015-07-17
Fix Window.innerWidth to be a def

mseddon:
Also move it next to `innerHeight`, because it was randomly tucked away between some event handlers.


sjrd:
LGTM

PR #138, opened 2015-07-26
Fix #137: TimeRanges start/end type incorrect

### LINKED ISSUE #137: TimeRanges start and stop have wrong type (Int instead of Double)
Here: 

- https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/raw/lib.scala#L5081
- https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/raw/lib.scala#L5088

I am getting runtime errors:

> Uncaught scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: 67.57 is not an instance of java.lang.Integer

I think these values are measured in seconds that have been buffered.

mseddon:


sjrd:
LGTM

PR #140, opened 2015-07-28
Fix #139: PerformanceTiming attributes should be Double

### LINKED ISSUE #139: PerformanceTiming claims to return Int but they overflow
I was trying to use it but I systematically get the following error when I access any of its values: `XXXXX is not an instance of java.lang.Integer` in the browser. Taking the js.Dynamic route works fine.

Fails:
```scala
    val window = dom.document.defaultView
    val perf = window.performance.timing
    PageLoad(perf.loadEventStart - perf.navigationStart)
```

Works:
```scala
    val window = DOMGlobalScope.window
    val perf = window.performance.timing
    PageLoad(perf.loadEventStart.asInstanceOf[Double] - perf.navigationStart.asInstanceOf[Double])
```

mseddon:
Fun fact- these are defined in the w3c spec as `unsigned long long`...

https://dvcs.w3.org/hg/webperf/raw-file/tip/specs/NavigationTiming/Overview.html#sec-navigation-timing-interface

Either way, `Int` is too small and so `Double` is the only sensible type we can ascribe here.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -4357,7 +4357,7 @@ class PerformanceTiming extends js.Object {
    *
    * MDN
    */
-  def redirectStart: Int = js.native
+  def redirectStart: Double = js.native
 
   /**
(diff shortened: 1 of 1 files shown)


sjrd:
LGTM

PR #141, opened 2015-08-21
Add NonDocumentTypeChildNode

japgolly:


japgolly:
**Don't merge yet**, this doesn't work properly.

Using this, I get:
```
Uncaught TypeError: Cannot read property 'previousElementSibling' of null
```

So... `null` and `undefined` aren't the same in Scala.JS?

sjrd:
No, [`null` is `null` and `undefined` is `()`](http://www.scala-js.org/doc/js-interoperability.html) (unit). Imagine how much of everything would make no sense whatsoever if two values that are different in js were "the same" in Scala.js ... I can't even fathom the situation.

lihaoyi:
>  Imagine how much of everything would make no sense whatsoever

lol that's just every day in Javascript land...

japgolly:
Hehe. Well let's put it down to misremembering and wishingful thinking :)

Corrected. Should be good to go now.

sjrd:
LGTM

> lol that's just every day in Javascript land...

No no. Your snippet still makes sense. Here we're talking about an entirely different kind of nonsense. The kind which truly has *no* sense. As opposed to *some* sense, even if unintuitive and possibly difficult to reason about.

PR #142, opened 2015-08-22
Turn Promise trait into a class with executor constructor parameter.

vpavkin:
I write a js library, that converts scala Futures into javascript Promises. To instantiate a Promise within scala.js code it needs to be a class

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -471,7 +471,7 @@ trait WindowTimers extends WindowTimersExtension {
  *
  * MDN
  */
-trait Promise[+A] extends js.Object {
+class Promise[+A](executor: js.Function2[js.Function1[A, Any], js.Function1[Any, Any], Any]) extends js.Object {
 
   /**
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -471,7 +471,7 @@ trait WindowTimers extends WindowTimersExtension {
  *
  * MDN
  */
-trait Promise[+A] extends js.Object {
+class Promise[+A](executor: js.Function2[js.Function1[A, Unit], js.Function1[Any, Unit], Unit]) extends js.Object {
```
sjrd: Use `Any` or `_` as the result type of functions instead of `Unit`. It helps with inference in some cases. The caller of the function does not *demand* that a `Unit` be returned; it doesn't care what value is returned because it ignores it. So `Any` is more appropriate.

sjrd:
That's all. Please amend and force-push your commit to the PR branch, then comment on the PR to notify me.

vpavkin:
@sjrd Done! Thanks for feedback!

sjrd:
LGTM

PR #143, opened 2015-08-26
DataTransfer.setData() return type can be undefined

japgolly:
In latest Chrome I'm get:
```
Uncaught scala.scalajs.runtime.UndefinedBehaviorError:
An undefined behavior was detected: undefined is not an instance of java.lang.Boolean
```

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -3777,7 +3777,7 @@ trait DataTransfer extends js.Object {
    *
    * MDN
    */
-  def setData(format: String, data: String): Boolean = js.native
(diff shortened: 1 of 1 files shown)


sjrd:
Is it in the specification, or is it a chrome bug?

mseddon:
The current w3c editors draft defines setData as ` void setData(DOMString format, DOMString data);`.

http://www.w3.org/html/wg/drafts/html/master/editing.html#the-datatransfer-interface

sjrd:
If it says `void`, then it should be `Unit` in Scala.js, instead of `js.UndefOr[Boolean]`.

japgolly:
Ya, I just assumed scala-js-dom was incomplete, didn't think it might be wrong. Checking the spec, it is and it's more than just this method. Let me submit a new PR.

PR #144, opened 2015-08-26
Add new and correct existing fns in DataTransfer

japgolly:


sjrd:
What spec did you use? I see http://www.w3.org/html/wg/drafts/html/master/editing.html#the-datatransfer-interface and there are discrepancies.

japgolly:
https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer

sjrd:
Hum ... is there some evidence that `addElement` is supported by other browsers? Is this a Firefox-only thing? Why would it appear in MDN but not in the spec? Wand why, why doesn't MDN reference the spec it bases itself on, as it usually does?

Regardless, I would stick to `format` instead of `type` for the parameter names. Since there two contradicting names, let's keep the Scala-friendly one ;)

japgolly:
Woah, so in 2011 they dropped it from the spec (http://lists.w3.org/Archives/Public/public-webapps/2011OctDec/0171.html) but for some reason MDN/Firefox has kept it around.

Ok I've removed it changed `type` back to `format` as requested. Should be good now.

sjrd:
LGTM, thanks.

PR #145, opened 2015-08-31
Add constants for legal DragEffect values

japgolly:


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3730,6 +3730,46 @@ trait NavigatorGeolocation extends js.Object {
...
 
+/**
+ * When dragging, there are several operations that may be performed. The copy operation is
+ * used to indicate that the data being dragged will be copied from its present location to
+ * the drop location. The move operation is used to indicate that the data being dragged will
+ * be moved, and the link operation is used to indicate that some form of relationship or
+ * connection will be created between the source and drop locations.
+ *
+ * You can specify which of the three operations are allowed for a drag source by setting the
+ * `effectAllowed` property within a `dragstart` event listener.
```
sjrd: Could you wrap these docs at 80 characters, please? It's too long to read on mobile.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -3730,6 +3730,46 @@ trait NavigatorGeolocation extends js.Object {
...
+ *
+ * Note that these values must be used exactly as defined below.
+ *
+ * https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Drag_operations#drageffects
+ */
+object DataEffect {
+
+  /** no operation is permitted */
+  val None = "none"
```
sjrd: I recommend `final val`s here, so that scalac can constant-fold them away.

sjrd:
That's all.

japgolly:
No worries, I'll make both changes shortly.

About constant-folding, didn't you say a while back that it only applied to primitives? Is `String` an exception to that?

sjrd:
Yes, String is considered a "primitive". As well as `classOf[C]`, IIRC.
Actually it would be more accurate to say *literals* rather primitives. `10`, `5.4d`, `"hello"` and `classOf[String]` are all literals. But `5.toByte` isn't.

japgolly:
Ahhh that makes a lot of sense. After learning macros somewhat I imagine it's probably `Literal(Constant(…))` that are foldable? When you said primitives, I was thinking in the Java sense so no `AnyRef` subtypes which I thought excluded `java.lang.String`.

Anyhoo, changes are done. Please take another look.

(Also as requested, I trimmed the comments to line-length 80 but they're now obviously inconsistent with the rest of the file which seems to be around 94. But as you wish. :open_hands: )

sjrd:
LGTM

PR #146, opened 2015-08-31
* added the html5 Notification api 

coreyauger:
Added Notification API as outlined here (https://developer.mozilla.org/en-US/docs/Web/API/notification)

### DIFF
--- a/example/src/main/scala/example/Example.scala
+++ b/example/src/main/scala/example/Example.scala
@@ -173,3 +173,35 @@ object AjaxExtension {
     }
   }
 }
+
+
+@JSExport
+object Notification {
+  @JSExport
+  def main(in: html.Button) = {
+
+    def notifyMe() {
+      // Let's check whether notification permissions have already been granted
+      if (dom.window.Notification.permission == "granted") {
+        // If it's okay let's create a notification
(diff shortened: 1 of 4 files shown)


coreyauger:
I was thinking that it would be nice to beak out the notification options into an options trait .. rather then passing it in as a js.Object like I am doing now.  Thoughts?

mseddon:
Yep, I'd definitely do that if you can, with a companion object to construct it, see https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/raw/lib.scala#L3824

It looks like this is not supported under IE or mobile browsers, so it may also better live in the experimental package.  We use implicit conversions defined in there to overlay support for apis that won't "just work" everywhere, to avoid confusion, see: https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/experimental/Vibration.scala for an example.

mseddon:
I would also change the name of the commit to fall in line with convention.  It should read `Add the Notification API" - present tense, no leading *. (and "html5" is possibly redundant here).

coreyauger:
Sounds good... I am going to close this and issue a new pull request.  Thanks.

PR #148, opened 2015-09-01
Add the Notification API

coreyauger:


mseddon:
See also: #146 #147

example/src/main/scala/example/Example.scala
```diff
@@ -173,3 +173,36 @@ object AjaxExtension {
     }
   }
 }
+
+
+@JSExport
+object Notification {
+  @JSExport
```
sjrd: The `Example.scala` is for examples in the Readme. And we want to keep the readme short with only an introduction to scala-js-dom as a whole (not with an example of every possible sub-part of the API). So you should not change `Example.scala`. Especially since this example can found [on MDN](https://developer.mozilla.org/en-US/docs/Web/API/notification).

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.raw.EventTarget
+import scala.language.implicitConversions
+import scala.scalajs.js
+
+
```
sjrd: Double blank line here. We only use single blank lines.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+  /**
+   * The Notification interface allows to notify users of an incoming message
+   * or event
+   * NOTE: requires permission
+   * Note: This feature is available in Web Workers.
+   *
+   * MDN
+   */
+  class Notification extends EventTarget {
```
sjrd: Make this a top-level class with `@JSName("Notification.Notification")`.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+ */
+object Notification {
+  implicit def toNotification(w: org.scalajs.dom.Window): NotificationWindow =
+    w.asInstanceOf[NotificationWindow]
+
+  class NotificationWindow extends js.Any {
+    /** Return an interface to the Notification API */
+    def Notification: Notification = js.native
+
+  }
```
sjrd: This is not needed, nor is `toNotification`.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+     * The data read-only property of the Notification interface returns a
+     * structured clone of the notification's data, as specified in the
+     * data option of the Notification() constructor.
+     *
+     * The notification's data can be any arbitrary data that you want
+     * associated with the notification.
+     *
+     * MDN
+     */
+    var data: js.Object = js.native
```
sjrd: This should be a `val`, as well as many other properties in this class, according to MDN. (it's read-only)

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+  class Notification extends EventTarget {
+    /**
+     * Construct a new Notification
+     *
+     * @param title     The text title of the notification
+     * @param options   The options to configure this notification
+     * @return a new Notification
+     */
+    def this(title:js.UndefOr[String] = js.undefined,
+             options:js.UndefOr[NotificationOptions] = js.undefined) = this()
```
sjrd: This should be the primary constructor. Moreover, `title` should a `String` and non-optional.
`options` should directly be a `NotificationOptions`, with a default value of `???`.

sjrd: Also, always put a space after `:` (there are other occurrences of this in this PR, I think)

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+
+    /**
+     * The onclose property of the Notification interface specifies an event
+     * listener to receive close events. These events occur when a Notification
+     * is closed.
+     *
+     * MDN
+     */
+    @deprecated("This event handler is no longer listed in the Notifications API spec.", "forever")
+    var onclose: js.Function0[Unit] = js.native
```
sjrd: Don't introduce deprecated stuff. Just omit them.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+    var noscreen: Boolean = js.native
+
+    /**
+     * The onclick property of the Notification interface specifies an event
+     * listener to receive click events. These events occur when the user
+     * clicks on a displayed Notification.
+     *
+     * MDN
+     */
+    var onclick: js.Function0[Unit] = js.native
```
sjrd: Use `Any` instead of `Unit` for result types of `js.Function`s.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+   * @param sticky    Boolean indicating whether the notification should be
+   *                  'sticky', i.e. not easily clearable by the user.
+   * @param tag       A text identifying tag for the notification.
+   * @param title     The text title of the notification
+   * @param vibrate   The vibration pattern for hardware to emit
+   * @return a new NotificationOptions
+   */
+  @inline
+  def apply(body: js.UndefOr[String] = js.undefined,
+            dir: js.UndefOr[String] = js.undefined,
```
sjrd: Parameters should be indented 4 spaces wrt. `def`, so something like this:
```scala
  @inline
  def apply(
      body: ...,
      dir: ...,
      ...): NotificationOptions = {
```

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+   */
+  val body: String = js.native
+
+  /**
+   * The dir property of the Notification interface indicates the
+   * text direction of the notification.
+   *
+   * MDN
+   */
+  var dir: String = js.native
```
sjrd: I suggest to also make all of these `val`s instead of `var`s.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * The title property of the Notification interface indicates
+   * the title of the notification.
+   *
+   * MDN
+   */
+  var title: String = js.native
+
+  var vibrate: js.Object = js.native
```
sjrd: This should be a `js.Array[Double]`.

sjrd:
I think that's all I have.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
coreyauger: If I delete this I am not clear on how to get the Notification object from the window now..  I have tried 
```
object Notification extends js.Any{
  def Notification: Notification = js.native
}
```
But this does not work.

sjrd: You don't. You just use `new Notification(...)`

coreyauger: There is still a need to access the object "Notification" on window

`Notification.requestPermission`
as well as
`Notification.permission`
In the example on MSD
```
function notifyMe() {
  // Let's check if the browser supports notifications
  if (!("Notification" in window)) {
    alert("This browser does not support desktop notification");
  }

  // Let's check whether notification permissions have already been granted
  else if (Notification.permission === "granted") {
    // If it's okay let's create a notification
    var notification = new Notification("Hi there!");
  }

  // Otherwise, we need to ask the user for permission
  else if (Notification.permission !== 'denied') {
    Notification.requestPermission(function (permission) {
      // If the user accepts, let's create a notification
      if (permission === "granted") {
        var notification = new Notification("Hi there!");
      }
    });
  }

  // At last, if the user has denied notifications, and you 
  // want to be respectful there is no need to bother them any more.
}
```

Or am I missing something?
Thanks.

sjrd: Ah, right. For that you want this top-level object:
```scala
object Notification extends js.Object {
  def permission: String = js.native
  def requestPermission(...): ...
}
```

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+    var onshow: js.Function0[Unit] = js.native
+
+    /**
+     * The permission read-only property of the Notification interface
+     * indicates the current permission granted by the user for the current
+     * origin to display web notifications.
+     *
+     * MDN
+     */
+    var permission: String = js.native
```
sjrd: Which makes me realize this should not be there.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,389 @@
...
+    var vibrate: js.Array[Double] = js.native
+
+    /**
+     * The requestPermission() method of the Notification interface requests
+     * permission from the user for the current origin to display
+     * notifications.
+     *
+     * MDN
+     */
+    def requestPermission(callback:js.Function1[String, Unit]): Unit = js.native
```
sjrd: Nor this.
Both should be in the `object Notification`, definitely not as members of the class.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
coreyauger: cool thanks :+1:

example/src/main/scala/example/Example.scala
```diff
@@ -18,7 +18,7 @@ object NodeAppendChild {
   @JSExport
   def main(div: html.Div) = {
     val child = dom.document
-                   .createElement("div")
+      .createElement("div")
```
sjrd: Please revert all the changes to `Example.scala`. It should not be touched by this PR.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,348 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.raw.EventTarget
+import scala.language.implicitConversions
```
sjrd: This import is not necessary anymore.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,348 @@
...
+   * @param sticky    Boolean indicating whether the notification should be
+   *                  'sticky', i.e. not easily clearable by the user.
+   * @param tag       A text identifying tag for the notification.
+   * @param title     The text title of the notification
+   * @param vibrate   The vibration pattern for hardware to emit
+   * @return a new NotificationOptions
+   */
+  @inline
+  def apply(
+             body: js.UndefOr[String] = js.undefined,
```
sjrd: Please follow the style I indicated earlier: parameters are indented 4 spaces wrt. `def`.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,348 @@
...
+    result.asInstanceOf[NotificationOptions]
+  }
+}
+
+/**
+ * Implicit imports for the notification api.
+ *
+ * https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API
+ */
+object Notification extends js.Object{
```
sjrd: Space before `{`.

src/main/scala/org/scalajs/dom/experimental/Notification.scala
```diff
@@ -0,0 +1,348 @@
...
+ * Note: This feature is available in Web Workers.
+ *
+ * @param title     The text title of the notification
+ * @param options   The options to configure this notification
+ * @return a new Notification
+ * MDN
+ */
+@JSName("Notification")
+class Notification( title: String,
+                    options: NotificationOptions = ???) extends EventTarget {
```
sjrd: Style:
```scala
class Notification(title: String,
    options: NotificationOptions = ???) extends EventTarget {
```

sjrd:
That's all.

coreyauger:
doh!  .. thought I reverted example back to the original state.  What am I missing to get it back there?

sjrd:
Superseded by #152

PR #150, opened 2015-09-13
Fix Touch identifier to be Double

mseddon:
`identifier` on the`Touch`class is specified as a `long` http://www.w3.org/TR/touch-events/#idl-def-Touch.  I'm seeing `UndefinedBehaviourError` under iOS safari when the identifier is outside the range of an Int.


sjrd:
LGTM

PR #151, opened 2015-09-15
Ajax.get default value of inputdata should be null

chandu0101:
Ajax.get failing with `method Get  must not have a request body.` in android mobile  , looks like they checking data/body!=null .


sjrd:
Superseded by #154

chandu0101:
thank you :+1:

PR #156, opened 2015-09-27
Fix #121: Make the dom package object a Scala object, not a js.Any.

### LINKED ISSUE #121: Package org.scalajs.dom can't contain vanilla scala definitions

sjrd:
This is a pretty major breaking change, as now `dom` does not extend `Window` anymore. This means that selecting fields of the window through `dom.someField` is broken. `dom.window.someField` should be explicitly used instead.

We explicitly expose `document` as `dom.document`, because lots of examples out there use it this way, and it seems reasonable to keep that use case.

### DIFF
--- a/example/src/main/scala/example/Example.scala
+++ b/example/src/main/scala/example/Example.scala
@@ -9,7 +9,7 @@ object Alert {
   @JSExport
   def main() = {
     import org.scalajs.dom
-    dom.alert("Hi from Scala-js-dom")
+    dom.window.alert("Hi from Scala-js-dom")
   }
 }
 
@@ -49,10 +49,10 @@ object LocalStorage {
     val key = "my-key"
 
     in.value =
-      dom.localStorage.getItem(key)
+      dom.window.localStorage.getItem(key)
 
     in.onkeyup = (e: dom.Event) => {
-      dom.localStorage.setItem(
+      dom.window.localStorage.setItem(
         key, in.value
       )
       box.textContent =
@@ -94,7 +94,7 @@ object Base64 {
            out: html.Div) = {
     in.onkeyup = { (e: dom.Event) =>
       out.textContent =
-        dom.btoa(in.value)
+        dom.window.btoa(in.value)
     }
   }
 }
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @lihaoyi and basically everyone interested, since this is a large departure from before.

jducoeur:
The change seems logical to me -- I hadn't consciously realized that dom extends Window, and I don't *think* I'm using that formulation, but I'll survive the changes if needed...

lihaoyi:
What will be the new convention be? We used to use `dom.*` to refer to DOM types, and I always thought it was really nice to visually distinguish them from `js.*` types and Scala/Java types. Will we be telling people to use `import org.scalajs.dom.window._`? Is the purpose of this to allow non-DOM types to be accessed via `dom.*`?

jducoeur:
Wait a sec -- I thought the change was just the way you refer to the *objects* -- specifically, the Window object -- not the types...

sjrd:
There won't be any change of convention. It is still recommended to import `org.scalajs.dom` and refer to types as `dom.SomeType`. Only properties and methods of the `window` object need an explicit `dom.window.prop` instead of `dom.prop`.

lihaoyi:
Oh ok, LGTM then. I guess there are few enough properties and methods on window that it's no big deal calling `.window` to access each one

mseddon:
LGTM, it'll be nice to be able to finally put raw Scala types in dom :+1:
It doesn't seem to affect huge swathes of my own codebase, so hopefully the impact should be fairly minimal for most people.

PR #158, opened 2015-10-03
fix DomTokenList interface

spaced:
- toggle with force parameter added

see also https://dom.spec.whatwg.org/#interface-domtokenlist


sjrd:
`update` is the write pendent of `apply`. It is used for assignment of an element:
```scala
domList(i) = element
// desugars to (in Scala):
domList.update(i, element)
```
which, because of `@JSBracketAccess`, "translates" to (JS):
```js
domList[i] = element;
```

spaced:
oh, ok, sorry. will amend commit with toggle force flag only

sjrd:
LGTM. Sorry for the delay. (When you force-push or add a commit, it's a good idea to post a comment on the PR, otherwise I receive no notification.)

PR #162, opened 2015-10-29
Window: `pageXOffset` and `pageYOffset` are Double

tindzk:
Page offsets are not necessarily Integer values. The following exception was thrown when I tried to access `pageYOffset`:

```
java.lang.ClassCastException: 525.6000125312809 is not an instance of java.lang.Integer
```


sjrd:
LGTM

PR #165, opened 2015-11-05
Fix #163: Bump Scala.js to 0.6.5 and @js.native everything.

### LINKED ISSUE #163: Move to scala-js 0.6.5
0.6.5 adds `|`, which will enable tightening up a lot of the signatures, so I'd vote to move to it if possible.
On the other hand, if we're intentionally staying with 0.6.2 for backwards compatibility, then perhaps we shouldn't just yet.  Thoughts?

mseddon:
Horay, union types! :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,9 +11,9 @@ version := "0.9.0-SNAPSHOT"
 
 organization := "org.scala-js"
 
-scalaVersion := "2.11.6"
+scalaVersion := "2.11.7"
 
-crossScalaVersions := Seq("2.10.5", "2.11.6")
+crossScalaVersions := Seq("2.10.6", "2.11.7")
 
 scalacOptions ++= Seq("-deprecation", "-feature", "-Xfatal-warnings")
 
--- a/src/main/scala/org/scalajs/dom/experimental/Vibration.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/Vibration.scala
@@ -12,7 +12,8 @@ object Vibration {
   implicit def toVibration(n: org.scalajs.dom.Navigator): VibrationNavigator =
     n.asInstanceOf[VibrationNavigator]
 
-  class VibrationNavigator extends js.Any {
+  @js.native
+  trait VibrationNavigator extends js.Any {
     /** Vibrate the device for the specified number of milliseconds. */
     def vibrate(duration: Double): Boolean = js.native
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: scala
 script:
   - sbt ++$TRAVIS_SCALA_VERSION package doc
-  - sbt ++2.11.6 readme/run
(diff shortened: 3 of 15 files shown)


mseddon:
Okay that failure was travis's fault, but this one char' PR is also firing warnings everywhere on the omission of `@js.native`. Now I see why the version was held back :P

I'll fix this up fully tommorrow.

mseddon:
image

mseddon:
That's all, hopefully, appears to work against my projects at least.

sjrd:
> so I hoisted it for lack of a better solution.

I did not understand that.

mseddon:
Just adding a `@js.native` annotation to that class raised the error, so I moved the definition out of the Vibration object to toplevel, which kept the compiler happy.

sjrd:
Oh. It should be kept where it was, but make it a `trait` instead, if it's a pimp.

mseddon:
Ah, right! nice, I'll do that, then.  Particularly since this is what I point people to who want to add to `experimental` ^^

mseddon:
Squashed together with a less silly commit message, since the initial commit would leave us in a broken state.

mseddon:
Hmm, shall we update the scala versions to 2.10.6 and 2.11.7 while we're at it?

mseddon:
Just fixed a formatting issue with Vibration API.  Holding off library version change for now.
I'll tweak those too tomorrow, if you like.

sjrd:
> Hmm, shall we update the scala versions to 2.10.6 and 2.11.7 while we're at it?

Yes, you can do that as well.

sjrd:
> Fix #163: Bump scala-js to 0.6.5 and @js.native everything.

Please spell "Scala.js" the right way ^^

src/main/scala/org/scalajs/dom/experimental/Vibration.scala
```diff
@@ -24,4 +25,4 @@ object Vibration {
      */
     def vibrate(pattern: js.Array[Double]): Boolean = js.native
   }
-}
+}
```
sjrd: Missing EOL at EOF.

mseddon:
> Please spell "Scala.js" the right way ^^

*cough*. Whoops. I was thinking of the artifact :)

build.sbt
```diff
@@ -11,9 +11,9 @@ version := "0.9.0-SNAPSHOT"
 
 organization := "org.scala-js"
 
-scalaVersion := "2.11.6"
+scalaVersion := "2.11.7"
 
-crossScalaVersions := Seq("2.10.5", "2.11.6")
+crossScalaVersions := Seq("2.10.6", "2.11.7")
```
sjrd: You should also update `.travis.yml`.

sjrd:
That's all.

sjrd:
LGTM, thanks :)

Issue #166, opened 2015-11-06
generateKey API not flexible enough


bblfish:
The crypto api mapping is not general enough. The definition I see is:
 ```scala
def generateKey(algorithm: String, extractable: Boolean,
      keyUsages: js.Array[String]): Promise[js.Any] = js.native
```
the algorithm seems to also potentially be an object according to [this blog post](http://blog.engelke.com/tag/webcrypto/) where the example code is 
```JavaScript
window.crypto.subtle.generateKey(
            {
                name: "RSASSA-PKCS1-v1_5",
                modulusLength: 2048,
                publicExponent: new Uint8Array([1, 0, 1]),
                hash: {name: "SHA-256"}
            },
            true,   // Must extract private key to create PEM files later
            ["sign", "verify"]).
```

It's actually not that easy to read these JS specs, and  [WebCrypto API](http://www.w3.org/TR/WebCryptoAPI/#subtlecrypto-interface) is not exceptions. Still the definition there is 

```JavaScript
Promise<any> generateKey(AlgorithmIdentifier algorithm,
                          boolean extractable,
                          sequence<KeyUsage> keyUsages );
```

bblfish:
One thing that one could do is make algorithms typesafe, as for example I have been trying with the [Fetch API](https://github.com/scala-js/scala-js-dom/issues/167).

Here things may be a bit different. I was thinking in this case algorithms for example could not be sealed traits, but it seems according to [this stackoverflow article](http://stackoverflow.com/questions/28429928/web-cryptography-api-howto-discover-supported-algorithms) that it is not actually possible to discover supported algorithms in the browser.  Still I suppose developers could integrate libraries that support them and ship those with the code....

mseddon:
When I initially PR'd the web crypto api, @sjrd suggested that we could use path dependent types to actually provide accurate types for all the promises, so I think it might be worth mapping out the various routes through the api.  I believe we can produce accurate typing for the known algorithms, with a fallback to Promise[Any] when the user uses something we've not seen before.

bblfish:
That sounds like a good idea. The above changes get me far enough to continue programming my implementation of [http-signatures](https://tools.ietf.org/html/draft-cavage-http-signatures-05)

PR #168, opened 2015-11-07
fix for #166 generateKey API not flexible enough

bblfish:
with these changes it is possible to code the [WebCrypto API signing example](http://www.w3.org/TR/WebCryptoAPI/#examples-signing) in the same style as the example fixing #166 

```scala
    GlobalCrypto.crypto.subtle.generateKey(
      RsaHashedKeyGenParams(
        2048,
        new Uint8Array(js.Array[Short](1,0,1)), //65537
        HashAlgorithm.`SHA-256`),
      false,
      js.Array(KeyUsages.sign,KeyUsages.verify)
    )
```

### DIFF
--- a/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
+++ b/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
@@ -1,8 +1,16 @@
 package org.scalajs.dom.crypto
 
-import scalajs.js
+import java.math.{BigInteger => _}
+
 import org.scalajs.dom.raw._
+
+import scala.scalajs.js
+import scala.scalajs.js.Dynamic.literal
+import scala.scalajs.js.`|`
 import scala.scalajs.js.typedarray.ArrayBufferView
+
+import org.scalajs.dom.crypto._
+
 
 @js.native
 object GlobalCrypto extends js.GlobalScope {
@@ -41,6 +49,22 @@ trait Algorithm extends js.Object {
 
 @js.native
 trait KeyAlgorithm extends Algorithm
+
+@js.native
+trait HashAlgorithm extends Algorithm
+
+object HashAlgorithm {
+
+  private def named(name: String): HashAlgorithm = literal(name=name).asInstanceOf[HashAlgorithm]
+
+  val `SHA-1` = named("SHA-1")
+  val `SHA-256` = named("SHA-256")
+  val `SHA-384` = named("SHA-384")
+  val `SHA-512` = named("SHA-512")
+  //todo add others
+
+}
+
 
 @js.native
 trait CryptoKey extends js.Object {
@@ -116,7 +140,7 @@ trait SubtleCrypto extends js.Object {
    *
    * MDN
    */
-  def encrypt(algorithm: String, key: CryptoKey,
+  def encrypt(algorithm: AlgorithmIdentifier, key: CryptoKey,
       data: BufferSource): Promise[js.Any] = js.native
 
   /**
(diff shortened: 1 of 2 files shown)


mseddon:
Uh oh.  I notice that this PR is in branch `master` - so Sebastien won't merge like this. (merging PR's from master can cause issues later with the repo which have bitten us in the past.).  I've done this more times than I care to admit, so my quick fix for this is below :)

As far as I know with github, once a PR is in flight the source branch can't be changed (this is a bit of an oversight), so you'll have to do something like the following:

Rewind back the 2 commits from your local master branch, preserving changes:

`git reset --soft HEAD~2`

Check out a new branch:

`git checkout -b fix-generate-key-api` (or whatever you want to call it)

Recommit:

`git commit -am 'Fix #166: generateKeyApi not flexible enough.'`

Note: If the commit message is in the format "Fix #xyz: message" (Note the capital F, and colon), then github will also automatically close the issue when the PR is merged.

Finally push back to your fork:

`git push origin fix-generate-key-api`

Then close and re-open the PR from the `fix-generate-key-api` branch.  

I'm just refreshing my memory with the webcrypto spec, but we can pick this up in the other PR.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -43,6 +51,22 @@ trait Algorithm extends js.Object {
...
+
+object HashAlgorithm {
+
+  private def named(name: String): HashAlgorithm = literal(name=name).asInstanceOf[HashAlgorithm]
+
+  val `SHA-1` = named("SHA-1")
+  val `SHA-256` = named("SHA-256")
+  val `SHA-384` = named("SHA-384")
+  val `SHA-512` = named("SHA-512")
+  //todo add others
```
mseddon: Are there any other hash algorithms defined in the spec?

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -1,9 +1,17 @@
 package org.scalajs.dom.crypto
 
-import scalajs.js
+import java.math.{BigInteger => _}
+
 import org.scalajs.dom.raw._
+
+import scala.scalajs.js
+import scala.scalajs.js.Dynamic.literal
+import scala.scalajs.js.`|`
```
mseddon: This import of `|` is not used,

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -1,9 +1,17 @@
...
 import scala.scalajs.js.typedarray.ArrayBufferView
 
+import org.scalajs.dom.crypto._
```
mseddon: Redundant import of our own package.

mseddon:
Other than that (and the branch issue), LGTM.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
bblfish: Ah I am a cryptography user, not a builder, so I thought there may be more.

bblfish:
ok closing in order to re-open in a new branch. Thanks for the careful detailed help there.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: As far as I can see from the spec, the only valid cases where a HashAlgorithm will work, is if the algorithm supports the `digest` operation, which only covers the SHA-x family.

PR #169, opened 2015-11-08
Fix #166: generateKeyApi not flexible enough.

### LINKED ISSUE #166: generateKey API not flexible enough

bblfish:
With these changes it is possible to code the [WebCrypto API signing example](http://www.w3.org/TR/WebCryptoAPI/#examples-signing) in the same style as the example fixing #166 

```scala
    GlobalCrypto.crypto.subtle.generateKey(
      RsaHashedKeyGenParams(
        2048,
        new Uint8Array(js.Array[Short](1,0,1)), //65537
        HashAlgorithm.`SHA-256`),
      false,
      js.Array(KeyUsages.sign,KeyUsages.verify)
    )
```

### DIFF
--- a/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
+++ b/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
@@ -1,7 +1,8 @@
 package org.scalajs.dom.crypto
 
-import scalajs.js
 import org.scalajs.dom.raw._
+
+import scala.scalajs.js
 import scala.scalajs.js.typedarray.ArrayBufferView
 
 @js.native
@@ -39,8 +40,36 @@ trait Algorithm extends js.Object {
   var name: String = js.native
 }
 
+/**
+ * The KeyAlgorithm dictionary represents information about the contents of a
+ * given CryptoKey object.
+ *
+ * See [[http://www.w3.org/TR/WebCryptoAPI/#key-algorithm-dictionary ¶12 KeyAlgorithm dictionary]]
+ * in w3c spec.
+ */
 @js.native
 trait KeyAlgorithm extends Algorithm
+
+/**
+ * A HashAlgorithm type is not defined in the
+ * [[http://www.w3.org/TR/WebCryptoAPI/ W3C Web Crypto API]], even though
+ * a [[http://www.w3.org/TR/WebCryptoAPI/#key-algorithm-dictionary KeyAlgorithm dictionary]]
+ * type is. There are nevertheless a number of indications that HashAlgorithm's
+ * are a type of their own, as searching the spec will show.
+ */
+@js.native
+trait HashAlgorithm extends Algorithm
+
+object HashAlgorithm {
+
+  private def named(name: String): HashAlgorithm =
+    js.Dynamic.literal(name = name).asInstanceOf[HashAlgorithm]
+
+  val `SHA-1` = named("SHA-1")
+  val `SHA-256` = named("SHA-256")
+  val `SHA-384` = named("SHA-384")
+  val `SHA-512` = named("SHA-512")
+}
 
 @js.native
 trait CryptoKey extends js.Object {
@@ -102,6 +131,8 @@ trait JsonWebKey extends js.Object {
 }
 
 /**
+ * [[http://www.w3.org/TR/WebCryptoAPI/#subtlecrypto-interface w3c ¶14 Subtle Crytpo interface]]
+ *
  * The SubtleCrypto interface represents a set of cryptographic primitives.
  * It is available via the Crypto.subtle properties available in a window
  * context (via Window.crypto).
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -257,17 +278,17 @@ object RsaKeyGenParams {
...
 
 @js.native
 object RsaHashedKeyGenParams {
   @inline
-  def apply(name: String, modulusLength: Long,
-      publicExponent: BigInteger, hash: String): RsaHashedKeyGenParams = {
-    js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
+  def apply( modulusLength: Long,
+      publicExponent: BigInteger, hash: HashAlgorithmIdentifier): RsaHashedKeyGenParams = {
+    js.Dynamic.literal(name = "RSASSA-PKCS1-v1_5", modulusLength = modulusLength.toDouble,
```
bblfish: note here the removal of the `name` variable, and its replacement in body with a constant.  If you agree with this, then it one should do this with the other Algorithms. It would be really nice to be able to get out of this pattern where names are passed as strings with no type safety.

mseddon:
@sjrd already reviewed in #168, LGTM. WDYT?

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: I absolutely agree in the normal case this should be discouraged, but perhaps we should at least expose an overloaded apply with a name in case one is using an algorithm not covered by the library?

bblfish: Is it actually possible to do that?

mseddon: Yes, this is absolutely correct, actually.  Let's try and get rid of these magic constants where possible.

mseddon: Bleh- sorry, deleted and reposted my comment by mistake and now we're all out of sync.  I think you're doing the right thing here, so `RsaHashedKeyGenParams` is correct (no need for the name parameter).

One thing though- the `BigInteger` exponent might want to be overloaded.  For example I might legitimately have a `Uint8Array` from somewhere else and want to instance an `RsaHashedKeyGenParams`, so while the `BigInteger` is definitely convenient, the lower level `Uint8Array` should be kept, too.

bblfish: I have a feeling that you have a better mental model of the cryptography algorithms than I do. You could either submit a different pull request that aligns the others with this pattern, or make a pull request to my branch with those changes, that I can accept and that will then appear here. I don't feel very confident to do this with the other algorithms ( having ever only really used RSA ).

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -290,14 +311,16 @@ object RsaKeyAlgorithm {
 
 @js.native
 trait RsaHashedKeyAlgorithm extends RsaKeyAlgorithm {
-  var hash: KeyAlgorithm = js.native
+  var hash: HashAlgorithm = js.native
```
bblfish: I fixed this automatically. I wonder if they may not have a mistake in the spec though.
http://www.w3.org/TR/WebCryptoAPI/#RsaHashedKeyAlgorithm-dictionary

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: Okay, that's no problem. There isn't a lot of heavy crypto api users (I believe the list is currently, us. :)), and I still want to go through this api and see if I can strengthen all these Promise[Any]'s.  I will touch most of these classes anyway in that PR.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -305,14 +328,14 @@ object RsaHashedKeyAlgorithm {
 }
 
 @js.native
-trait RsaHashedImportParams extends js.Object {
-  var hash: String = js.native
+trait RsaHashedImportParams extends js.Object with HashAlgorithm {
+  var hash: HashAlgorithm = js.native
```
bblfish: I should probably change this to `HashAlgorithmIdentifier` as that is what is in the spec http://www.w3.org/TR/WebCryptoAPI/#RsaHashedImportParams-dictionary
( perhaps we should put links for each of these to the spec )
I only added a HashAlgorithmIdentifier in the last round of changes, and must have missed this one.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: I had a gut feeling about that line, too.  With the spec as it currently stands, `HashAlgorithm` is definitely an appropriate type.  We might have to weaken that due to future changes, but I think this for now is closer to the semantics.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -257,17 +278,17 @@ object RsaKeyGenParams {
 
 @js.native
 trait RsaHashedKeyGenParams extends RsaKeyGenParams {
-  var hash: String = js.native
+  var hash: HashAlgorithm = js.native
```
mseddon: http://www.w3.org/TR/WebCryptoAPI/#RsaHashedKeyGenParams-dictionary

I think this one wants to be `HashAlgorithmIdentifier`, too.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -305,14 +328,14 @@ object RsaHashedKeyAlgorithm {
...
+trait RsaHashedImportParams extends js.Object with HashAlgorithm {
+  var hash: HashAlgorithm = js.native
 }
 
 @js.native
 object RsaHashedImportParams {
   @inline
-  def apply(hash: String): RsaHashedImportParams =
+  def apply(hash: HashAlgorithm): RsaHashedImportParams =
     js.Dynamic.literal(hash = hash).asInstanceOf[RsaHashedImportParams]
```
bblfish: Note that when I change line 339 to `hash: HashAlgorithIdentifier` then I need to change the line below to `hash.asInsanceOf[js.Dynamic]` or it won't compile. You'll see this elswhere too.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: Actually, also `HashAlgorithm` extends `js.Object`, so just `extends HashAlgorithm`

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
bblfish: On monday I can ask on a number of w3c channels to see if we have misunderstood or if it is a bug.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: Huh. Yes, I see, seems to be that Scala.js doesn't think `HashAlgorithm | String` extends `js.Any`. This appears to be a deficiency in the union types. @sjrd thoughts?

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: That would be good.  I suspect it's just not as tight as it could be, but if we're off fundamentally, that would be nice to know :)

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
sjrd: Well since `String` is not a subtype of `js.Any`, it's normal that a union of String with something else is not either. Casting is the way to go, here, but cast to `js.Any`, not `js.Dynamic`

mseddon: Ah, right, of course that won't work.  Not a deficiency in union types, a deficiency in my thinking. :)

bblfish:
I pushed a little update to this PR with the following improvements:

1. changed `KeyUsages` to `KeyUsage` as per w3c spec
2. added some initial ScalaDoc pointing to w3c spec, which is the standard after all
3. fixed the `.asInstanceOf[js.Dynamic]` to `asInstanceOf[js.Any]`

That should be good enough for this PR I think.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -1,9 +1,15 @@
 package org.scalajs.dom.crypto
 
-import scalajs.js
+import java.math.{BigInteger => _}
```
sjrd: Why is this needed? `java.math.*` is not imported by default.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -43,6 +49,21 @@ trait Algorithm extends js.Object {
...
+  private def named(name: String): HashAlgorithm = literal(name=name).asInstanceOf[HashAlgorithm]
+
+  val `SHA-1` = named("SHA-1")
+  val `SHA-256` = named("SHA-256")
+  val `SHA-384` = named("SHA-384")
+  val `SHA-512` = named("SHA-512")
+
+}
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -43,6 +49,21 @@ trait Algorithm extends js.Object {
 trait KeyAlgorithm extends Algorithm
 
 @js.native
+trait HashAlgorithm extends Algorithm
+
+object HashAlgorithm {
+
+  private def named(name: String): HashAlgorithm = literal(name=name).asInstanceOf[HashAlgorithm]
```
sjrd: Try to keep line length under 80 characters. You can easily put the body on a separate line.
Also, add spaces around `=`.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -1,9 +1,15 @@
...
+import java.math.{BigInteger => _}
+
 import org.scalajs.dom.raw._
+
+import scala.scalajs.js
+import scala.scalajs.js.Dynamic.literal
 import scala.scalajs.js.typedarray.ArrayBufferView
 
+
+
```
sjrd: Triple blank line.

src/main/scala/org/scalajs/dom/crypto/package.scala
```diff
@@ -2,12 +2,16 @@ package org.scalajs.dom
...
+import scala.scalajs.js.`|`
+
 import scala.language.implicitConversions
 
 package object crypto {
   val crypto = GlobalCrypto.crypto
 
   type BigInteger = Uint8Array
+  type AlgorithmIdentifier = KeyAlgorithm | String
+  type HashAlgorithmIdentifier = HashAlgorithm | String
```
sjrd: According to http://www.w3.org/TR/WebCryptoAPI/#algorithm-dictionary, `HashAlgorithmIdentifier` is the same as `AlgorithmIdentifier`. Why is it not the case here?

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -725,16 +758,63 @@ trait Pbkdf2Params extends Algorithm {
...
+object KeyUsage {
+  val encrypt    = "encrypt".asInstanceOf[KeyUsage]
+  val decrypt    = "decrypt".asInstanceOf[KeyUsage]
+  val sign       = "sign".asInstanceOf[KeyUsage]
+  val verify     = "verify".asInstanceOf[KeyUsage]
+  val deriveKey  = "deriveKey".asInstanceOf[KeyUsage]
+  val deriveBits = "deriveBits".asInstanceOf[KeyUsage]
+  val wrapKey    = "wrapKey".asInstanceOf[KeyUsage]
+  val unwrapKey  = "unwrapKey".asInstanceOf[KeyUsage]
```
sjrd: Do not vertically align `=`s.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -725,16 +758,63 @@ trait Pbkdf2Params extends Algorithm {
...
-        hash = hash).asInstanceOf[Pbkdf2Params]
+        hash = hash.asInstanceOf[js.Any]).asInstanceOf[Pbkdf2Params]
   }
 }
+
+/**
+ * See [[http://www.w3.org/TR/WebCryptoAPI/#cryptokey-interface ¶ 13. CryptoKey Interface]] of w3c spec
+ */
+@js.native
+trait KeyUsage  extends js.Any
```
sjrd: Double space before `extends`.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -725,16 +758,63 @@ trait Pbkdf2Params extends Algorithm {
...
+/**
+ * see [[http://www.w3.org/TR/WebCryptoAPI/#cryptokey-interface ¶13 CryptoKey interface]] in W3C doc
+ */
+@js.native
+trait KeyType extends js.Any
+
+object KeyType {
+  val public    = "public".asInstanceOf[KeyType]
+  val `private` = "private".asInstanceOf[KeyType]
+  val secret    = "secret".asInstanceOf[KeyType]
```
sjrd: Do not vertically align.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -725,16 +758,63 @@ trait Pbkdf2Params extends Algorithm {
...
+
+object KeyFormat {
+  /** An unformatted sequence of bytes. Intended for secret keys. */
+  val raw   = "raw".asInstanceOf[KeyFormat]
+  /** The DER encoding of the PrivateKeyInfo structure from RFC 5208. */
+  val pkcs8 = "pkcs8".asInstanceOf[KeyFormat]
+  /** The DER encoding of the SubjectPublicKeyInfo structure from RFC 5280. */
+  val spki  = "spki".asInstanceOf[KeyFormat]
+  /** The key is a JsonWebKey dictionary encoded as a JavaScript object */
+  val jwk   = "jwk".asInstanceOf[KeyFormat]
```
sjrd: Do not vertically align.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -725,16 +758,63 @@ trait Pbkdf2Params extends Algorithm {
...
+
+}
```
sjrd: Missing EOL at EOF

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -725,16 +758,63 @@ trait Pbkdf2Params extends Algorithm {
...
 }
+
+/**
+ * See [[http://www.w3.org/TR/WebCryptoAPI/#cryptokey-interface ¶ 13. CryptoKey Interface]] of w3c spec
+ */
+@js.native
+trait KeyUsage  extends js.Any
+
+object KeyUsage {
+  val encrypt    = "encrypt".asInstanceOf[KeyUsage]
```
sjrd: Maybe these constants should start with an uppercase in Scala? I am not sure, here. It might be better to stick to the identifier version of the string they represent.

src/main/scala/org/scalajs/dom/crypto/package.scala
```diff
@@ -2,12 +2,16 @@ package org.scalajs.dom
 
 import scala.scalajs.js
 import scala.scalajs.js.typedarray.{ArrayBufferView, ArrayBuffer, Uint8Array}
+import scala.scalajs.js.`|`
```
sjrd: The backsticks are unnecessary.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -288,32 +311,42 @@ object RsaKeyAlgorithm {
...
+  /**
+   * Note that section [[http://www.w3.org/TR/WebCryptoAPI/#RsaHashedKeyAlgorithm-dictionary 20.6. RsaHashedKeyAlgorithm dictionary]]
+   * of the W3C documentation uses a KeyAlgorithm here, and not what seems more correct a
+   * HashAlgorithmIdentifier.
+   */
+  var hash: HashAlgorithmIdentifier = js.native
 }
 
+
+
```
sjrd: Triple blank line.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -257,17 +280,17 @@ object RsaKeyGenParams {
...
+  var hash: HashAlgorithmIdentifier = js.native
 }
 
 @js.native
 object RsaHashedKeyGenParams {
   @inline
-  def apply(name: String, modulusLength: Long,
-      publicExponent: BigInteger, hash: String): RsaHashedKeyGenParams = {
-    js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
+  def apply( modulusLength: Long,
```
sjrd: Spurious space after `(`

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
sjrd: In fact, this does not seem right. `HashAlgorithm` extends `Algorithm`, which has a `name` field. But a `String` does not have a `name` field. This cast is misleading. What is happening, here?

src/main/scala/org/scalajs/dom/crypto/package.scala
bblfish: I was thinking that one should use the type system of the JS spec as much as possible, but improve on it where it makes sense. JS developers and spec writers are perhaps not so used to working with a strong type system, and may be missing out from time to time on some nice semantic differences they could gain by being clearer.

For example their definition is this:

```
typedef (object or DOMString) AlgorithmIdentifier;
```

It seems a bit weird that they use `object` rather than `Algorithm` there for example. 

In this case there seems to be a clear distinction between a `KeyAlgorithm` and a `HashAlgorithm`. Even though they have the same structure ( they are both objects with a "name" field ), they seem to be non overlapping classes. If this is right then making the distinction in the type system would make it easier for the user to find the right types to use.

I could leave a note explaining the difference in the code, so that if at some point we do find it justified, then it would be easy to see where the difference came from.

sjrd: OK

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
bblfish: This is `js.Dynamic.literal(name=name)` which I think produces 
```JSON
{
  "name": "$name"
}
```

sjrd: Oh right. I completely misread the whole thing. Never mind that.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
bblfish: ok. I'll stick to it.

bblfish:
* added documentation pointing to the W3C spec.
* pointed out where we go a bit beyond the w3c spec in terms of type safety ( so that if I made a mistake it's easy for someone to work out where it may have come from )
* removed the white space
* changed `literal(...)` to `js.Dynamic.literal(...)` to keep code consistent with other uses

Hope that's good enough for this PR.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -112,106 +142,145 @@ trait JsonWebKey extends js.Object {
...
+   * Defined at
+   * [[http://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-verify ¶14.3.4 The verify method]]
+   *
+   * We are being a bit more precise than the official definition by requiring a KeyAlgorithmIdentifier
+   * rather than an AlgorithmIdentifier.
    */
-  def verify(algorithm: String, key: CryptoKey, signature: BufferSource,
-      data: BufferSource): Promise[js.Any] = js.native
+  def verify(algorithm: KeyAlgorithmIdentifier, key: CryptoKey,
+             signature: BufferSource, data: BufferSource): Promise[js.Any] = js.native
```
sjrd: Indentation issue.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -112,106 +142,145 @@ trait JsonWebKey extends js.Object {
...
-   * derivated from a master key and a specific algorithm given as parameters.
+   * derivated from a master key and a specific algorithm given as parameters. MDN
    *
-   * MDN
+   * Defined at
+   * [[http://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-deriveBits ¶14.3.8 The deriveBits method]]
+   *
+   * We are being a bit more precise than the official definition by requiring a KeyAlgorithmIdentifier
+   * rather than an AlgorithmIdentifier.
+
```
sjrd: Spurious blank line.

sjrd:
That's all. Could you also squash all the commits into a single commit, please?

bblfish:
done

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
bblfish: oops. I just realised which line you meant was spurious.

sjrd:
LGTM

bblfish:
Ok should do now. I removed the supeflous blank line.

bblfish:
Ok should do now, I really removed the blank line!

mseddon:
@bblfish Thanks!

PR #173, opened 2015-11-14
Fix #160: HTML escape scaladoc

### LINKED ISSUE #160: scaladoc needs to be properly html-escaped.
as reported by @ochrons via gitter, a trivial issue with an easy fix:

> 'for some reason in http://www.scala-js.org/api/scalajs-dom/0.8/#org.scalajs.dom.raw.Document the "comment" section is not closed and all the method names look big because of that.
Probably because the comment includes elements such as &lt;body&gt; and &lt;table&gt; making ScalaDoc confused'

Good to catch them all in a single PR (we don't use nested /** */ comments afaik so an easy regex search ought to work). Perhaps put a note in the `scala-js` project's `CONTRIBUTING.md`?

Something worth looking out for when reviewing PRs for now.

Scalastyle also has an `org.scalastyle.scalariform.ScalaDocChecker` plugin, and I'm not sure we're using it in `scala-js`, (or how good it is for catching this kind of error, or if it introduces more problems than it would solve...).  Eventually maybe we should maintain a duplicate of that task here.

mseddon:


sjrd:
I guess you've verified that the output HTML render appropriately?

mseddon:
Yes, I've skimmed through the scaladoc output- it appears to work properly.

sjrd:
LGTM, then, thanks!

PR #174, opened 2015-11-17
added CryptoKeyPair trait. Fix RsaHashedKeyGenParams

bblfish:
Implementing [HTTP-Signature](https://tools.ietf.org/html/draft-cavage-http-signatures-05) (see:   [KeyStore.scala](https://github.com/read-write-web/rww-scala-js/blob/7667bb1473fae1909c6503251e08f4a05287b577/src/main/scala/rww/store/KeyStore.scala)), I found the need for the CryptoKeyPair trait. 

During discussions to get that to work, I also discovered that a previous change of mine was mistaken.

And I found a documentation bug.

### DIFF
--- a/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
+++ b/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
@@ -71,15 +71,33 @@ object HashAlgorithm {
   val `SHA-512` = named("SHA-512")
 }
 
+/**
+ * The CryptoKey object represents an opaque reference to keying material that
+ * is managed by the user agent.
+ *
+ * defined at [[http://www.w3.org/TR/WebCryptoAPI/#cryptokey-interface ¶13 The CryptoKey Interface]]
+ */
 @js.native
 trait CryptoKey extends js.Object {
   val `type`: String = js.native
 
   val extractable: Boolean = js.native
 
-  val algorithm: js.Object = js.native
+  val algorithm: KeyAlgorithm = js.native
 
-  val usages: js.Object = js.native
+  val usages: js.Array[KeyUsage] = js.native
+}
+
+/**
+ * The CryptoKeyPair dictionary represents an asymmetric key pair that is comprised
+ * of both public and private keys.
+ * defined at [[http://www.w3.org/TR/WebCryptoAPI/#keypair ¶17 CryptoKeyPair dictionary]]
+ * of spec
+ */
+@js.native
+trait CryptoKeyPair extends js.Object {
+  val publicKey: CryptoKey = js.native
+  val privateKey: CryptoKey = js.native
 }
 
 @js.native
@@ -337,9 +355,9 @@ trait RsaHashedKeyGenParams extends RsaKeyGenParams {
 @js.native
 object RsaHashedKeyGenParams {
   @inline
-  def apply(modulusLength: Long,
+  def apply(name: String, modulusLength: Long,
       publicExponent: BigInteger, hash: HashAlgorithmIdentifier): RsaHashedKeyGenParams = {
-    js.Dynamic.literal(name = "RSASSA-PKCS1-v1_5", modulusLength = modulusLength.toDouble,
+    js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
         publicExponent = publicExponent,
         hash = hash.asInstanceOf[js.Any]).asInstanceOf[RsaHashedKeyGenParams]
   }
@@ -383,6 +401,13 @@ object RsaHashedKeyAlgorithm {
     js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
         publicExponent = publicExponent,
         hash = hash.asInstanceOf[js.Any]).asInstanceOf[RsaHashedKeyAlgorithm]
+  }
+
+  def `RSASSA-PKCS1-v1_5`(
+    modulusLength: Long,
+    publicExponent: BigInteger,
+    hash: HashAlgorithmIdentifier): RsaHashedKeyAlgorithm = {
+    apply("RSASSA-PKCS1-v1_5",modulusLength,publicExponent,hash)
   }
 }
 
@@ -866,3 +891,9 @@ object KeyFormat {
   /** The key is a JsonWebKey dictionary encoded as a JavaScript object */
   val jwk = "jwk".asInstanceOf[KeyFormat]
 }
+
+//
+// types defined in JSON Web Key (JWK) RFC
+// http://tools.ietf.org/html/rfc7517
+//
+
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -337,9 +355,9 @@ trait RsaHashedKeyGenParams extends RsaKeyGenParams {
 @js.native
 object RsaHashedKeyGenParams {
   @inline
-  def apply(modulusLength: Long,
+  def apply(name: String, modulusLength: Long,
       publicExponent: BigInteger, hash: HashAlgorithmIdentifier): RsaHashedKeyGenParams = {
-    js.Dynamic.literal(name = "RSASSA-PKCS1-v1_5", modulusLength = modulusLength.toDouble,
+    js.Dynamic.literal(name = name, modulusLength = modulusLength.toDouble,
```
bblfish: There are a number of RSA Hashed Protocols. I forget right now what the other one is.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -384,6 +402,13 @@ object RsaHashedKeyAlgorithm {
         publicExponent = publicExponent,
         hash = hash.asInstanceOf[js.Any]).asInstanceOf[RsaHashedKeyAlgorithm]
   }
+
+  def `RSASSA-PKCS1-v1_5`(
```
bblfish: This would be one way to make it easier to collect correct usages of algorithms, so as to make useage of the API much more straightforward for non crypto experts.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: `RSA-PSS` and `RSA-OAEP` also fit in that slot.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
mseddon: And save people from remembering where to insert an underscore, and where to insert a dash.  I think this is probably a good choice.

bblfish: are you ok with this way of doing things @sjrd ? I can add `RSA-PSS` and `RSA-OAEP` for this class, and leave the others to be filled in later.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -866,3 +891,9 @@ object KeyFormat {
   /** The key is a JsonWebKey dictionary encoded as a JavaScript object */
   val jwk = "jwk".asInstanceOf[KeyFormat]
 }
+
+//
+// types defined in JSON Web Key (JWK) RFC
+// http://tools.ietf.org/html/rfc7517
+//
```
bblfish: I have not filled this in, but there is some work here to be done to create classes for each of the `jwk` constructs, so that one does not to write dynamic code [as I did here](https://github.com/read-write-web/rww-scala-js/blob/7667bb1473fae1909c6503251e08f4a05287b577/src/main/scala/rww/store/KeyStore.scala#L164)

```Scala
def asTurtle(key: CryptoKey): raw.Promise[String] =
    GlobalCrypto.crypto.subtle.exportKey(KeyFormat.jwk, key).andThen ((x: js.Any) =>
    {
      import com.github.marklister.base64.Base64._
      val pk = x.asInstanceOf[Dictionary[String]]
      val ba = pk("n").toByteArray(base64Url)
      val modHex = BigInt(1,ba).toString(16)
      val exp = BigInt(pk("e").toByteArray(base64Url))
      s"""<#> cert:modulus "$modHex"^^xsd:hexBinary;
          |   cert:exponent $exp""".stripMargin
    }).asInstanceOf[raw.Promise[String]]
}
```

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
sjrd: Yes, I think that's reasonable.

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -866,3 +891,9 @@ object KeyFormat {
   /** The key is a JsonWebKey dictionary encoded as a JavaScript object */
   val jwk = "jwk".asInstanceOf[KeyFormat]
 }
+
+//
+// types defined in JSON Web Key (JWK) RFC
+// http://tools.ietf.org/html/rfc7517
```
sjrd: I don't understand the meaning of this comment. Is it a TODO? If yes, it should clearly mention "TODO" ;)

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -384,6 +402,13 @@ object RsaHashedKeyAlgorithm {
         publicExponent = publicExponent,
         hash = hash.asInstanceOf[js.Any]).asInstanceOf[RsaHashedKeyAlgorithm]
   }
+
+  def `RSASSA-PKCS1-v1_5`(
+    modulusLength: Long,
+    publicExponent: BigInteger,
+    hash: HashAlgorithmIdentifier): RsaHashedKeyAlgorithm = {
+    apply("RSASSA-PKCS1-v1_5",modulusLength,publicExponent,hash)
```
sjrd: Coding style: params should be indented 4 spaces wrt. `def` + spaces after `,`

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
bblfish: yes, is meant as a TODO. :-)

src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -866,3 +914,24 @@ object KeyFormat {
...
+// Todo: fill in the full list of types defined in JSON Web Key (JWK) RFC
+// http://tools.ietf.org/html/rfc7517
+//
+
+/**
+ * see example http://tools.ietf.org/html/rfc7517#appendix-A.1
+ * //todo: where is the specification of n and e?
+ */
+@js.native
+trait RSAPublicKey extends js.Object {
```
bblfish: made use of that here: https://github.com/read-write-web/rww-scala-js/blob/65afbf0f12ae809fe6bedd6b5db5b4961c98d29c/src/main/scala/rww/store/KeyStore.scala#L164-L163

bblfish:
I think this is ready to merge :-)

sjrd:
LGTM, I guess.

Issue #175, opened 2015-11-18
removeEventListener should take `js.Function1[_ <: Event, _]`


mpartel:
`EventTarget#addEventListener` takes a `js.Function1[T, _]` for any `T <: Event` but `EventTarget#removeEventListener` only accepts a `js.Function1[Event, _]`. It should probably also take a `js.Function1[T, _]`.

mseddon:
This has already been addressed in https://github.com/scala-js/scala-js-dom/commit/3c689a74ed7b7cc669d236ecedf53b9efc95329c

It should be released already, in `0.8.2`.

mpartel:
Ah, sorry, I thought my deps were more up to date than they actually were.

mseddon:
No problem :)

PR #176, opened 2015-11-25
Update readme to use the latest release of scala-js-dom

mseddon:
I suspect this out of date documentation caused #175 to be reported.

I'm unsure how to actually update the hosted pages (and won't have permissions anyway), so once merged can one of the admins ensure this is done?  Thanks.

### DIFF
--- a/readme/Index.scalatex
+++ b/readme/Index.scalatex
@@ -55,10 +55,10 @@
       Add the following to your sbt build definition:
 
     @hl.scala
-      libraryDependencies += "org.scala-js" %%% "scalajs-dom" % "0.8.1"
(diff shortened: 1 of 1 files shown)


sjrd:
LGTM

lihaoyi:
Since nobody's regenerated the docsite in 9 hours i'll go do it

sjrd:
Could we set up a Travis post-script that would do that automatically every time something is pushed on master?

lihaoyi:
We probably could; I've looked into it before and decided it wasn't worth the effort because it changes once in a blue moon, but if you want it's there https://gist.github.com/domenic/ec8b0fc8ab45f39403dd

sjrd:
#178 OK I'll do that eventually. Not in the coming 2 weeks, though.

PR #177, opened 2015-11-25
Fix #167: Add Fetch API + Streams API

### LINKED ISSUE #167: Fetch and Streaming API
I have a beginning of a Fetch and Streaming API 
https://github.com/read-write-web/rww-scala-js/tree/akka.js/src/main/scala/org/scalajs/dom/experimental

As I use it and make sure it works I'd be happy to add it to the repo here. I have been trying to work on getting as much typesafety as possible out of the [Fetch API](https://fetch.spec.whatwg.org/) spec.

I have some initial code demonstrating the working of the Fetch API. I have not yet seriously tested the streaming one yet. As it stabilises I'd be happy to contribute it back. Or else publish it seperately of the library I am building right now.

bblfish:
I have been using the Fetch API for my [rww-scala-js akka branch](https://github.com/read-write-web/rww-scala-js/blob/6a24b5ffdf62b1d07542a178a912882e844c7b7c/src/main/scala/rww/store/WebResourceActor.scala). The code that I have tested and used works. So I thought I'd push this over before someone goes through the same work again.

Probably still a lot to be improved.

This would close https://github.com/scala-js/scala-js-dom/issues/167

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/Fetch.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/Fetch.scala
@@ -1,0 +1,498 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.Blob
+import org.scalajs.dom.raw.{FormData, Promise}
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.{JSName, ScalaJSDefined}
+import scala.scalajs.js.dom.experimental.JSIterable
+import scala.scalajs.js.typedarray.{ArrayBuffer, Uint8Array}
+
+@js.native
+object Fetch extends js.GlobalScope {
+
+  /**
+   * [[https://fetch.spec.whatwg.org/#fetch-method ¶6.0 Fetch method]] defined in WhatWG spec
+   *
+   * @param info
+   * @param init
+   * @return
+   */
+  def fetch(info: RequestInfo, init: RequestInit = null): Promise[Response] = js.native
+}
+
+/**
+ * The Request interface of the Fetch API represents a resource request.  MDN
+ *
+ * see [[https://fetch.spec.whatwg.org/#request-class ¶6.3 Request Class]]
+ * in whatwg spec
+ *
+ * @param input the url of the requested resource or an unused Request object.
+ * @param init initialisation information
+ */
+@js.native
+class Request(input: RequestInfo, init: RequestInit = null) extends js.Object {
+
+  /**
+   * Contains the request's method (GET, POST, etc.)
+   */
+  def method: HttpMethod = js.native
+
+  @JSName("type")
+  def `mediaType`: RequestType = js.native
+
+  /**
+   * Contains the URL of the request.
+   */
+  def url: String = js.native //should be USVString
+
+  /**
+   * Contains the associated Headers object of the request.
+   */
+  def headers: Headers = js.native
+
+  def destination: RequestDestination = js.native
+
+  def referrer: String = js.native
+
+  def referrerPolicy: ReferrerPolicy = js.native
+
+  def mode: RequestMode = js.native
+
+  def credentials: RequestCredentials = js.native
+
+  def cache: RequestCache = js.native
+
+  def redirect: RequestRedirect = js.native
+
+  def integrity: String = js.native //should be DOMString
+}
+
+object RequestInit {
+
+  /**
+   * Create a RequestInit with some minimal typesafety on attributes.
+   *
+   * This method creates as light weight a RequestInit literal as possible
+   * from the provided parameters, so as to allow the browser to work through
+   * its default setting strategy.
+   *
+   * It is actually quite difficult to work out the defaults, which parameters
+   * work together and which don't. Check the rules here in case of doubt:
+   * - [[https://fetch.spec.whatwg.org/#requests ¶3.1.5 Requests]]
+   * - [[https://fetch.spec.whatwg.org/#request-class ¶6.3 Request class]], especially the constructor function
+   * of the whatwg Fetch spec.
+   *
+   * //todo: it would help a lot if there were a way to make this fully type safe
+   */
+  @inline
+  def apply(method: js.UndefOr[HttpMethod] = js.undefined,
+      headers: js.UndefOr[HeadersInit] = js.undefined,
+      body: js.UndefOr[BodyInit] = js.undefined,
+      referrer: js.UndefOr[String] = js.undefined, //should be USVString
+      referrerPolicy: js.UndefOr[ReferrerPolicy] = js.undefined,
+      mode: js.UndefOr[RequestMode] = js.undefined,
+      credentials: js.UndefOr[RequestCredentials] = js.undefined,
+      requestCache: js.UndefOr[RequestCache] = js.undefined,
+      requestRedirect: js.UndefOr[RequestRedirect] = js.undefined,
+      integrity: js.UndefOr[String] = js.undefined, //see [[https://w3c
+      // .github.io/webappsec-subresource-integrity/ integrity spec]]
+      window: js.UndefOr[Null] = js.undefined
+  ): RequestInit = {
+    val result = js.Dynamic.literal()
+
+    @inline
+    def set[T](attribute: String, value: js.UndefOr[T]) = value.foreach { x =>
+      result.updateDynamic(attribute)(x.asInstanceOf[js.Any])
+    }
+
+    set("method", method)
+    set("headers", headers)
+    set("body", body)
+    set("referrer", referrer)
+    set("referrerPolicy", referrerPolicy)
+    set("mode", mode)
+    set("credentials", credentials)
+    set("requestCache", requestCache)
+    set("requestRedirect", requestRedirect)
+    set("integrity", integrity)
+    set("window", window)
+
+    result.asInstanceOf[RequestInit]
+  }
+}
+
+/**
+ * See [[https://fetch.spec.whatwg.org/#requestinit RequestInit]] in Fetch API
+ * The underlying object is a dictionary. This trait is here to help encode
+ * the types.
+ */
+@ScalaJSDefined
+trait RequestInit extends js.Object {
+  var method: js.UndefOr[HttpMethod]
+
+  var headers: js.UndefOr[HeadersInit]
+
+  var body: js.UndefOr[BodyInit]
+
+  var referrer: js.UndefOr[String]
+
+  var referrerPolicy: js.UndefOr[ReferrerPolicy]
+
+  var mode: js.UndefOr[RequestMode]
+
+  var credentials: js.UndefOr[RequestCredentials]
+
+  var requestCache: js.UndefOr[RequestCache]
+
+  var requestRedirect: js.UndefOr[RequestRedirect]
+
+  var integrity: js.UndefOr[String]
+
+  /**
+   * The whatwg spec section on [[https://fetch.spec.whatwg.org/#requestinit RequestInit dictionary]]
+   * has a comment that states that this value "can only be set to null".
+   * In the detailed steps section for [[https://fetch.spec.whatwg.org/#dom-request the Request(input,init) constructor]]
+   * it says even more clearly:
+   * "If init's window member is present and it is not null, throw a TypeError."
+   */
+  var window: js.UndefOr[Null]
+}
+
+/**
+  * [[https://fetch.spec.whatwg.org/#response ¶6.4 Response class]] in whatwg spec
+  *
+  * @param content optional content
+  * @param init optional response initialisiton
+  */
+@js.native
+class Response(content: BodyInit = null, init: ResponseInit = null)
+    extends Body {
+
+  /** Contains the type of the response */
+  def `type`: ResponseType = js.native
+
+  /** Contains the URL of the response. */
+  def url: String = js.native
+
+  /**
+   * Contains a boolean stating whether the response was successful (status in
+   * the range 200-299) or not.
+   **/
+  def ok: Boolean = js.native
+
+  /** Contains the status code of the response (e.g., 200 for a success). */
+  val status: Int = js.native //actually returns unsigned short
+
+  /** Contains the status message corresponding to the status code (e.g., OK for 200). */
+  def statusText: ByteString = js.native //actually returns ByteString
+
+  /** Contains the Headers object associated with the response. */
+  val headers: Headers = js.native
+
+  val body: ReadableStream[Uint8Array] = js.native
+
+  override def clone(): Response = js.native
+}
+
+/**
+ * static methods associated with a Response object in
+ * [[https://fetch.spec.whatwg.org/#response ¶6.4 Response class]] of
+ * whatwg Fetch spec.
+ */
+@js.native
+object Response extends js.Object {
+  /**
+   * @return a new Response object associated with a network error.
+   */
+  def error(): Response = js.native
+
+  /**
+   * @param url
+   * @param status redirect status
+   * @return a new Response
+   */
+  def redirect(url: String, status: Int = 302): Response = js.native
+}
+
+/**
+ * See [[https://fetch.spec.whatwg.org/#response-class ¶6.4 Response class]]
+ * definition in whatwg Fetch spec.
+ *
+ */
+@ScalaJSDefined
+trait ResponseInit extends js.Object {
+    var status: Int
+    var statusText: ByteString
+    var headers: HeadersInit
+}
+
+object ResponseInit {
+  def apply(_status: Int = 200,
+      _statusText: ByteString = "OK",
+      _headers: HeadersInit = js.Dictionary[String]()): ResponseInit = new ResponseInit {
+    var status = _status
+    var statusText = _statusText
+    var headers = _headers
+  }
+}
+
+/**
+ * See [[https://fetch.spec.whatwg.org/#body body interface]] in whatwg Fetch spec.
+ *
+ * see also [[https://developer.mozilla.org/en-US/docs/Web/API/Body Body]] in MDN
+ */
+@js.native
+trait Body extends js.Object {
+
+  /**
+   * MDN: Contains a Boolean that indicates whether the body has been read.
+   */
+  def bodyUsed: Boolean = js.native
+
+  /**
+   * MDN: Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with an ArrayBuffer.
+   */
+  def arrayBuffer(): Promise[ArrayBuffer] = js.native
+
+  /**
+   * Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with a Blob.
+   */
+  def blob(): Promise[Blob] = js.native
+
+  /**
+   * Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with a FormData object.
+   */
+  def formData(): Promise[FormData] = js.native
+
+  /**
+   * Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with a JSON object.
+   * //todo: define the JSON type, and return a Promise[JSON] as per spec
+   */
+  def json(): Promise[js.Any] = js.native
+
+  /**
+   * Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with a USVString (text).
+   */
+  def text(): Promise[String] = js.native
+}
+
+/**
+ * [[https://fetch.spec.whatwg.org/#headers-class ¶6.1 Headers Class]] of whatwg spec
+ *
+ * The Headers interface of the Fetch API allows you to perform various actions
+ * on HTTP request and response headers. These actions include retrieving, setting,
+ * adding to, and removing. A Headers object has an associated header
+ * list, which is initially empty and consists of zero or more name and value pairs.
+ * You can add to this using methods like append() (see Examples.) In all methods of
+ * this interface, header names are matched by case-insensitive byte sequence.
+ *
+ * For security reasons, some headers can only be controller by the user agent.
+ * These headers include the forbidden header names  and forbidden response header
+ * names.
+ *
+ * A Headers object also has an associated guard, which takes a value of immutable,
+ * request, request-no-cors, response, or none. This affects whether the set(),
+ * delete(), and append() methods will mutate the header. For more information
+ * see Guard.
+ *
+ * You can retrieve a Headers object via the Request.headers and Response.headers
+ * properties, and create a new Headers object using the Headers.Headers() constructor.
+ *
+ * MDN
+ */
+@js.native
+class Headers(map: HeadersInit = js.Array[js.Array[String]]())
+    extends JSIterable[js.Array[ByteString]] {
+
+  /**
+   * The append() method of the Headers interface appends a new value onto an
+   * existing header inside a Headers object, or adds the header if it does
+   * not already exist.
+   *
+   * The difference between Headers.set and append() is that if the specified
+   * header already exists and accepts multiple values, Headers.set will overwrite
+   * the existing value with the new one, whereas append() will append the new
+   * value onto the end of the set of values.
+   */
+  def append(name: ByteString, value: ByteString): Unit = js.native
+
+  /**
+   * The set() method of the Headers interface sets a new value for an existing
+   * header inside a Headers object, or adds the header if it does not already exist.
+   *
+   * The difference between set() and Headers.append is that if the specified
+   * header already exists and accepts multiple values, set() overwrites the
+   * existing value with the new one, whereas Headers.append appends the new
+   * value to the end of the set of values.
+   *
+   * Not all headers can be set. See the list of forbidden header names in
+   * https://fetch.spec.whatwg.org/#terminology-headers
+   */
+  def set(name: ByteString, value: ByteString): Unit = js.native
+
+  /**
+   * The delete() method of the Headers interface deletes a header from the current Headers object.
+   */
+  def delete(name: ByteString): Unit = js.native
+
+  /**
+   * The get() method of the Headers interface returns the first value of a given header from within a Headers object
+   * . If the requested header doesn't exist in the Headers object, the call returns null.
+   *
+   * @param name a ByteString, ie. only ASCII chars
+   */
+  def get(name: ByteString): js.UndefOr[ByteString] = js.native
+
+  /**
+   * The getAll() method of the Headers interface returns an array of all the
+   * values of a header within a Headers object with a given name. If the
+   * requested header doesn't exist in the Headers object, it returns an empty array.
+   *
+   * @param name a ByteString, ie. only ASCII chars
+   */
+  def getAll(name: ByteString): Sequence[ByteString] = js.native
+
+  /**
+   * The has() method of the Headers interface returns a boolean stating whether
+   * a Headers object contains a certain header.
+   *
+   * @param name a ByteString, ie. only ASCII chars
+   */
+  def has(name: ByteString): Boolean = js.native
+}
+
+
+@js.native
+sealed trait ReferrerPolicy extends js.Any
+
+object ReferrerPolicy {
+  val empty = "".asInstanceOf[ReferrerPolicy]
+  val `no-referrer` = "no-referrer".asInstanceOf[ReferrerPolicy]
+  val `no-referrer-when-downgrade` = "no-referrer-when-downgrade".asInstanceOf[ReferrerPolicy]
+  val `origin-only` = "origin-only".asInstanceOf[ReferrerPolicy]
+  val `origin-when-cross-origin` = "origin-when-cross-origin".asInstanceOf[ReferrerPolicy]
+  val `unsafe-url` = "unsafe-url".asInstanceOf[ReferrerPolicy]
+}
+
+/**
+ * This is not typed in the Fetch API but it is easy to create the most common
+ * defaults.
+ */
+@js.native
+trait HttpMethod extends js.Any
+
+object HttpMethod {
+  val GET = "GET".asInstanceOf[HttpMethod]
+  val POST = "POST".asInstanceOf[HttpMethod]
+  val PUT = "PUT".asInstanceOf[HttpMethod]
+  val PATCH = "PATCH".asInstanceOf[HttpMethod]
+  val DELETE = "DELETE".asInstanceOf[HttpMethod]
+  val QUERY = "QUERY".asInstanceOf[HttpMethod]
+  val HEAD = "HEAD".asInstanceOf[HttpMethod]
+  val OPTIONS = "OPTIONS".asInstanceOf[HttpMethod]
+}
+
+/**
+ *  Fetch APIs [[https://fetch.spec.whatwg.org/#requesttype RequestType enum]]
+ */
+@js.native
+sealed trait RequestType extends js.Any
+
+object RequestType {
+  val empty = "".asInstanceOf[RequestType]
+  val audio = "audio".asInstanceOf[RequestType]
+  val font = "font".asInstanceOf[RequestType]
+  val image = "image".asInstanceOf[RequestType]
+  val script = "script".asInstanceOf[RequestType]
+  val style = "style".asInstanceOf[RequestType]
+  val track = "track".asInstanceOf[RequestType]
+  val video = "video".asInstanceOf[RequestType]
+}
+
+/**
+ * Fetch APIs [[https://fetch.spec.whatwg.org/#requestdestination RequestDestination enum]]
+ */
+@js.native
+sealed trait RequestDestination extends js.Any
+
+object RequestDestination {
+  val empty = "".asInstanceOf[RequestDestination]
+  val document = "document".asInstanceOf[RequestDestination]
+  val sharedworker = "sharedworker".asInstanceOf[RequestDestination]
+  val subresource = "subresource".asInstanceOf[RequestDestination]
+  val unknown = "unknown".asInstanceOf[RequestDestination]
+  val worker = "worker".asInstanceOf[RequestDestination]
+}
+
+/**
+ * Fetch API's [[https://fetch.spec.whatwg.org/#requestmode RequestMode enum]]
+ */
+@js.native
+sealed trait RequestMode extends js.Any
+
+object RequestMode {
+  val navigate = "navigate".asInstanceOf[RequestMode]
+  val `same-origin` = "same-origin".asInstanceOf[RequestMode]
+  val `no-cors` = "no-cors".asInstanceOf[RequestMode]
+  val cors = "cors".asInstanceOf[RequestMode]
+}
+
+/**
+ * Fetch APIs [[https://fetch.spec.whatwg.org/#requestcredentials RequestCredentials enum]]
+ */
+@js.native
+sealed trait RequestCredentials extends js.Any
+
+object RequestCredentials {
+  val omit = "omit".asInstanceOf[RequestCredentials]
+  val `same-origin` = "same-origin".asInstanceOf[RequestCredentials]
+  val include = "include".asInstanceOf[RequestCredentials]
+}
+
+/**
+ * Fetch APIs [[https://fetch.spec.whatwg.org/#requestcache RequestCache enum]]
+ */
+@js.native
+sealed trait RequestCache extends js.Any
+
+object RequestCache {
+  val default = "default".asInstanceOf[RequestCache]
+  val `no-store` = "no-store".asInstanceOf[RequestCache]
+  val reload = "reload".asInstanceOf[RequestCache]
+  val `no-cache` = "no-cache".asInstanceOf[RequestCache]
+  val `force-cache` = "force-cache".asInstanceOf[RequestCache]
+  val `only-if-cached` = "only-if-cached".asInstanceOf[RequestCache]
+}
+
+/**
+ * Fetch API's [[https://fetch.spec.whatwg.org/#requestredirect RequestRedirect enum]]
+ */
+@js.native
+sealed trait RequestRedirect extends js.Any
+
+object RequestRedirect {
+  val follow = "follow".asInstanceOf[RequestRedirect]
+  val error = "error".asInstanceOf[RequestRedirect]
+  val manual = "manual".asInstanceOf[RequestRedirect]
+}
+
+@js.native
+sealed trait ResponseType extends js.Any
+
+/**
+ * see [[https://fetch.spec.whatwg.org/#responsetype]] of whatwg Fetch spec
+ */
+object ResponseType {
+  val basic = "basic".asInstanceOf[ResponseType]
+  val cors = "cors".asInstanceOf[ResponseType]
+  val default = "default".asInstanceOf[ResponseType]
+  val error = "error".asInstanceOf[ResponseType]
+  val opaque = "opaque".asInstanceOf[ResponseType]
+  val opaqueredirect = "opaqueredirect".asInstanceOf[ResponseType]
+}
--- a/src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
@@ -1,0 +1,56 @@
+package scala.scalajs.js.dom.experimental
+
+import scala.scalajs.js
+import scala.scalajs.js.Dynamic.{global => g}
+import scala.scalajs.js.annotation.{ScalaJSDefined, JSBracketCall}
+
+/**
+ * See [[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols The iterable protocol]]
+ * on MDN
+ *
+ * todo: to be replaced by official version once
+ * https://github.com/scala-js/scala-js/issues/1141
+ * is completed
+ */
+@ScalaJSDefined
+trait JSIterable[A] extends js.Object
+
+@ScalaJSDefined
+trait JSIterator[+A] extends js.Object {
+  def next(): JSIterator.Entry[A]
+}
+
+object JSIterator {
+  private val iteratorSymbol = g.Symbol.iterator
+
+  @js.native
+  private trait IteratorMethodAccess extends js.Object {
+    @JSBracketCall
+    def bracketCall[A](method: js.Any)(): A
+  }
+
+  @ScalaJSDefined
+  trait Entry[+A] extends js.Object {
+    def done: Boolean
+    def value: A
+  }
+
+  def toIterator[A](it: JSIterator[A]): scala.collection.Iterator[A] = {
+    new scala.collection.Iterator[A]() {
+      private var nextEntry: Entry[A] = it.next()
+
+      def hasNext: Boolean = !nextEntry.done
+
+      def next(): A = {
+        val result = nextEntry.value
+        nextEntry = it.next()
+        result
+      }
+    }
+  }
+
+  implicit class IterableW[+A](it: JSIterable[A]) {
+    def iterator(): Iterator[A] =
+      toIterator(it.asInstanceOf[IteratorMethodAccess].bracketCall[JSIterator[A]](iteratorSymbol)())
+  }
+}
--- a/src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
@@ -1,0 +1,228 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.{MessagePort, raw}
+import org.scalajs.dom.webgl.RenderingContext
+
+import scala.scalajs.js
+import scala.scalajs.js.|
+
+//todo: move all ServiceWorker code together ( currently some code in raw )
+//todo: some of the ServiceWorker code in raw there could be typed a bit more precisely using some classes defined here
+
+@js.native
+sealed trait FrameType extends js.Any
+
+/**
+ * part of ServiceWorker
+ * [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#client-frametype ¶4.2.2 frameType]]
+ * of serviceWorker spec
+ */
+object FrameType {
+  /**
+   * The window client's global object's browsing context is an auxiliary browsing context.
+   */
+  val auxiliary = "auxiliary".asInstanceOf[FrameType]
+
+  /** The window client's global object's browsing context is a top-level browsing context. */
+  val `top-level` = "top-level".asInstanceOf[FrameType]
+
+  /** The window client's global object's browsing context is a nested browsing context. */
+  val nested = "nested".asInstanceOf[FrameType]
+
+  val none = "none".asInstanceOf[FrameType]
+}
+
+/**
+ * [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#client ¶4.2 Client]]
+ * of Service Workers 1 spec.
+ */
+@js.native
+trait Client extends js.Object {
+  /**
+   * The url attribute must return the context object's associated service worker
+   * client's serialized creation url.
+   */
+  def url: String = js.native
+
+  def frameType: FrameType = js.native
+
+  /**
+   * The id attribute must return its associated service worker client's id.
+   */
+  def id: String = js.native
+
+  /**
+   *
+   * @param message the spec says this is of type any (?!)
+   * @param transfer https://html.spec.whatwg.org/multipage/infrastructure.html#transferable-objects
+   *
+   */
+  def postMessage(message: Any, transfer: Sequence[Transferable] = null): Unit = js.native
+}
+
+/**
+ * see [[https://html.spec.whatwg.org/multipage/scripting.html#canvasproxy ¶4.12.4.1 Proxying canvases to workers]]
+ * in whatwg html spec.
+ */
+@js.native
+trait CanvasProxy extends js.Any {
+  def setContext(context: RenderingContext): Unit = js.native
+}
+
+/**
+ * See [[https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent FetchEvent]] on MDN
+ *
+ * [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#fetch-event-section ¶4.5 FetchEvent]]
+ * on whatwg ServiceWorker spec.
+ */
+@js.native
+class FetchEvent extends raw.Event {
+  /**
+   * @return Boolean that is true if the event was dispatched with the user's
+   *         intention for the page to reload, and false otherwise. Typically,
+   *         pressing the refresh button in a browser is a reload, while clicking a
+   *         link and pressing the back button is not.
+   */
+  def isReload: Boolean = js.native
+
+  /**
+   * @return the Request that triggered the event handler.
+   */
+  def request: Request = js.native
+
+  /**
+   * See [[https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith respondWith]]
+   * page on MDN.
+   *
+   * The respondWith() method of the FetchEvent interface is intended for
+   * containing code that generates custom responses to the requests coming
+   * from the controlled page. This code will resolve by returning a Response
+   * or network error to Fetch.
+   *
+   * Renderer-side security checks about tainting for cross-origin content are
+   * tied to the transparency (or opacity) of the Response body, not URLs. If
+   * the request is a top-level navigation and the return value is a Response
+   * whose type attribute is opaque (i.e. an opaque response body), a network
+   * error is returned to Fetch. The final URL of all successful (non
+   * network-error) responses is the requested URL.
+   */
+  def respondWith(promisedResponse: raw.Promise[Response]): Unit = js.native
+}
+
+object ServiceWorker {
+
+  /**
+   * typed methods to set event handlers on ServiceWorkers. See
+   * [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#service-worker-global-scope-activate-event ¶4.7 Events]]
+   * of whatwg ServiceWorker spec.
+   */
+  implicit class ServiceWorkerOpts(serviceWorker: raw.ServiceWorker) {
+
+    def addInstallListener(fn: ExtendableEvent => Unit) =
+      serviceWorker.addEventListener("install", fn)
+
+    def addActivateListener(fn: ExtendableEvent => Unit) =
+      serviceWorker.addEventListener("activate", fn)
+
+    def addFetchListener(fn: FetchEvent => Unit) =
+      serviceWorker.addEventListener("fetch", fn)
+
+    def addMessageListener(fn: MessageEvent => Unit) =
+      serviceWorker.addEventListener("message", fn)
+
+  }
+}
+
+/**
+ * See [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#extendable-event-interface ¶4.4 ExtendableEvent]]
+ * of whatwg ServiceWorker spec.
+ *
+ * An ExtendableEvent object has an associated extend lifetime promises (an
+ * array of promises). It is initially set to null.
+ */
+@js.native
+class ExtendableEvent extends raw.Event {
+  def waitUntil(promise: raw.Promise[Any]): Unit = js.native
+}
+
+/**
+ * See [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#extendablemessage-event-interface ¶4.6 ExtendableMessageEvent]]
+ * of whatwg ServiceWorker spec.
+ *
+ * Service workers define the extendable message event that extends the
+ * message event defined in HTML to allow extending the lifetime of the event.
+ *
+ * See also: [[https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerMessageEvent ServiceWorkerMessageEvent]] on MDN
+ */
+@js.native
+class MessageEvent extends ExtendableEvent {
+
+  /**
+   * Returns the event's data. It can be any data type.
+   */
+  val data: Any = js.native
+
+  /**
+   * Returns the origin of the service worker's environment settings object.
+   */
+  val origin: String = js.native //should be type DOMString
+
+  /**
+   * Represents, in server-sent events, the last event ID of the event source.
+   */
+  val lastEventId: String = js.native //should be String
+
+  /**
+   * @return a reference to the service worker that sent the message.
+   */
+  def source: Client | raw.ServiceWorker | MessagePort = js.native
+
+  /**
+   *  It represents the MessagePort array being sent, if any.
+   */
+  def ports: js.Array[MessagePort] = js.native
+}
+/**
+ * See [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#service-worker-obj ¶3.1 ServiceWorker]]
+ * of ServiceWorker whatwg spec.
+ */
+@js.native
+sealed trait ServiceWorkerState extends js.Any
+
+object ServiceWorkerState {
+  /**
+   * The service worker in this state is considered an installing worker.
+   * During this state, event.waitUntil(f) can be called inside the oninstall
+   * event handler to extend the life of the installing worker until the
+   * passed promise resolves successfully. This is primarily used to ensure
+   * that the service worker is not active until all of the core caches are
+   * populated.
+   */
+  val installing = "installing".asInstanceOf[ServiceWorkerState]
+
+  /**
+   * The service worker in this state is considered a waiting worker.
+   */
+  val installed = "installed".asInstanceOf[ServiceWorkerState]
+
+  /**
+   * The service worker in this state is considered an active worker. During
+   * this state, event.waitUntil(f) can be called inside the onactivate event
+   * handler to extend the life of the active worker until the passed promise
+   * resolves successfully. No functional events are dispatched until the
+   * state becomes activated.
+   */
+  val activating = "activating".asInstanceOf[ServiceWorkerState]
+
+  /**
+   * The service worker in this state is considered an active worker ready to
+   * handle functional events.
+   */
+  val activated = "activated".asInstanceOf[ServiceWorkerState]
+
+  /**
+   * A new service worker is replacing the current service worker, or the
+   * current service worker is being discarded due to an install failure.
+   */
+  val redundant = "redundant".asInstanceOf[ServiceWorkerState]
+}
--- a/src/main/scala/org/scalajs/dom/experimental/Stream.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/Stream.scala
@@ -1,0 +1,379 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.raw.Promise
+
+import scala.scalajs.js
+
+// the stream API is defined in https://streams.spec.whatwg.org/
+
+/**
+ * [[https://streams.spec.whatwg.org/#ws-state ¶4.2.4.3. get state]] of
+ * whatwg streams spec
+ */
+@js.native
+sealed trait WriteableState extends js.Any
+
+object WriteableState {
+
+  /**
+   * The stream’s internal queue is full; that is, the stream is exerting
+   * backpressure. Use .ready to be notified of when the pressure subsides.
+   */
+  val waiting = "waiting".asInstanceOf[WriteableState]
+
+  /**
+   * The stream’s internal queue is not full; call .write() until
+   * backpressure is exerted.
+   */
+  val writable = "writable".asInstanceOf[WriteableState]
+
+  /**
+   * The stream’s .close() method has been called, and a command to close is
+   * in the queue or being processed by the underlying sink; attempts to
+   * write will now fail. */
+  val closing = "closing".asInstanceOf[WriteableState]
+
+  /** The underlying sink has been closed; writing is no longer possible. */
+  val closed = "closed".asInstanceOf[WriteableState]
+
+  /** An error occurred interacting with the underlying sink or the stream
+   * has been aborted, so the stream is now
+   * dead. */
+  val errored = "errored".asInstanceOf[WriteableState]
+}
+
+/**
+ * [[https://streams.spec.whatwg.org/#ws-class ¶4.2. Class WritableStream]] of
+ * whatwg Stream spec
+ *
+ *
+ * todo: the constructor
+ * @tparam T Type of the Chunks to be written to the Stream
+ */
+@js.native
+trait WriteableStream[-T] extends js.Object {
+
+  /**
+   * The closed getter returns a promise that will be fulfilled when the
+   * stream becomes closed, or rejected if it ever errors.
+   * see [[https://streams.spec.whatwg.org/#ws-closed ¶4.2.4.1. get closed]]
+   */
+  val closed: Promise[WriteableStream[T]] = js.native
+
+  /**
+   * The ready getter returns a promise that will be fulfilled when the
+   * stream transitions away from the "waiting" state to any other state.
+   * Once the stream transitions back to "waiting", the getter will return a
+   * new promise that stays pending until the next state transition. In
+   * essence, this promise gives a signal as to when any backpressure has let
+   * up (or that the stream has been
+   * closed or errored).
+   *
+   * see [[https://streams.spec.whatwg.org/#ws-ready ¶4.2.4.2. get ready]]
+   * of whatwg streams spec.
+   */
+  val ready: Promise[WriteableStream[T]] = js.native
+
+  /**
+   * The state getter returns the state of the stream
+   * see [[https://streams.spec.whatwg.org/#ws-state ¶4.2.4.3. get state]]
+   * of whatwg streams spec
+   */
+  def state: WriteableState = js.native
+
+  /**
+   * The abort method signals that the producer can no longer successfully
+   * write to the stream and it should be immediately moved to an "errored"
+   * state, with any queued-up writes discarded. This will also execute any
+   * abort mechanism of the underlying sink.
+   * see [[https://streams.spec.whatwg.org/#ws-abort ¶4.2.4.4. abort(reason)]]
+   *
+   * @param reason spec specifies Any (!?)
+   */
+  def abort(reason: Any): Unit = js.native
+
+  /**
+   * The close method signals that the producer is done writing chunks to the
+   * stream and wishes to move the stream to a "closed" state. This queues an
+   * action to close the stream, such that once any currently queued-up
+   * writes complete, the close mechanism of the underlying sink will
+   * execute, releasing any held resources. In the meantime, the stream will
+   * be in a "closing" state.
+   *
+   * [[https://streams.spec.whatwg.org/#ws-close ¶4.2.4.5. close()]]
+   *
+   * @return a promise of this stream being closed
+   */
+  def close(): Promise[WriteableStream[T]] = js.native
+
+  /**
+   * The write method adds a write to the stream’s internal queue,
+   * instructing the stream to write the given chunk of data to the
+   * underlying sink once all other pending writes have finished successfully
+   * . It returns a promise that will be fulfilled or rejected depending on
+   * the success or failure of writing the chunk to the underlying sink. The
+   * impact of enqueuing this chunk will be immediately reflected in the
+   * stream’s state property; in particular, if the internal queue is now
+   * full according to the stream’s queuing strategy, the stream will exert
+   * backpressure by changing its state to "waiting".
+   *
+   * see [[https://streams.spec.whatwg.org/#ws-write 4.2.4.6. write(chunk)]]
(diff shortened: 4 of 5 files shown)


bblfish:
note: I am quite sure there will be improvements needed to the formatting. But before going through all that, I thought I'd first check if this is of interest. 

Btw. is there a script to get the formatting just right for this repo?

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
...
+
+
+/**
+  * to be replaced by official version once
+  * https://github.com/scala-js/scala-js/issues/1141
+  * is completed
+  * @tparam A
+  */
+@js.native
+trait JSIterable[A] extends js.Object {
```
bblfish: not sure actually where this should go. But it has proven useful in my code to be able to iterate through the headers.

bblfish:
Not sure what the travis failures are related to.

sjrd:
```
[error] /home/travis/build/scala-js/scala-js-dom/src/main/scala/org/scalajs/dom/experimental/package.scala:12:
  Classes, traits and objects inheriting from js.Any should be annotated with @js.native,
  unless they have @ScalaJSDefined. The default will switch to Scala.js-defined
  in the next major version of Scala.js.
[error] package object experimental extends js.GlobalScope {
[error]                ^
[error] one error found
[error] (root/compile:compile) Compilation failed
```

bblfish:
yes but when I add `@js.native` then I get

```sbt
> compile
[info] Compiling 6 Scala sources to /Volumes/Dev/Programming/ScalaJS/scala-js-dom/target/scala-2.11/classes...
[error] /Volumes/Dev/Programming/ScalaJS/scala-js-dom/src/main/scala/org/scalajs/dom/experimental/package.scala:13: expected start of definition
[error] package object experimental extends js.GlobalScope {
[error] ^
[error] one error found
``` 

what I don't understand is that [my own code](https://github.com/read-write-web/rww-scala-js/blob/c49494776048a2596e7b07c00969ef7f45fef75d/src/main/scala/org/scalajs/dom/experimental/package.scala#L11) seems to compile.

sjrd:
Yep, native package objects are sort of deprecated (https://github.com/scala-js/scala-js/issues/1892). This build fails because we have `-Xfatal-warnings`.

bblfish:
What should I do then?

sjrd:
Yes, we should change it.

Use a named object, like `@js.native object Fetch extends js.GlobalScope`.

sjrd:
If you absolutely want the `fetch` method visible at the top-level of the package (which I'm sure is a good idea), you can write a Scala forwarder.

bblfish:
Ok, it compiles now. I did not add fetch to the package object because I am not sure if that is necessary.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+  * defined at https://streams.spec.whatwg.org/#readable-stream
+  */
+@js.native
+trait ReadableStream extends js.Object {
+  /**
+    * The locked getter returns whether or not the readable stream is locked to a reader.
+    * throws scala.scalajs.js.TypeError  if the stream is not readable
+    */
```
bblfish: It would not `publishLocal` with `@throws`

bblfish:
I tried to get as close to the preferred formatting of this project in the last commit. I also took out some classes related to streaming to a separate file. I worked on this before my previous commits, so the code may not be as good as my last one. But it is quite a lot of code.

sjrd:
I don't have the bandwidth to process this these days. Could someone else review this? @mseddon? @nicolasstucki?

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
...
+  @JSBracketCall
+  def method[R](symbol: js.Any)(): R = js.native
+}
+
+@js.native
+trait JSIterator[+A] extends js.Object {
+  def next(): JSIterator.Entry[A] = js.native
+}
+
```
bblfish: I suppose this one is for @sjrd . I added this here as there is none available yet in the standard library. Is that ok?

sjrd: It would be better to submit this as a PR to scala-js/scala-js.

bblfish: Ok, I'll do that too.

mseddon:
@sjrd @bblfish I have a bit of free time tomorrow, so I can pick this up from then.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+
+}
+
+/**
+ * https://fetch.spec.whatwg.org/#body
+ * https://developer.mozilla.org/en-US/docs/Web/API/Body
+ *
+ * bblfish: not clear from the definition if this is a number of traits that
+ * may be implemented by the Response or not
+ */
```
mseddon: Unclear what you mean by this- certainly body is *a* trait that is implemented by Response with this signature.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+ * A Headers object also has an associated guard, which takes a value of immutable,
+ * request, request-no-cors, response, or none. This affects whether the set(),
+ * delete(), and append() methods will mutate the header. For more information
+ * see Guard.
+ *
+ * You can retrieve a Headers object via the Request.headers and Response.headers
+ * properties, and create a new Headers object using the Headers.Headers() constructor.
+ */
+@js.native
+class Headers(map: js.Dictionary[String] | Array[Array[String]] | Headers = Dictionary[String]())
```
mseddon: Slightly inconsistent usage of `js.Dictionary` and unprefixed `Dictionary` here.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+    *
+    * @return
+    */
+  def respondWith(promisedResponse: Promise[Response]): Unit = js.native
+}
+
+
+object RequestInit {
+  def apply(method: HttpMethod = HttpMethod.GET,
+    headers: Headers | js.Dictionary[String]=js.Dictionary[String](), //could also be sequence<sequence<ByteString>>
```
mseddon: WebIDL `sequence<T>` is `js.Array[T]` | JSIterable[T]`.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+   */
+  def get(name: String): UndefOr[String] = js.native
+
+  /**
+   * The getAll() method of the Headers interface returns an array of all the
+   * values of a header within a Headers object with a given name. If the
+   * requested header doesn't exist in the Headers object, it returns an empty array.
+   *
+   */
+  def getAll(name: String): js.Array[String] = js.native
```
mseddon: This is technically `sequence<T>`, although it might be worth avoiding that declaration for now. See below.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+  val headers: Headers | js.Dictionary[String],
+  val body: UndefOr[String],
+  val referrer: UndefOr[String],
+  val referrerPolicy: ReferrerPolicy,//=ReferrerPolicy.empty,
+  val mode: RequestMode,//=RequestMode.navigate,
+  val credentials: RequestCredentials,//=RequestCredentials.omit,
+  val requestCache: RequestCache,//=RequestCache.default,
+  val requestRedirect: RequestRedirect,//=RequestRedirect.follow,
+  val integrity: UndefOr[String], //=null, //should be DomString
+  val window: UndefOr[js.Any]
```
mseddon: window appears to be `"client"` - https://fetch.spec.whatwg.org/#dom-request

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+@js.native
+sealed trait HttpMethod extends js.Any
+
+object HttpMethod {
+  val GET = "GET".asInstanceOf[HttpMethod]
+  val PUT = "PUT".asInstanceOf[HttpMethod]
+  val PATCH = "PATCH".asInstanceOf[HttpMethod]
+  val DELETE = "DELETE".asInstanceOf[HttpMethod]
+  val QUERY = "QUERY".asInstanceOf[HttpMethod]
+}
```
mseddon: No `POST`?

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+
+@js.native
+sealed trait RequestMode extends js.Any
+
+object RequestMode {
+  val navigate = "navigate".asInstanceOf[RequestMode]
+  val `same-origin` = "same-origin".asInstanceOf[RequestMode]
+  val `no-cors` = "no-cors".asInstanceOf[RequestMode]
+  val cors = "cors".asInstanceOf[RequestMode]
+  /** see https://github.com/whatwg/fetch/issues/152 */
```
mseddon: Is it possible to *read* that value (since that issue states you can't *set* it from a script)? If not I would elect to remove it.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
+package scala.scalajs.js.collection
```
mseddon: In reference to the discussion below, I am obviously worried that `scala-js-dom` will clash with a future `scala-js` implementation currently.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
bblfish: oops :-) Would have spotted that quite soon...

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
bblfish: I could call it scala.scalajs.js.collection.tmp for the moment, and put a note to replace it with the official one for the next version of scala-js.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
bblfish: also missing `OPTIONS`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
mseddon: Maybe just pop it in `org.scalajs.dom.experimental` for now? If you're using `experimental` and something breaks, that is to be expected :)

bblfish: good idea.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,67 @@
...
+  def frameType: FrameType = js.native
+
+  /**
+   * The id attribute must return its associated service worker client's id.
+   * bblfish: not clear what the type is
+   */
+  def id: Any = js.native
+
+  //todo: missing the CanvasProxy type
+  //todo: is it really Seq ? it says `sequence` in the spec
```
mseddon: See my previous comment about IDL `sequence[T]`.  For now I think we can get away with `js.Array`.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,67 @@
...
+  //todo: The ArrayBuffer object is a lot more complex, see
+  //https://html.spec.whatwg.org/multipage/infrastructure.html#arraybuffer
```
mseddon: The `ArrayBuffer` (and corresponding subtypes) are found in Scala.js itself, in `scalajs.js.typedarray.ArrayBuffer`

bblfish: yep, thanks. I used that a few weeks later.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.crypto.BufferSource
+import org.scalajs.dom.raw.{FormData, Promise}
+import org.scalajs.dom.{Blob, Event}
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.ScalaJSDefined
+import scala.scalajs.js.typedarray.ArrayBuffer
+import scala.scalajs.js.{Dictionary, UndefOr, collection, |}
```
mseddon: It is better to write `|` as `` `|` `` in import declarations- if you don't IntelliJ's lexer stupidly derails and assumes a syntax error, and then breaks the rest of the file.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+
+@js.native
+object Fetch extends js.GlobalScope {
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+ *
+ * @param input the url of the requested resource or an unused Request object.
+ * @param init initialisation information
+ *
+ *
+ *             see [[https://fetch.spec.whatwg.org/#request-class ¶6.3 Request Class]] in whatwg spec
+ */
+@js.native
+class Request(input: String | Request, init: RequestInit|js.Dictionary[js.Any] = null) extends js.Object {
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
bblfish: They have fixed that now in the latest IntelliJ and latest plugin ( download from server perhaps ). I sent a bug report on that a few weeks ago.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+  /** Contains the status message corresponding to the status code (e.g., OK for 200). */
+  val statusText: String = js.native //actuall returns ByteString
+
+  /** Contains the Headers object associated with the response. */
+  val headers: Headers = js.native //the definition says SameObject, what does that mean?
+
+  /** @return a new Response object associated with a network error. */
+  def error(): Response = js.native
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a FormData object. */
+  def formData(): Promise[FormData] = js.native
+
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a JSON object. */
+  //todo: define the JSON type
+  // def json(): Promise[JSON] = js.native
+
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a USVString (text). */
+  def text(): Promise[String] = js.native
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+}
+
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+   * existing header inside a Headers object, or adds the header if it does
+   * not already exist.
+   *
+   * The difference between Headers.set and append() is that if the specified
+   * header already exists and accepts multiple values, Headers.set will overwrite
+   * the existing value with the new one, whereas append() will append the new
+   * value onto the end of the set of values.
+   */
+  def append(name: String, value: String): Unit = js.native
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+   *
+   */
+  def getAll(name: String): js.Array[String] = js.native
+
+  /**
+   * The has() method of the Headers interface returns a boolean stating whether
+   * a Headers object contains a certain header.
+   */
+  def has(name: String): Boolean = js.native
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import scala.scalajs.js
+import scala.scalajs.js.Dynamic.{global => g}
+import scala.scalajs.js.annotation.JSBracketCall
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+
+  implicit class IterableW[+A](it: JSIterable[A]) {
+    def iterator(): Iterator[A] = {
+      val jsIt = it.method[JSIterator[A]](iteratorSymbol)
+      toIterator(jsIt)
+    }
+  }
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
...
+
+}
+
```
mseddon: Double blank line.  There should be only one at the end.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,67 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.raw.MessagePort
+
+import scala.scalajs.js
+import scala.scalajs.js.|
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,67 @@
...
+  //todo: What really is sequence here?
+
+  /**
+   *
+   * @param message bblfish not actually clear what its type is
+   * @param transfer https://html.spec.whatwg.org/multipage/infrastructure.html#transferable-objects
+   *
+   */
+  def postMessage(message: js.Any, transfer: Seq[js.Array[Byte] | MessagePort]): Unit = js.native
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+
+  /**
+   * The abort method signals that the producer can no longer successfully
+   * write to the stream and it should be immediately moved to an "errored"
+   * state, with any queued-up writes discarded. This will also execute any
+   * abort mechanism of the underlying sink.
+   * @param reason bblfish: not really sure if it is as large as being of
+   *               type Any
+   */
+  def abort(reason: Any): Unit = js.native
```
mseddon: Would this at least be `js.Any` rather than `Any`? :)

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+
+  /**
+   * The cancel method cancels the stream, signaling a loss of interest in
+   * the stream by a consumer. The supplied reason argument will be given to
+   * the underlying source, which may or may not use it.
+   *
+   * @param reason the reason <- actually not what type this is
+   * @return a Promise, but not quite sure what it can contain
+   */
+  def cancel(reason: String): Promise[Any] = js.native
```
mseddon: `js.Any`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   * The way in which the piping process behaves under various error
+   * conditions can be customized with a number of passed options. It returns
+   * a promise that fulfills when the piping process completes successfully,
+   * or rejects if any errors were encountered. Piping a stream will
+   * generally lock it for the duration of the pipe, preventing any other
+   * consumer from acquiring a reader. This method is intentionally generic;
+   * it does not require that its this value be a ReadableStream object.
+   * @param dest
+   */
+  def pipeTo(dest: WriteableStream, options: Any = js.undefined)
```
mseddon: `js.Any`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   * stream ever errors.
+   */
+  def closed: Promise[ReadableStreamReader] = js.native
+
+  /**
+   * If the reader is active, the cancel method behaves the same as that for
+   * the associated stream. When done, it automatically releases the lock.
+   * @param reason - not actually sure what type of object this should be
+   */
+  def cancel(reason: Any): Promise[Any] = js.native //not actually sure what
```
mseddon: `js.Any`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   *
+   * The desiredSize getter returns the desired size to fill the controlled
+   * stream’s internal queue. It can be negative, if the queue is over-full.
+   * An underlying source should use this information to determine when and how
+   * to apply backpressure.
+   *
+   * @return the size of the strema - no idea if this actually is an int
+   */
+  def desiredSize: Int = js.native
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/package.scala
```diff
@@ -0,0 +1,12 @@
+package org.scalajs.dom
+
+
+import org.scalajs.dom.raw.Promise
+
+import scala.scalajs.js
+import scala.scalajs.js.|
```
mseddon: Again, I'd prefer `` `|` `` here if possible, thanks to IntelliJ.

src/main/scala/org/scalajs/dom/experimental/package.scala
```diff
@@ -0,0 +1,12 @@
+package org.scalajs.dom
+
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/package.scala
```diff
@@ -0,0 +1,12 @@
...
+import org.scalajs.dom.raw.Promise
+
+import scala.scalajs.js
+import scala.scalajs.js.|
+
+package object experimental {
+  type RequestInfo = String | Request
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   * a promise that fulfills when the piping process completes successfully,
+   * or rejects if any errors were encountered. Piping a stream will
+   * generally lock it for the duration of the pipe, preventing any other
+   * consumer from acquiring a reader. This method is intentionally generic;
+   * it does not require that its this value be a ReadableStream object.
+   * @param dest
+   */
+  def pipeTo(dest: WriteableStream, options: Any = js.undefined)
+
+
```
mseddon: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,67 @@
...
+   */
+  def url: String = js.native
+
+  def frameType: FrameType = js.native
+
+  /**
+   * The id attribute must return its associated service worker client's id.
+   * bblfish: not clear what the type is
+   */
+  def id: Any = js.native
```
mseddon: `js.Any`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   * the success or failure of writing the chunk to the underlying sink. The
+   * impact of enqueuing this chunk will be immediately reflected in the
+   * stream’s state property; in particular, if the internal queue is now
+   * full according to the stream’s queuing strategy, the stream will exert
+   * backpressure
+   * by changing its state to "waiting".
+   * @param chunk
+   * @return bblfish: not sure what the type of the promise returned is
+   */
+  def write(chunk: Array[Byte]): Promise[Any] = js.native
```
mseddon: `js.Any`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   * reader.
+   *
+   * This method is intentionally generic; it does not require that its this
+   * value be a ReadableStream object. It also does not require that its
+   * writable argument be a WritableStream instance, or that its readable
+   * argument be a ReadableStream instance.
+   */
+  def pipeThrough(
+    pair: (WriteableStream, ReadableStream),
+    options: Any = js.undefined
```
mseddon: `js.Any`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,317 @@
...
+   * The way in which the piping process behaves under various error
+   * conditions can be customized with a number
+   * of passed options. It returns a promise that fulfills when the piping
+   * process completes successfully, or rejects if any errors were
+   * encountered. Piping a stream will generally lock it for the duration of
+   * the pipe, preventing any other consumer from acquiring a reader. This
+   * method is intentionally generic; it does not require that its this
+   * value be a ReadableStream object.
+   */
+  def tee(): Array[ReadableStream] = js.native
```
mseddon: `js.Array[ReadableStream]`

src/main/scala/org/scalajs/dom/experimental/package.scala
bblfish: Btw. Do you have a setting to get the formatting just right for this project in IntelliJ. Can these be saved to disk for re-use?

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,371 @@
...
+object RequestCache {
+  val default = "default".asInstanceOf[RequestCache]
+  val `no-store` = "no-store".asInstanceOf[RequestCache]
+  val reload = "reload".asInstanceOf[RequestCache]
+  val `no-cache` = "no-cache".asInstanceOf[RequestCache]
+  val `force-cache` = "force-cache".asInstanceOf[RequestCache]
+  val `only-if-cached` = "only-if-cached".asInstanceOf[RequestCache]
+}
+
+
```
mseddon: Double blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
mseddon: `HEAD`, I suppose, too, if it's possible to send with the fetch api.

mseddon:
That's all.

bblfish:
Thanks for that work. That's give me the motivation I needed to go through the code in detail a second time :-)

src/main/scala/org/scalajs/dom/experimental/package.scala
mseddon: This particular problem is with IntelliJ's lexer, so no setting will fix it.  I tried to coax it to get the indentation in line, but it's not *quite* right.  If would like to help put a formatter together, I'd be happy to work with you on that.

mseddon:
No problem :) thanks for the work so far!

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
mseddon: Oh, that's good to know.  Perhaps leave it then

src/main/scala/org/scalajs/dom/experimental/package.scala
bblfish: I am using IntelliJ EAP (Minerva) IU-143.1015.7  and scala plugin 2.0.231 nightly build.
the above code displays nicely with it.

I suppose one needs to find out first if there is a special file one can save the formatting to. It would be really cool if one could place those in the project. Then people could help tweak that file for the project to get it as close as possible for each IDE to the standards required there.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
mseddon: Also, `js.Any`

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
sjrd: Actually, I disagree with the recommendation to use `js.Any`. There are virtually no valid reason to use things statically typed as `js.Any`s (it's about as uncommon as demanding `AnyRef`, basically).

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
sjrd: That said, I recommend *not* importing the other types: `Dictionary`, `UndefOr` and `collection`. The recommended best practice in Scala.js code is to import `scala.scalajs.js`, then use `js.Dictionary` and so on.

`|` is an exception because we want to use it as `A | B`, which only works if it's imported. But other members of `js` should never be directly imported.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
mseddon: Ah, alright then.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,487 @@
...
+    referrer: js.UndefOr[String] = js.undefined, //should be USVString
+    referrerPolicy: js.UndefOr[ReferrerPolicy] = js.undefined,
+    mode: js.UndefOr[RequestMode] = js.undefined,
+    credentials: js.UndefOr[RequestCredentials] = js.undefined,
+    requestCache: js.UndefOr[RequestCache] = js.undefined,
+    requestRedirect: js.UndefOr[RequestRedirect] = js.undefined,
+    integrity: js.UndefOr[String] = js.undefined,  //see [[https://w3c.github.io/webappsec-subresource-integrity/ integrity spec]]
+    window: js.UndefOr[js.Any] = js.undefined
+  ) = {
+    val result = js.Dynamic.literal()
```
bblfish: The above requires me to write

```scala
RequestInit(
      headers = any2undefOrA(js.Dictionary("Accept" -> rdfMimeTypes ).asInstanceOf[HeadersInit]),
      requestCache = RequestCache.reload,
      credentials = RequestCredentials.include //<- does not work if server's Access-Control-Allow-Origin is set to *
    )
```
will this be fixed by [Issue 2067: Better implicits for combinations of js.| and js.UndefOr?](https://github.com/scala-js/scala-js/issues/2067) ?

sjrd: > will this be fixed by Issue 2067: Better implicits for combinations of js.| and js.UndefOr? ?

Yes, it should.

bblfish:
Ok, I think this is as far as I should go for this PR. We have read through the relevant specs twice at least.

I have tested at least a part of the streaming API with this commit to [WebResourceActor.scala](https://github.com/read-write-web/rww-scala-js/blob/902e3161046ca45e41e0682b0c759cb932456a4e/src/main/scala/rww/store/WebResourceActor.scala#L237).

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,487 @@
...
+  /** Contains the status message corresponding to the status code (e.g., OK for 200). */
+  def statusText: ByteString = js.native //actually returns ByteString
+
+  /** Contains the Headers object associated with the response. */
+  val headers: Headers = js.native
+
+  val body: ReadableStream = js.native
+
+  override
+  def clone(): Response = js.native
```
mseddon: Unsure about the override on it's own line here- there's no explicit mention of this in the Scala.js coding style document, but looking at the compiler, it seems these are usually on the same line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,487 @@
...
+
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a Blob. */
+  def blob(): Promise[Blob] = js.native
+
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a FormData object. */
+  def formData(): Promise[FormData] = js.native
+
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a JSON object. */
+  //todo: define the JSON type
+  // def json(): Promise[JSON] = js.native
```
mseddon: Should we uncomment this and perhaps just assign `Any` for the initial release?

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * see [[https://html.spec.whatwg.org/multipage/scripting.html#canvasproxy ¶4.12.4.1 Proxying canvases to workers]]
+ * in whatwg html spec.
+ */
+@js.native
+trait CanvasProxy extends js.Any {
+  def setContext(context: RenderingContext): Unit = js.native
+}
```
mseddon: No newline at end of file.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,340 @@
...
+   * a promise that fulfills when the piping process completes successfully,
+   * or rejects if any errors were encountered. Piping a stream will
+   * generally lock it for the duration of the pipe, preventing any other
+   * consumer from acquiring a reader. This method is intentionally generic;
+   * it does not require that its this value be a ReadableStream object.
+   * @param dest
+   */
+  def pipeTo(dest: WriteableStream, options: Any = js.undefined)
+
+
```
mseddon: Double blank line here.

mseddon:
Other than that, LGTM.

mseddon:
Oh, I'd change the commit message to "Fix #167: Add Fetch API", too.  That way the issue will automatically close on merge.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,340 @@
...
+   * stream ever errors.
+   */
+  def closed: Promise[ReadableStreamReader] = js.native
+
+  /**
+   * If the reader is active, the cancel method behaves the same as that for
+   * the associated stream. When done, it automatically releases the lock.
+   * @param reason - not actually sure what type of object this should be
+   */
+  def cancel(reason: Any): Promise[Any] = js.native //not actually sure what the return type is here
```
bblfish: should reason here be  `js.Any`. My guess that's the only thing it can take right?

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
bblfish: ok, I'll return a `Promise[js.Any]`

bblfish:
Ok. done. Perhaps check that the use of `Any` is correct.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,339 @@
...
+/**
+ *
+ * See [[https://streams.spec.whatwg.org/#chunk ¶2 Model]] but mostly the
+ * examples in the whatwg streams spec
+ *
+ */
+@js.native
+trait Chunk extends js.Object {
+  /* The actual chunk, can be of any type, not just array byte apparently */
+  def value: js.Array[Byte] = js.native
```
bblfish: Mhh, I am having a bit of problems with this one still...

src/main/scala/org/scalajs/dom/experimental/Stream.scala
mseddon: as per sjrd's comment earlier, `Any` should be fine here.

mseddon:
LGTM I think- we've scoured that spec for as much information as possible and even contacted them directly with clarifications.  It's in `experimental` so we can fix other things up as people find them I'd say.  That was a big one, thanks for all that!

mseddon:
Oh, one suggestion though- would it be possible to put this inside `dom.experimental.fetch`? I'm a bit worried about `experimental` becoming a huge cluttered mess (and this also allows some consistency for importing any neccessary pimps and types with some precision according to a given spec.)

bblfish:
What about the code from the Streams API? Would you also want that over? Or in a different directory? Btw, the main dom project for some reason here has everything in one huge file.... Perhaps that should be a new PR: structure the code nicely into independent files, etc...

mseddon:
Hum, yes, we should probably have a separate `streams` package, too. (I presume it's conceivable you might want one but not the other).  I agree with the huge dom project thing.  We've sort of grown organically from very small beginnings when this wasn't such a problem, but I'm not sure what to do about it yet.  I think it's definitely something that needs to be discussed.

It has slightly less practical impact, in the sense that in theory (although it's not [true](https://github.com/scala-js/scala-js-dom/issues/180)), you can blindly import `dom` and not run the risk of accidentally importing something that is not compatible across all major browsers.

bblfish:
On the type of Chunk I asked a question on the Streams API. https://github.com/whatwg/streams/issues/409

Btw. I think there is stuff in raw that belongs to ServiceWorker, so that would be another task to move it all out...

mseddon:
Good catch, I'll add that to the list of candidates for migration. Thanks :)

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,348 @@
...
+ */
+@js.native
+trait Chunk extends js.Object {
+  /**
+   * The value of the chunk.
+   * It is not clear what the exact type of the value is.
+   * Warning: This was tested in one instance.
+   *    Please open issues if you find that other types can be returned too.
+   */
+  def value: Uint8Array = js.native
```
bblfish: Ok, so this is wrong. The type of a Chunk is actually open. [See issue 410](https://github.com/whatwg/streams/issues/410) in the Sreams API github. 

We could have 

```
@js.native 
trait Chunk[T] extends js.Object {
    def value: T = js.native
    ...
}
```

The problem is that apparently a stream can return different types of chunks at each go. Well that's ok, such a stream would Presumably be a `Stream[Any]`.  It would be nice if we could then type such Streams too, so that we could then type the stream returned by the Fetch api's `Response.body` method. See the [documentation question - issue 167](https://github.com/whatwg/fetch/issues/167) I posed on the Fetch API page

mseddon: Icky.  I guess `Chunk[T]` is as good as we can expect for now.

bblfish:
Ok, now that I have genericised Streams I think we're good.
I have tested the generics stream with https://github.com/read-write-web/rww-scala-js/commit/a7b0322e06869235ce64866308c22110c90fa0ff

We can re-org in a different PR. Otherwise I fear this one is never going to be finished.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+    body: js.UndefOr[BodyInit] = js.undefined,
+    referrer: js.UndefOr[String] = js.undefined, //should be USVString
+    referrerPolicy: js.UndefOr[ReferrerPolicy] = js.undefined,
+    mode: js.UndefOr[RequestMode] = js.undefined,
+    credentials: js.UndefOr[RequestCredentials] = js.undefined,
+    requestCache: js.UndefOr[RequestCache] = js.undefined,
+    requestRedirect: js.UndefOr[RequestRedirect] = js.undefined,
+    integrity: js.UndefOr[String] = js.undefined,  //see [[https://w3c.github.io/webappsec-subresource-integrity/ integrity spec]]
+    window: js.UndefOr[js.Any] = js.undefined
+  ) = {
```
sjrd: Public method should have an explicit result type.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+    set("headers",headers)
+    set("body",body )
+    set("referrer",referrer)
+    set("referrerPolicy",referrerPolicy)
+    set("mode", mode)
+    set("credentials",credentials)
+    set("requestCache",requestCache)
+    set("requestRedirect",requestRedirect)
+    set("integrity",integrity)
+    set("window",window)
```
sjrd: Missing spaces after `,`

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+    credentials: js.UndefOr[RequestCredentials] = js.undefined,
+    requestCache: js.UndefOr[RequestCache] = js.undefined,
+    requestRedirect: js.UndefOr[RequestRedirect] = js.undefined,
+    integrity: js.UndefOr[String] = js.undefined,  //see [[https://w3c.github.io/webappsec-subresource-integrity/ integrity spec]]
+    window: js.UndefOr[js.Any] = js.undefined
+  ) = {
+    val result = js.Dynamic.literal()
+
+    def set[T](attribute: String, value: js.UndefOr[T]) =
+      value.foreach{x=> result.updateDynamic(attribute)(value.asInstanceOf[js.Any])}
```
sjrd: `value.foreach { x => ... }`
(missing spaces)

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+    mode: js.UndefOr[RequestMode] = js.undefined,
+    credentials: js.UndefOr[RequestCredentials] = js.undefined,
+    requestCache: js.UndefOr[RequestCache] = js.undefined,
+    requestRedirect: js.UndefOr[RequestRedirect] = js.undefined,
+    integrity: js.UndefOr[String] = js.undefined,  //see [[https://w3c.github.io/webappsec-subresource-integrity/ integrity spec]]
+    window: js.UndefOr[js.Any] = js.undefined
+  ) = {
+    val result = js.Dynamic.literal()
+
+    def set[T](attribute: String, value: js.UndefOr[T]) =
```
sjrd: Public method should have an explicit result type.

sjrd: Mark this `@inline`, as well as the enclosing `apply` method.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+
+/**
+  * [[https://fetch.spec.whatwg.org/#response ¶6.4 Response class]] in whatwg spec
+  *
+  * @param content optional content
+  * @param init optional response initialisiton
+  */
+@js.native
+class Response(content: BodyInit = null, init: ResponseInit = null)
+  extends Body {
```
sjrd: 4 spaces before `extends`

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+ *
+ * @param status
+ * @param statusText
+ * @param headers
+ */
+@js.native
+class ResponseInit(
+  status: Int=200,
+  statusText: ByteString="OK",
+  headers: HeadersInit) extends js.Object
```
sjrd: 4 spaces inside parentheses

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+ * in whatwg Fetch spec.
+ *
+ * @param status
+ * @param statusText
+ * @param headers
+ */
+@js.native
+class ResponseInit(
+  status: Int=200,
+  statusText: ByteString="OK",
```
sjrd: Missing spaces around `=`

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+   * promise that resolves with a FormData object.
+   */
+  def formData(): Promise[FormData] = js.native
+
+  /**
+   * Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with a JSON object.
+   * //todo: define the JSON type, and return a Promise[JSON] as per spec
+   */
+   def json(): Promise[js.Any] = js.native
```
sjrd: Indentation issue.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+ * see Guard.
+ *
+ * You can retrieve a Headers object via the Request.headers and Response.headers
+ * properties, and create a new Headers object using the Headers.Headers() constructor.
+ *
+ * MDN
+ */
+@js.native
+class Headers(map: HeadersInit = js.Array[js.Array[String]]())
+  extends JSIterable[js.Array[ByteString]] {
```
sjrd: 4 spaces

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+}
+
+/**
+ * This is not typed in the Fetch API but it is easy to create the most common
+ * defaults.
+ */
+@js.native
+trait HttpMethod extends js.Any
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * todo: to be replaced by official version once
+ * https://github.com/scala-js/scala-js/issues/1141
+ * is completed
+ */
+@js.native
+trait JSIterable[A] extends js.Object {
+  @JSBracketCall
+  def method[R](symbol: js.Any)(): R = js.native
```
sjrd: This should definitely not be made public like that. It's essentially a hack just to access `Symbol.iterator`.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+        nextEntry = it.next()
+        result
+      }
+    }
+
+  implicit class IterableW[+A](it: JSIterable[A]) {
+    def iterator(): Iterator[A] = {
+      val jsIt = it.method[JSIterator[A]](iteratorSymbol)
+      toIterator(jsIt)
+    }
```
sjrd: Instead, write this as:
```scala
def iterator(): Iterator[A] = {
  @ScalaJSDefined
  trait IteratorMethodAccess extends js.Object {
    @JSBracketCall
    def bracketCall[A](method: js.Any)(): A
  }
  toIterator(it.asInstanceOf[IteratorMethodAccess].bracketCall[JSIterator[A]](iteratorSymbol)())
}
```

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+}
+
+@js.native
+trait JSIterator[+A] extends js.Object {
+  def next(): JSIterator.Entry[A] = js.native
+}
+
+object JSIterator {
+  val iteratorSymbol = g.Symbol.iterator
+  @js.native
```
sjrd: Blank line above this line.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+  def method[R](symbol: js.Any)(): R = js.native
+}
+
+@js.native
+trait JSIterator[+A] extends js.Object {
+  def next(): JSIterator.Entry[A] = js.native
+}
+
+object JSIterator {
+  val iteratorSymbol = g.Symbol.iterator
```
sjrd: `private`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+  @js.native
+  trait Entry[+A] extends js.Object {
+    val done: Boolean = js.native
+    val value: A = js.native
```
sjrd: These should be `def`s. There's nothing in the specification of iterators saying that the entries are immutable. Moreover, accessing `value` could throw with UBE, so it really shouldn't be a `val`.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  def toIterator[A](it: JSIterator[A]): scala.collection.Iterator[A] =
+    new scala.collection.Iterator[A]() {
+      var nextEntry: Entry[A] = it.next()
+
+      def hasNext: Boolean = ! nextEntry.done
```
sjrd: No space after `!`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+  val iteratorSymbol = g.Symbol.iterator
+  @js.native
+  trait Entry[+A] extends js.Object {
+    val done: Boolean = js.native
+    val value: A = js.native
+  }
+
+  def toIterator[A](it: JSIterator[A]): scala.collection.Iterator[A] =
+    new scala.collection.Iterator[A]() {
+      var nextEntry: Entry[A] = it.next()
```
sjrd: `private`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+
+object JSIterator {
+  val iteratorSymbol = g.Symbol.iterator
+  @js.native
+  trait Entry[+A] extends js.Object {
+    val done: Boolean = js.native
+    val value: A = js.native
+  }
+
+  def toIterator[A](it: JSIterator[A]): scala.collection.Iterator[A] =
```
sjrd: `{}` around the body of this method

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+}
+
+/**
+ * See [[https://fetch.spec.whatwg.org/#requestinit RequestInit]] in Fetch API
+ * The underlying object is a dictionary. This trait is here to help encode
+ * the types.
+ * //todo: should the fields be methods or vars?
+ */
+@js.native
+trait RequestInit extends js.Object {
```
sjrd: Consider declaring all traits as `@ScalaJSDefined` instead. `@js.native` traits have few reasons to exist any more.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,68 @@
...
+   */
+  def id: String = js.native
+
+  /**
+   *
+   * @param message the spec says this is of type any (?!)
+   * @param transfer https://html.spec.whatwg.org/multipage/infrastructure.html#transferable-objects
+   *
+   */
+  def postMessage(message: Any, transfer: sequence[Transferable]=null): Unit = js.native
```
sjrd: Spaces around `=`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+   * value be a ReadableStream object. It also does not require that its
+   * writable argument be a WritableStream instance, or that its readable
+   * argument be a ReadableStream instance.
+   *
+   * //todo: determine the type of options
+   */
+  def pipeThrough(
+    pair: (WriteableStream[T], ReadableStream[T]),
+    options: Any = js.undefined
+  ): ReadableStream[T] = js.native
```
sjrd: 4 spaces, and don't wrap before `): Re...`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+}
+
+/**
+ *
+ * See [[https://streams.spec.whatwg.org/#chunk ¶2 Model]] but mostly the
+ * examples in the whatwg streams spec
+ *
+ */
+@js.native
+trait Chunk[T] extends js.Object {
```
sjrd: `+T`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+ *
+ * The ReadableStreamController constructor cannot be used directly; it only
+ * works on a ReadableStream that is in the middle of being constructed.
+ *
+ * @param stream can be null
+ * @tparam T Type of the Chunks to be enqueued to the Stream
+
+ */
+@js.native
+class ReadableStreamController[T](stream: ReadableStream[T] = null) extends js.Object {
```
sjrd: `-T`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+ *
+ * The ReadableStreamReader constructor is generally not meant to be used directly;
+ * instead, a stream’s getReader() method should be used. This allows different
+ * classes of readable streams to vend different classes of readers without the
+ * consumer needing to know which goes with which.
+ *
+ * @tparam T Type of the Chunks returned by the Stream
+ */
+@js.native
+class ReadableStreamReader[T](stream: ReadableStream[T]) extends js.Object {
```
sjrd: `+T`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+   */
+  def write(chunk: Chunk[T]): Promise[Any] = js.native
+}
+
+/**
+ * defined at https://streams.spec.whatwg.org/#readable-stream
+ * @tparam T Type of the Chunks returned by the Stream
+ */
+@js.native
+trait ReadableStream[T] extends js.Object {
```
sjrd: `+T`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
sjrd: This might require `[U >: T]` with `WriteableStream[U]` and `ReadableStream[U]`.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+   *
+   * This method is intentionally generic; it does not require that its this
+   * value be a ReadableStream object. It also does not require that its
+   * writable argument be a WritableStream instance, or that its readable
+   * argument be a ReadableStream instance.
+   *
+   * //todo: determine the type of options
+   */
+  def pipeThrough(
+    pair: (WriteableStream[T], ReadableStream[T]),
```
sjrd: This must be a `js.Tuple2`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
sjrd: In fact I think this whole definition is not type-safe. From the doc, I think It should be:
```scala
  def pipeThrough[U](pair: js.Tuple2[WriteableStream[T], ReadableStream[U]],
      options: Any = js.undefined): ReadableStream[U]
```

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+   * conditions can be customized with a number of passed options. It returns
+   * a promise that fulfills when the piping process completes successfully,
+   * or rejects if any errors were encountered. Piping a stream will
+   * generally lock it for the duration of the pipe, preventing any other
+   * consumer from acquiring a reader. This method is intentionally generic;
+   * it does not require that its this value be a ReadableStream object.
+   *
+   * //todo: determine the type of options
+   */
+  def pipeTo(dest: WriteableStream[T], options: Any = js.undefined)
```
sjrd: `: Unit`

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+   * The way in which the piping process behaves under various error
+   * conditions can be customized with a number
+   * of passed options. It returns a promise that fulfills when the piping
+   * process completes successfully, or rejects if any errors were
+   * encountered. Piping a stream will generally lock it for the duration of
+   * the pipe, preventing any other consumer from acquiring a reader. This
+   * method is intentionally generic; it does not require that its this
+   * value be a ReadableStream object.
+   */
+  def tee(): js.Array[ReadableStream[T]] = js.native
```
sjrd: The doc mentions a "given writable" stream", but the signature doesn't. That's suspicious.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+/**
+ * [[https://streams.spec.whatwg.org/#ws-class ¶4.2. Class WritableStream]] of
+ * whatwg Stream spec
+ *
+ *
+ * todo: the constructor
+ * @tparam T Type of the Chunks to be written to the Stream
+ */
+@js.native
+trait WriteableStream[T] extends js.Object {
```
sjrd: `-T`

src/main/scala/org/scalajs/dom/experimental/package.scala
```diff
@@ -0,0 +1,55 @@
...
+   * in whatwg Fetch spec
+   */
+  type BodyInit = Blob | BufferSource | FormData | String //todo: add URLSearchParams
+
+  /**
+   * WebIDL sequence<T> is js.Array[T] | JSIterable[T]. However @mseddon knows
+   * at least Blink's IDL compiler treats these as simply js.Array[T] for now.
+   * We keep this type as a reminder to check in more detail
+   */
+  type sequence[T] = js.Array[T]
```
sjrd: Uppercase `S`.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,368 @@
...
+ * See [[https://streams.spec.whatwg.org/#chunk ¶2 Model]] but mostly the
+ * examples in the whatwg streams spec
+ *
+ */
+@js.native
+trait Chunk[T] extends js.Object {
+  /**
+   * The value of the chunk.
+   */
+  def value: T  = js.native
```
sjrd: Double space

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+import scala.scalajs.js.Dynamic.{global => g}
+import scala.scalajs.js.annotation.JSBracketCall
+
+/**
+ * todo: to be replaced by official version once
+ * https://github.com/scala-js/scala-js/issues/1141
+ * is completed
+ */
+@js.native
+trait JSIterable[A] extends js.Object {
```
sjrd: This one should definitely be `@ScalaJSDefined`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+  @JSBracketCall
+  def method[R](symbol: js.Any)(): R = js.native
+}
+
+@js.native
+trait JSIterator[+A] extends js.Object {
```
sjrd: This one should definitely be `@ScalaJSDefined`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,49 @@
...
+  def next(): JSIterator.Entry[A] = js.native
+}
+
+object JSIterator {
+  val iteratorSymbol = g.Symbol.iterator
+  @js.native
+  trait Entry[+A] extends js.Object {
```
sjrd: This one should definitely be `@ScalaJSDefined`

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+   * @return Boolean that is true if the event was dispatched with the user's
+   *         intention for the page to reload, and false otherwise. Typically,
+   *         pressing the refresh button in a browser is a reload, while clicking a
+   *         link and pressing the back button is not.
+   */
+  def isReload: Boolean = js.native
+
+  /**
+    * @return the Request that triggered the event handler.
+    */
```
sjrd: Scaladoc style issue

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+  def request: Request = js.native
+
+  /**
+    *
+    * @return
+    */
```
sjrd: Scaladoc style issue

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+   *
+   * @param name a ByteString, ie. only ASCII chars
+   */
+  def getAll(name: ByteString): sequence[ByteString] = js.native
+
+  /**
+   * The has() method of the Headers interface returns a boolean stating whether
+   * a Headers object contains a certain header.
+   *
+
```
sjrd: Spurious blank line.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,498 @@
...
+@js.native
+trait Body extends js.Object {
+
+  /**
+   * MDN: Contains a Boolean that indicates whether the body has been read.
+   */
+  def bodyUsed: Boolean = js.native
+
+  /** MDN: Takes a Response stream and reads it to completion. It returns a
+   * promise that resolves with an ArrayBuffer. */
```
sjrd: `*/` on the next line.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
bblfish: can a `@ScalaJSDefined` trait contain `js.native` methods?

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
bblfish: I suppose you mean I should remove then the `trait JSIterable[A]` defined above line 19. But then I no longer can have a class implement it such as as [the class Headers in the Fetch.scala file](https://github.com/bblfish/scala-js-dom/blob/ae4f2a7817bb7694a2edec0ef70b17450ea94498/src/main/scala/org/scalajs/dom/experimental/Fetch.scala#L287).

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
sjrd: No, they only contain *abstract* methods, which is exactly what we want in JS traits most of the time.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
sjrd: No, the `JSIterable[A]` trait still exist but it doesn't have any method. It's only method is `iterator()` and accessible through this implicit class.

bblfish: Mhh for the above code I get:
```
[error] /Volumes/Dev/Programming/ScalaJS/scala-js-dom/src/main/scala/org/scalajs/dom/experimental/JSIterable.scala:51: @JSBracketCall is not allowed in Scala.js-defined JS classes
[error]         def bracketCall[A](method: js.Any)(): A
```

sjrd: True. I had forgotten that restriction ^^
Make it `@js.native` instead, then.

src/main/scala/org/scalajs/dom/experimental/package.scala
bblfish: ok. In the html doc it's lower case, that's why I put it that way here...

bblfish:
I fixed a few of the points with `JSIterable` and `JSIterator`, but quite a few of the methods do need a `js.native`, in order to access the underlying JS functions.  I have tested it with my application https://github.com/read-write-web/rww-scala-js/commit/ebb8b7ea50b2257eb5ad5ac4fab1a20673f3c748

sjrd:
> but quite a few of the methods do need a js.native, in order to access the underlying JS functions.

What makes you think that?

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
bblfish: I think the `Entry` is a JS object and its two functions are `js.native`

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
```diff
@@ -0,0 +1,56 @@
...
+}
+
+object JSIterator {
+  private val iteratorSymbol = g.Symbol.iterator
+
+  @js.native
+  trait Entry[+A] extends js.Object {
+    def done: Boolean = js.native
+    def value: A = js.native
+  }
```
bblfish: `value` and `done` are to js.native functions, so `Entry` can't be `@ScalaJSDefined` right?

bblfish: perhaps not. I suppose it's such a basic class...

bblfish:
Is that better?

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
bblfish: I also get then 

```
[info] Compiling 1 Scala source to /Volumes/Dev/Programming/ScalaJS/scala-js-dom/target/scala-2.11/classes...
[error] /Volumes/Dev/Programming/ScalaJS/scala-js-dom/src/main/scala/org/scalajs/dom/experimental/JSIterable.scala:48: Traits and classes may not have inner native JS traits, classes or objects
[error]     trait IteratorMethodAccess extends js.Object {
```

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -0,0 +1,491 @@
...
+
+  def credentials: js.UndefOr[RequestCredentials]
+
+  def requestCache: js.UndefOr[RequestCache]
+
+  def requestRedirect: js.UndefOr[RequestRedirect]
+
+  def integrity: js.UndefOr[String]
+
+  def window: js.UndefOr[Null]
```
sjrd: `js.UndefOr[Null]` looks suspicious ^^

bblfish: yes, but it's actually what they seem to say in the docs. "can only be null" .
That's apparently to allow you to use it like ServiceWorkers from a window.

src/main/scala/org/scalajs/dom/experimental/JSIterable.scala
sjrd: Riight ... That's why I wanted it Scala.js-defined in the first place. :p

So you have to put it `private` in the `object JSIterator`.

src/main/scala/org/scalajs/dom/experimental/ServiceWorker.scala
```diff
@@ -0,0 +1,229 @@
...
+   */
+  val activated = "activated".asInstanceOf[ServiceWorkerState]
+
+  /**
+   * A new service worker is replacing the current service worker, or the
+   * current service worker is being discarded due to an install failure.
+   */
+  val redundant = "redundant".asInstanceOf[ServiceWorkerState]
+}
+
```
sjrd: Spurious blank line, here.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
sjrd: Lol. OK, then.

bblfish:
No more comments left in the code tab. Ready to go?

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,379 @@
...
+   * acquiring a reader. To cancel the stream, cancel both of the resulting
+   * branches; a composite cancellation reason will then be propagated to the
+   * stream’s underlying source.
+   *
+   * Note that the chunks seen in each branch will be the same object. If the
+   * chunks are not immutable, this could allow interference between the two
+   * branches. (Let us know if you think we should add an option to tee that
+   * creates structured clones of the chunks for each branch.)
+   */
+  def tee(): js.Array[ReadableStream[T]] = js.native
```
sjrd: Ah I see this was blocking the covariant thing.
Would `js.Array[_ <: ReadableStream[T]]` allow you to make `T` covariant?

Ideally this should even be a `js.Tuple2[ReadableStream[T], ReadableStream[T]]`, which would also solve the covariant issue.

src/main/scala/org/scalajs/dom/experimental/Stream.scala
```diff
@@ -0,0 +1,379 @@
...
+   *
+   * This method is intentionally generic; it does not require that its this
+   * value be a ReadableStream object. It also does not require that its
+   * writable argument be a WritableStream instance, or that its readable
+   * argument be a ReadableStream instance.
+   *
+   * //todo: determine the type of options
+   */
+  def pipeThrough[U](
+      pair: Tuple2[WriteableStream[T], ReadableStream[U]],
```
sjrd: A `Tuple2` is still a `scala.Tuple2`, which is not a correct JavaScript type. It should be a `js.Tuple2`.

bblfish: My IDE could not find js.Tuple2. I'll check again...

sjrd: Ah but of course `js.Tuple`s are not yet in 0.6.5 -_-'

I suggest you write `Any` for now, and keep this as comment:
```scala
pair: Any, // TODO js.Tuple2[WriteableStream[T], ReadableStream[U]]
```

bblfish:
done.

sjrd:
Good!

Last step: could you squash all the commits, please?

bblfish:
commits squashed. I learned a lot with the PR. :-)

mseddon:
:+1: whew! Thanks for that @bblfish, good work!

sjrd:
LGTM, thanks!

PR #182, opened 2015-12-04
Add window.devicePixelRatio

nornagon:
https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio


mseddon:
I ran into this omission a while back, but completely forgot about it :)

sjrd:
Is this supported? MDN says ??? for all browsers?

nornagon:
Works everywhere but IE <= 11:

http://caniuse.com/#feat=devicepixelratio

sjrd:
LGTM

mseddon:
:+1: we need to support those who care for now. Everyone else must treat this as 1:1 :)

PR #183, opened 2015-12-09
make NodeList items have Element properties

vn971:
make NodeList items have Element properties

Extending `Element` gives methods like `getElementById`, `innerHTML`.

`NodeList` itself can be obtained by using
`getElementsByClassName`, `childNodes` etc.

I could not find an explicit type definition for NodeList elements,
but it seems that all major browsers understand NodeList similarly.

NB: Merge with care! JS is not my expertise area.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -5693,7 +5693,7 @@ trait DOMList[T] extends js.Object{
  * MDN
  */
 @js.native
-class NodeList extends DOMList[Node]
+class NodeList extends DOMList[Element]
 
 
 @js.native


mseddon:
`Node` is a superclass of `Element`, and for good reason we use `NodeList` rather than `ElementList`.  For example- `querySelectorAll("*")` will return TextNodes (which are not `Element`) as well as elements such as `<div>`.

The IDL definition of NodeList can be found [here](http://www.w3.org/TR/1998/REC-DOM-Level-1-19981001/idl-definitions.html).

vn971:
@mseddon Are you sure `HTMLDivElement` is not an `Element`? It sounds like being one.
I created this code to test, it worked without any errors:

```
		val all = document.querySelectorAll("*")
		for {
			i <- 0 until all.length
			s = all(i)
		} {
			dom.window.console.log(s.asInstanceOf[js.Dynamic].innerHTML)
		}
```

mseddon:
`HTMLDivElement` is definitely an `Element`.  But `TextNode` is not- that is, `<div>Hello</div>` is a `HTMLDivElement` containing a `TextNode` child, having the string value "Hello".

mseddon:
The reason your test appeared to succeed, was because you are only looking at the top-level nodes in the NodeList (each may have children).  The TextNodes are buried within these at deeper nesting.

mseddon:
e.g.: https://jsfiddle.net/ejL10c85/

mseddon:
Note, I am not suggesting here that things like `getElementsByTagName` might not be better typed towards an `ElementList` type, but that care must be taken, since `NodeList` is the most specific type possible for things like `childNodes`.

vn971:
@mseddon I think I understood you now.
Thank you very much for the clarification.

vn971:
@mseddon well, unfortunately, the same link you posted suggests that `getElementsByTagName` _could_ contain non-Element-s.

I don't know what's better in this situation. This method works and returns `Element` in all major browsers, but specification says it may not.

mseddon:
That's the problem :/, there are places I feel the specification is just too loose, but it's tricky to be certain.  I definitely think there are places where things can be tightened (particularly since the recent introduction of `|` union types, there's a lot of places that still declare `Any`), and I am starting to believe that `HTMLDocument.createElement` could return a `HTMLElement` and `SVGDocument.createElement` could return `SVGElement` instances.  Any help rummaging through the specifications for things that can be tightened would be very much appreciated!

Issue #180, opened 2015-12-02
We're leaking experimental APIs


mseddon:
It appears the WebAudio api accidentally introduced a `MediaStream` trait (stubbed) into `scala.js.dom.raw` [here](https://github.com/scala-js/scala-js-dom/blob/e5b10a70771ae65bf666c55a6b21c626da2adbd5/src/main/scala/org/scalajs/dom/raw/lib.scala#L5508).  This type is also forwarded in `dom`.

As you can see [though](http://caniuse.com/#feat=stream), it's not compatible enough to meet the criteria of being outside of `experimental`.  But sadly, we've released it in `0.8.1` and `0.8.2`.

There may be others, and I will investigate.

@sjrd @lihaoyi I wonder since the next release will be `0.9.0` if we do some slightly breaking reorganization to fix up anything that slipped out? Hopefully it won't impact too many people.  

https://github.com/scala-js/scala-js-dom/pull/179 fleshes out the ~~`MediaStream`~~ class (because it's needed for WebRTC), but I've been afraid to move it into experimental without checking with you guys that this breaking change is acceptable.

sjrd:
Sure, we can break things.

You can also add a deprecated forwarder in the original location.

mseddon:
Ah, that's good- I wasn't sure if it was possible to annotate `types`.  That makes this kind of reorg is much less worrying.

mseddon:
`ServiceWorker`, `ServiceWorkerRegistration`, `ServiceWorkerContainer` and `NavigatorID`, all want to live in `experimental.servieworker`.

mseddon:
`Promise` is sort of annoying- http://caniuse.com/#feat=promises since they're not present in IE11.  They are in Edge, but that unfortunately doesn't mean anything to windows 7 and windows 8 users.

sjrd:
`Promise` needs to go to core anyway, because it's specified in ES6 itself.

mseddon:
`EventSource` is not supported in IE or Edge.

mseddon:
That's everything I found in `lib.scala`.  It appears `WebAudio.scala` in general should live in `experimental`, because it is not supported under IE.

mseddon:
IndexDB is an interesting one.  IE and Edge don't support some operations, and Safari's implementation is just broken.  Technically it's "present" though.

kapunga:
Is this still an issue?  If it is, I'd be happy to go clean up what is left.

[...]

PR #187, opened 2015-12-13
Add Web Workers API (SharedWorker, WorkerScopes, ServiceWorker) (fixes #186)

### LINKED ISSUE #186: Support for SharedWorker
There is already an interface for `Worker` and `AbstractWorker`. It would be nice if we could support the [`SharedWorker`](https://developer.mozilla.org/de/docs/Web/API/SharedWorker) too!

danielwegener:
This PR adds interfaces for the following Web Workers API interfaces (all in the experimental package for now - even they are not really tagged as experimental in MDN):

- NavigatorLanguage (mixed into Navigator)
- WorkerNavigator
- WorkerLocation
- WorkerGlobalScope
- DedicatedWorkerGlobalScope
- SharedWorkerGlobalScope
- ServiceWorkerGlobalScope

The following interfaces for ServiceWorkers are added
- ClientType
- ClientQueryOptions
- Clients

The idea is that a user can create a `serviceWorker.js` with scala.js ([with a wrapper that directly calls the main method](https://gitter.im/scala-js/scala-js?at=5625da510b33c42d5f81e968)) like this (note that the message event comes from ServiceWorker):

```scala
@JSExport
object SharedServiceWorker {
  val self = SharedWorkerGlobalScope.self
  @JSExport
  def main(): Unit = {
    self.onconnect = (m:MessageEvent) => {
      m.ports(0).postMessage(s"hello client! I am ${self.name}")
    }
  }
}
```


danielwegener:
Disclaimer: Untested yet, I am on it

mseddon:
:+1:

danielwegener:
Okay, gonna pick up https://github.com/scala-js/scala-js-dom/issues/180#issuecomment-161286422 and move them to the experimental package. Would you be fine with merging all worker related api into a `WebWorkers.scala` (like `ServiceWorker.scala` and my new `ChannelMessagingApi.scala` - which is misleading anyway)?

mseddon:
Yes, I think that'll work. From a user perspective that won't have much impact.

mseddon:
Thanks :)

danielwegener:
`NavigatorID` should stay where it is but `Navigator.serviceWorker` should go into the experimental package.
Thought about creating an
```
implicit class NavigatorOps(private val navigator:Navigator) extends AnyVal {
  val serviceWorker: ServiceWorkerContainer = js.native
}
```
in `experimental/package.scala`. Thoughts on that?

I'll also keep WebWorkers and ServiceWorker separated (different W3C standards) since WebWorkers should already be pretty stable but ServiceWorker probably still will get some changes.

mseddon:
Actually, I'd put it in a subpackage specifically for serviceworkers- that way people can import just the features that they want without accidentally pulling in *all* experimental apis (and getting tripped up believing they're accessing a standard method).

Rather than deleting the currently exposed types from the dom package object, for now we should probably leave them there, pointing to their new location, but add a `@deprecated` attribute mentioning they should now be imported from the new location.  Unsure how many people are actually using it from the old (incorrect) location, but it's nicer to just follow the deprecation warnings (or choose to ignore them) if you have to upgrade your dependencies on a tight deadline. :)

danielwegener:
Yeah thats what i thought too - so we'd also need a org.scalajs.dom.raw package-object that has `@deprecated` type-aliases pointing to the new location.

mseddon:
Yep, This is what happens when things go in the wrong place. :/

danielwegener:
_One_ could argue that the types in the raw-package are non-public anyway and one should only rely on types on `org.scalajs.dom` ;)

mseddon:
You could.  Unfortunately Intellij sucks import classes from their package of definition, so people are probably already doomed :)

mseddon:
Oh, definitely think you should use something like `dom.experimental.worker` (perhaps you can come up with a better name) to keep those experimental features separated.

danielwegener:
Thats what I ment - if they only existed in dom.raw - they never have been public and we could move them freely - but i know what you mean ;) I'd prefer the original spec name (webworkers, serviceworkers -maybe to _longish_?) - think it is a good idea anyway to roughly split our packages by w3c api - or _feature_ - just tried to stick with the existing convention.

mseddon:
I definitely agree with classification by specification.  As for 'convention', apart from a few apis, it has generally been "chuck it in `dom.raw`", something I'm trying to incrementally move us away from :)

I notice actually that [Channel Messaging](http://caniuse.com/#feat=channel-messaging) is flagged here as 'very compatible'.  But [Service Workers](http://caniuse.com/#search=service%20workers) are (extremely) experimental.  For now we can probably keep them in `experimental` though.

danielwegener:
(renamed PR Channel Messaging API to [Web Workers API](http://caniuse.com/#search=Web%20Workers) - but that one looks stable too)

danielwegener:
Okay - I came up with the following solution:

- WebWorkers (except their SharedWorker part) are stable and [available](http://caniuse.com/#search=web%20workers). Hence I'll put them into `org.scalajs.dom.webworkers` package. We need no raw package, because WebWorker classes are not prefixed.
- SharedWorkers are experimental and moved into `org.scalajs.dom.experimental.sharedworkers` (I am not quite sure if the contained classes should all be within one file)
- ServiceWorkers are experimental and moved into `org.scalajs.dom.experimental.serviceworkers`
  - ServiceWorker specifics are moved from `org.scalajs.dom` and `org.scalajs.dom.raw` into this package
- All moved classes have type aliases with deprecation-marker at the place where they were removed

Caveats:
- now we have two `Transferrables`, one in `dom.webworkers` and one in `dom.experimental.serviceworkers`

Sorry for this back and forth noise (and those tons of builds)

mseddon:
No worries- it's better to think these things through allowed. :)

I need to famialrize myself more with the shared worker and channel specifications, which I'll aim to do tomorrow.  Am I understanding correctly that it is the *same* `Transferable` class, implemented by both specifications, or are there differences?

Also, am I right thinking you're effectively working from the [channel messaging](https://html.spec.whatwg.org/multipage/comms.html#channel-messaging) and [web worker](https://html.spec.whatwg.org/multipage/workers.html) sections in the whatwg living standard?

danielwegener:
For webWorkers I tried to stick with the latest [W3C Web Workers Working Draft](http://www.w3.org/TR/2015/WD-workers-20150924/)  although the descriptions are from MDN because they are more user centric.

For ServiceWorkers I took the [WHATWG Service Workers 1 Editors Draft](https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/) because it was already provided as reference in the existing ServiceWorkers code. I took the MDN description for ServiceWorkers where it was available.

This PR is not really about channel messaging but just different flavours of Web Workers (I confused it when I opened the PR in the first place) - The [Channel Messaging API](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API) is already present in the scalajs.dom lib.

mseddon:
Alright, thanks for that clarification (certainly no issue with using those sources, just makes things easier if we're all looking at the same spec :))

kapunga:
Any luck with the implicit trait extending Navigator?  I've tried a few similar things and they all seem to result in a compiler error when attempting to use them in a project: `Compilation error[value serviceWorker is not a member of org.scalajs.dom.raw.Navigator`

mseddon:
@kapunga see https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/experimental/Vibration.scala for a simple example of how to do this.

kapunga:
@mseddon Thanks.  I forgot that implicits need to be in scope with a single identifier and it's not enough to just `import org.scalajs.dom.experimental.serviceworkers` for example.  :-)

mseddon:
;)

danielwegener:
@mseddon I use the current version of this branch in a (unfortunately closed) that covers most of the serviceworker specific api. It is a bit hard to write an integration-test or example since it requires two separate javascript files (hence two sbt modules) that depend on each other at runtime. Do you have any ideas on that?

lihaoyi:
I'm totally unfamiliar with these APIs, but the convention I started this `raw`/`dom` stuff with was that *everything* is inside the `raw` namespace, and is only aliased inside `dom`/`dom.html`/etc. as necessary. Furthermore, the stuff in the sub-packages of `dom` are all done based on name-prefixes. Stuff without a prefix is aliased in `dom` itself.

I don't have enough context on this particular diff, and I'm not actively involved in scala-js-dom enough to block this review, so do whatever you want with this feedback. If you all feel it's worth breaking the convention in this case (or in all cases) that's fine, but make sure you update the *english* description of this package at http://scala-js.github.io/scala-js-dom/#Usage to contain whatever new convention you want to follow

danielwegener:
Hi @lihaoyi, cheers for the feedback! I thought the earlier APIs had this raw/prefix/typealiasing because the specs itself actually used the prefixes in their IDL (html, css, svg, webgl) - and having unprefixed but namespaced aliases would make them much easier to use (less typing).

However, other specs do not really have these prefixes (Audio, Fetch, Stream, WebWorkers, ServiceWorkers, Channel Messaging) and I can not really see what benefit we would get from having a raw type and a public type alias (it is rather confusing when the IDE tries to resolve auto imports :)).

I am fine do rework this PR into the raw+.. layout and I have absolutely no preferences how the source tree is structured. I just think it would be a good idea to have all js types in a namespace that matches  the spec where they come from.

sjrd:
I think the organization in the present of this PR is fine. It makes sense to reorganize stuff into appropriate subpackages IMO.

mseddon:
Yes, one of the fun things with Intellij at least, is it autoimports from the package of definition.  That said, a casual user of `scala-js-dom` may only know the JS name of the definition, and defining the class with that name (possibly aliasing it elsewhere for convenience).  In my current commits I've always used the fully specified name (without shortening) for both this purpose and so it's easy for others to compare my code against the IDL.  That said I do find myself prefer `html.Element` rather than `dom.raw.HTMLElement` in day to day usage.

This throws up an interesting problem though- if we were to take this to the extreme in the library (assume `dom.html._` contains the actual definitions), it would be important to note in the scaladoc that `Element` actually refers to `HTMLElement`, rather than `SVGElement`.  It would also beg the question where does `Element` (also a type!) live?

My main issue isn't really with the presence of the `raw` package, which @lihaoyi had (and I still argue, has) a rational basis for, but with `lib.scala` which has slowly deteriorated into an enormous miscellaneous file, which probably needs to be broken up into several others.

mseddon:
Yes, one of the fun things with Intellij at least, is it autoimports from the package of definition.  That said, a casual user of `scala-js-dom` may only know the JS name of the definition, and defining the class with that name (possibly aliasing it elsewhere for convenience) seems to make sense.

In my current commits I've always used the fully specified name (without shortening) for both this purpose and so it's easy for others to compare my code against the IDL.  That said I do find myself prefer `html.Element` rather than `dom.raw.HTMLElement` in day to day usage.

This throws up an interesting problem though- if we were to take this to the extreme in the library (assume `dom.html._` contains the actual definitions), it would be important to note in the scaladoc that `Element` actually refers to `HTMLElement`, rather than `SVGElement`.  It would also beg the question where does `Element` (also a type!) live?

After reading the comments above, I realise my main issue isn't really with the presence of the `raw` package, which @lihaoyi had (and I would now still argue, has) a rational basis for, but with `lib.scala` which has slowly deteriorated into an enormous miscellaneous file, which probably needs to be broken up into several others.

I'm starting to think the (non experimental) webworkers classes *should* be declared in `raw` and exposed through `dom.webworkers`. See the file layout: https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/webgl.scala and https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/raw/WebGL.scala, which is fairly tidy, and provides both a raw and package organized api, while not all being stuffed in `lib.scala`.

mseddon:
@danielwegener wrt testing- it's particularly tricky to do in most cases and as you've probably noticed, there are no test suites for `scala-js-dom` at present.  Currently we release the library and wait for feedback from users and PR in corrections as they are noticed.  This isn't ideal, but I think everyone's drawn a blank on how to do this so far.

danielwegener:
I don't want this packaging discussion to stop this PR so I moved the (stable) webworkers api back to the raw package (in its own `WebWorkers.scala`) and added type aliases in `org.scalajs.dom.webworkers`. Left the experimental packages for sharedworkers and serviceworkers as they are.

However, given the raw/alias layout - where should I put types that are aliases themself then? (like `type Transferable = ArrayBuffer | MessagePort`, which is specific to webworkers and is even "widened" in the serviceworker package)

mseddon:
That's a good point.  Actually `Transferable` is part of the [HTML5 Specification](https://www.w3.org/TR/2015/WD-html51-20151008/infrastructure.html#transferable-objects), although it is widened by ServiceWorker, so in this situation I'd guess putting the more general HTML5 implementation in `dom.raw`, and do your widening of `Transferable` inside the experimental service worker.  Does this feel right to you?

danielwegener:
Good point - so we'd need a package object for `dom.raw` (need it anyway for backward compatible deprecated type aliases for the old exposed ServiceWorker types).  I think I have all parts together (sqashed, rebased).

mseddon:
Ick, yes we would.  Of course the irritating thing about package objects is we can't then break them over several files.

mseddon:
Yes, that looks about right, I'll do a proper review of that tomorrow.  Thanks for hanging in there :)

danielwegener:
Cheers @mseddon, thanks for your time! Very good catches!

mseddon:
https://travis-ci.org/scala-js/scala-js-dom/jobs/102190678 - Whoops my "This guy seems obsolete" statement was miles off! :D sorry about that

mseddon:
Other than that, only minor formatting issues remain from what I can see. Thanks!

mseddon:
That looks better, thanks.  (Sorry, didn't notice you'd made a change, if you post a comment when you're pushing it will notify people).  I think other than squashing into single commit, that's all.  Once you've done that if you post a comment, @sjrd will review and merge.  This was a bit of a monster cleanup, thanks for hanging in there :)

danielwegener:
@sjrd @mseddon There you go! I'd rather thank you for your patience :)

src/main/scala/org/scalajs/dom/experimental/package.scala
```diff
@@ -47,9 +47,6 @@ package object experimental {
    */
   type OpenEndedDictionary[T] = js.Dictionary[T]
 
-  /**
-   * [[https://html.spec.whatwg.org/multipage/infrastructure.html#transferable ¶2.7.4 Transferable Objects]]
-   * in whatwg html spec.
-   */
-  type Transferable = js.typedarray.ArrayBuffer | CanvasProxy | MessagePort
+
+
```
sjrd: Triple blank line (there should only be one)

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -0,0 +1,683 @@
...
+sealed trait FrameType extends js.Any
+
+/**
+ * part of ServiceWorker
+ * [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#client-frametype ¶4.2.2 frameType]]
+ * of serviceWorker spec
+ */
+object FrameType {
+  /**
+   * The window client's global object's browsing context is an auxiliary browsing context.
```
sjrd: Try and keep line length under 80 characters (obviously not possible with long URLs)

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -0,0 +1,683 @@
...
+/**
+ * See [[https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent FetchEvent]] on MDN
+ *
+ * [[https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/#fetch-event-section ¶4.5 FetchEvent]]
+ * on whatwg ServiceWorker spec.
+ */
+@js.native
+class FetchEvent extends Event {
+  /**
+   * @return Boolean that is true if the event was dispatched with the user's
```
sjrd: Drop `@return`. (the same thing happens elsewhere)

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -0,0 +1,683 @@
...
+   * the request is a top-level navigation and the return value is a Response
+   * whose type attribute is opaque (i.e. an opaque response body), a network
+   * error is returned to Fetch. The final URL of all successful (non
+   * network-error) responses is the requested URL.
+   */
+  def respondWith(promisedResponse: Response | Promise[Response]): Unit = js.native
+}
+
+/**
+  * The ServiceWorker interface of the ServiceWorker API provides a reference to a service worker.
```
sjrd: Avoid that Scaladoc style. The `*`'s should be aligned with the *first* of the two stars of `/**`, not the second one.

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -0,0 +1,683 @@
...
+
+  /**
+   * Returns the event's data. It can be any data type.
+   */
+  val data: Any = js.native
+
+  /**
+   * Returns the origin of the service worker's environment settings object.
+   */
+  val origin: String = js.native //should be type DOMString
```
sjrd: `DOMString` *is* `String`. This comment is irrelevant. Look for "should be", there are others like that.

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -0,0 +1,683 @@
...
+
+/**
+  * The ServiceWorkerMessageEvent interface of the ServiceWorker API contains information about an event sent to a ServiceWorkerContainer target.
+  * This extends the default message event to allow setting a ServiceWorker object as the source of a message.
+  * The event object is accessed via the handler function of a message event, when fired by a message received from a service worker.
+  *
+  * MDN
+  */
+@js.native
+class ServiceWorkerMessageEvent(`type`: String, eventInitDict: ServiceWorkerMessageEventInit = js.native) extends Event {
```
sjrd: Wrap before `eventInitDict` (with 4 spaces)

src/main/scala/org/scalajs/dom/experimental/sharedworkers/SharedWorkerGlobalScope.scala
```diff
@@ -0,0 +1,56 @@
...
+
+}
+
+@js.native
+object SharedWorkerGlobalScope extends js.GlobalScope {
+  def self: SharedWorkerGlobalScope = js.native
+}
+
+
+
```
sjrd: Remove those 3 blank lines at EOF

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -206,8 +210,10 @@ package object dom {
   type WindowTimersExtension = raw.WindowTimersExtension
 
   type WindowBase64 = raw.WindowBase64
+  @deprecated(message = "Import from org.scalajs.dom.webworkers instead.", since = "0.9.0")
   type Worker = raw.Worker
-  lazy val Worker: raw.Worker.type = raw.Worker
+  @deprecated(message = "No replacement.", since = "0.9.0")
+  lazy val Worker: Any = webworkers.Worker
```
sjrd: Isn't `webworkers.Worker` precisely the replacement?

src/main/scala/org/scalajs/dom/raw/WebWorkers.scala
```diff
@@ -0,0 +1,332 @@
...
+    * The AbstractWorker.onerror property represents an EventHandler, that is a
+    * function to be called when the error event occurs and bubbles through the Worker.
+    *
+    * MDN
+    */
+  var onerror: js.Function1[ErrorEvent, _] = js.native
+
+}
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/WebWorkers.scala
```diff
@@ -0,0 +1,332 @@
...
+  * Of note is the fact that workers may in turn spawn new workers as long as
+  * those workers are hosted within the same origin as the parent page.  In
+  * addition, workers may use XMLHttpRequest for network I/O, with the exception
+  * that the responseXML and channel attributes on XMLHttpRequest always return
+  * null.
+  *
+  * MDN
+  */
+@js.native
+class Worker (stringUrl: String) extends AbstractWorker {
```
sjrd: No space between class name and `(`

src/main/scala/org/scalajs/dom/raw/WebWorkers.scala
```diff
@@ -0,0 +1,332 @@
...
+  /**
+    * The Worker.terminate() method immediately terminates the Worker. This does not
+    * offer the worker an opportunity to finish its operations; it is simply stopped at
+    * once.
+    *
+    * MDN
+    */
+  def terminate(): Unit = js.native
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/WebWorkers.scala
```diff
@@ -0,0 +1,332 @@
...
+}
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/WebWorkers.scala
```diff
@@ -0,0 +1,332 @@
...
+  var onoffline: js.Function1[Event, _] = js.native
+  /**
+    * The ononline property of the WorkerGlobalScope interface represents an EventHandler
+    * to be called when the online event occurs and bubbles through the Worker.
+    *
+    * MDN
+    */
+  var ononline: js.Function1[Event, _] = js.native
+
+
```
sjrd: Double blank line.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -693,11 +545,8 @@ trait ServiceWorkerContainer extends EventTarget {
  * MDN
  */
 @js.native
-class Navigator extends NavigatorID with NavigatorOnLine with NavigatorContentUtils with NavigatorGeolocation with NavigatorStorageUtils {
-
-  val serviceWorker: ServiceWorkerContainer = js.native
-
-}
+class Navigator extends NavigatorID with NavigatorOnLine with NavigatorContentUtils
+with NavigatorGeolocation with NavigatorStorageUtils with NavigatorLanguage
```
sjrd: Indent 4 spaces

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -4795,6 +4644,26 @@ trait NavigatorOnLine extends js.Object {
...
+@js.native
+trait NavigatorLanguage extends js.Object {
+
+  /**
+    * Returns a DOMString representing the preferred language of the user, usually
+    * the language of the browser UI. The null value is returned when this is unknown.
+    *
+    * MDN
+    */
+  def language:String = js.native
```
sjrd: Space after `:`

src/main/scala/org/scalajs/dom/webworkers.scala
```diff
@@ -0,0 +1,16 @@
+package org.scalajs.dom
+
+
+object webworkers {
+
+  type AbstractWorker = raw.AbstractWorker
+  type DedicatedWorkerGlobalScope = raw.DedicatedWorkerGlobalScope
+  lazy val DedicatedWorkerGlobalScope : raw.DedicatedWorkerGlobalScope.type = raw.DedicatedWorkerGlobalScope
```
sjrd: No space before `:`

src/main/scala/org/scalajs/dom/webworkers.scala
```diff
@@ -0,0 +1,16 @@
...
+  type Worker = raw.Worker
+  @deprecated(message = "No replacement", since = "0.9.0")
+  lazy val Worker : Any = ???
```
sjrd: That's not useful. Remove it completely.

sjrd:
That's all.

src/main/scala/org/scalajs/dom/package.scala
danielwegener: webworkers.Worker itself is `???`. It is a stale object that has no purpose and is not even mentioned in the spec

danielwegener:
@sjrd cheers. all done. I am off writing a scalariform config. or trying to push this into the idea formatter. that was a bit cruel :) (no doubt, reviewing for code style must be no fun either)

mseddon:
stoopid travis lost network. :/

mseddon:
LGTM

danielwegener:
@mseddon squashed!

mseddon:
Thanks! @sjrd okay to merge?

sjrd:
LGTM

PR #189, opened 2015-12-15
Fix #184: Add support for the Push API

### LINKED ISSUE #184: Add support for the Push API
Add support for the Push API specified here:

https://developer.mozilla.org/en-US/docs/Web/API/Push_API

kapunga:
I did not include any of the following:
* Any item marked deprecated in MDN.

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
@@ -1,0 +1,225 @@
+package org.scalajs.dom.experimental.push
+
+import org.scalajs.dom.experimental.serviceworkers.ExtendableEvent
+import org.scalajs.dom.raw.{Blob, Promise}
+
+import scala.scalajs.js
+import scala.scalajs.js.typedarray.ArrayBuffer
+
+/**
+ * The PushManager interface of the Push API provides a way to receive notifications
+ * from third-party servers as well as request URLs for push notifications.
+ *
+ * This interface is accessed via the ServiceWorkerRegistration.pushManager property.
+ *
+ * MDN
+ *
+ * The Push API is currently specified here: [[http://www.w3.org/TR/2015/WD-push-api-20151215/]]
+ */
+@js.native
+trait PushManager extends js.Object {
+  /**
+   * Retrieves an existing push subscription. It returns a Promise that resolves to a
+   * PushSubscription object containing details of an existing subscription. If no
+   * existing subscription exists, this resolves to a null value.
+   *
+   * MDN
+   */
+  def getSubscription(): Promise[PushSubscription] = js.native
+
+  /**
+   * The permissionState() method of the PushManager interface returns a Promise that
+   * resolves to a DOMString indicating the permission state of the push manager.
+   * Possible values are  'prompt', 'denied', or 'granted'.
+   *
+   * @param options An object containing optional configuration parameters. It can have
+   * the following properties:
+   * - userVisibleOnly: A boolean indicating that the returned push subscription will only
+   *                    be used for messages whose effect is made visible to the user.
+   *
+   * MDN
+   */
+  def permissionState(options: PushSubscriptionOptions = js.native): Promise[PushPermissionState] = js.native
+
+  /**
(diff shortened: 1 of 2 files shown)


kapunga:
Question wrt Push API methods in ServiceWorker.  Once ServiceWorkers are moved into experimental, should developers still be required to import push package implicits to access fields like `ServiceWorkerRegistration.pushManager`, or is it assumed once you are using experimental packages, you are using all of them?

mseddon:
In general, users should enable experimental features per api, rather than as a whole.  For example, [ServiceWorkers](http://caniuse.com/#search=service%20workers) are supported by Opera, but the [Push API](http://caniuse.com/#search=push%20api) is not.

kapunga:
Thanks a lot for the feedback!  I'll get to work tweaking things.

kapunga:
Question:
What do we do when there is something defined to either accept or return an enum like this?
```
enum PushEncryptionKeyName {
    "p256dh",
    "auth"
};
```

kapunga:
Question:
What do we do when there is something defined to either accept or return an enum like this?
```
enum PushEncryptionKeyName {
    "p256dh",
    "auth"
};
```
It seems you could offer additional safety by using an enum, but I couldn't find an example in the code and there doesn't seem to be any reference to such things here:
[[http://www.scala-js.org/doc/interoperability/facade-types.html]]

sjrd:
https://github.com/scala-js/scala-js-dom/blob/87ab4d70649078f5cf9a4c0e0e22c189de86a198/src/main/scala/org/scalajs/dom/crypto/Crypto.scala#L872-L887

kapunga:
Thanks.

src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
```diff
@@ -0,0 +1,225 @@
...
+   * Possible values are  'prompt', 'denied', or 'granted'.
+   *
+   * @param options An object containing optional configuration parameters. It can have
+   * the following properties:
+   * - userVisibleOnly: A boolean indicating that the returned push subscription will only
+   *                    be used for messages whose effect is made visible to the user.
+   *
+   * MDN
+   */
+  def permissionState(options: PushSubscriptionOptions): Promise[PushPermissionState] = js.native
```
mseddon: Optional parameter, so define `def permissionState(options: PushSubscriptionOptions = js.native): Promise[PushPermissionState] = js.native`

src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * The subscribe() method of the PushManager interface subscribes to a push service.
+   *
+   * It returns a Promise that resolves to a PushSubscription object containing details of a
+   * push subscription. A new push subscription is created if the current service worker
+   * does not have an existing subscription.
+   *
+   * MDN
+   */
+  def subscribe(options: PushSubscriptionOptions): Promise[PushSubscription] = js.native
```
mseddon: Same here.

mseddon:
That's all.

kapunga:
@mseddon I was waiting for #187 to close, since I was waiting on an interface there.  Pushing it now.

mseddon:
LGTM, thanks!

mseddon:
LGTM, thanks! @sjrd will take it from here.

src/main/scala/org/scalajs/dom/experimental/push/package.scala
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * The pushManager property of the ServiceWorkerRegistration interface returns
+     * a reference to the PushManager interface for managing push subscriptions;
+     * this includes support for subscribing, getting an active subscription, and
+     * accessing push permission status.
+     *
+     * MDN
+     */
+    var pushManager: PushManager = js.native
```
sjrd: This should probably be a `val`

sjrd:
That's all.

kapunga:
@mseddon @sjrd Thanks!  Just double checked it, it should be done!

sjrd:
LGTM

PR #190, opened 2015-12-21
Correct algorithm parameter types in SubtleCrypto interface

povder:
This PR is an outcome of this discussion: https://github.com/bblfish/scala-js-dom/commit/d0800da8fb13d14817d556dbce000df9f24e5211#commitcomment-15037842

### DIFF
--- a/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
+++ b/src/main/scala/org/scalajs/dom/crypto/Crypto.scala
@@ -165,11 +165,8 @@ trait SubtleCrypto extends js.Object {
    *
    * Defined at
    * [[http://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-encrypt ¶14.3.1 The encrypt method]]
-   *
-   * We are being a bit more precise than the official definition by requiring a KeyAlgorithmIdentifier
-   * rather than an AlgorithmIdentifier for the algorithm field.
    */
-  def encrypt(algorithm: KeyAlgorithmIdentifier, key: CryptoKey,
+  def encrypt(algorithm: AlgorithmIdentifier, key: CryptoKey,
       data: BufferSource): Promise[js.Any] = js.native
 
   /**
@@ -178,11 +175,8 @@ trait SubtleCrypto extends js.Object {
    *
    * Defined at
    * [[http://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-decrypt ¶14.3.2 The decrypt method]]
-   *
-   * We are being a bit more precise than the official definition by requiring a KeyAlgorithmIdentifier
-   * rather than an AlgorithmIdentifier for the algorithm field.
    */
-  def decrypt(algorithm: KeyAlgorithmIdentifier, key: CryptoKey,
+  def decrypt(algorithm: AlgorithmIdentifier, key: CryptoKey,
       data: BufferSource): Promise[js.Any] = js.native
 
   /**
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/crypto/Crypto.scala
```diff
@@ -242,10 +235,10 @@ trait SubtleCrypto extends js.Object {
    * Defined at
    * [[http://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-deriveKey ¶14.3.7 The deriveKey method]]
    *
-   * We are being a bit more precise than the official definition by requiring a KeyAlgorithmIdentifier
-   * rather than an AlgorithmIdentifier for the algorithm and derivedKeyType field
+   * We are being a bit more precise than the official definition by requiring a HashAlgorithmIdentifier
+   * rather than an AlgorithmIdentifier for the algorithm and KeyAlgorithmIdentifier for derivedKeyType
    */
-  def deriveKey(algorithm: KeyAlgorithmIdentifier, baseKey: CryptoKey,
+  def deriveKey(algorithm: HashAlgorithmIdentifier, baseKey: CryptoKey,
```
mseddon: Hmm. Other algorithms (e.g. ECDH and DH) also support `deriveKey` and `deriveBits` [http://www.w3.org/TR/WebCryptoAPI/#algorithm-overview].  These are asymmetric ciphers, not hash algorithms.  This makes me think `AlgorithmIdentifier` is the only correct typing here.

povder: Thanks for checking. I'll relax the type here. I'll relax the type in ```sign``` method as well.

I think I'll create another pull request in the future introducing MacAlgorithm type for verify and sign. I'll do it in another PR since I think it will need some more discussion.

povder:
@sjrd Can you please take a look at this PR?

mseddon:
Sorry about the delay- things got a little idle over christmas, I'll review this tomorrow properly.  Thanks!

example/src/main/scala/example/Example.scala
```diff
@@ -173,3 +173,173 @@ object AjaxExtension {
     }
```
mseddon: @sjrd do we want this example + deps in the example project?

mseddon:
That's all really, the fixes seem good- overall relaxing those signatures fixes the underlying problem, and the strengthening of digest to require `HashAlgorithmIdentifier` is correctly supporting exactly the SHA family.

build.sbt
```diff
@@ -72,4 +72,9 @@ lazy val readme = ScalatexReadme(
 lazy val example = project
   .dependsOn(root)
   .enablePlugins(ScalaJSPlugin)
-  .settings(scalaVersion := "2.11.6")
+  .settings(
+    scalaVersion := "2.11.6",
+    libraryDependencies += "com.github.marklister" %%% "base64" % "0.2.1",
+    resolvers += Resolver.bintrayRepo("marklister", "maven")
```
sjrd: This dependency is a deal breaker. We cannot afford to have scala-js-dom depend on anything but Scala.js core; otherwise we'll be unable to publish scala-js-dom for new versions of Scala.js until `base64` is published, and that will be a nightmare.

mseddon:
Alright, suggest we drop commit 6c4db04 and squash the rest, pending any other style issues.

sjrd:
I agree. (I don't have other style issues to report)

mseddon:
@povder Thanks! Post a comment when this is done and we can merge, thanks for the fix!

mseddon:
Oh also, if you add "Fix: #196" to the beginning of the commit message, that will also close that issue once merged.

povder:
Commit 6c4db0493ea9036600cb37a97dac2627d2210f9d removed, other commits squashed. Thanks for the review.

mseddon:
Thanks!

mseddon:
@sjrd okay for a merge?

sjrd:
I suggest removing the changes to `Example.scala`. This file is only meant to contain the examples in the README. And this example is too complex for the readme.

Maybe we need a different discussion about how we want to go about more complex examples. But in the current state, it should not be included.

povder:
Changes to ```Example.scala``` removed

sjrd:
LGTM

PR #193, opened 2015-12-28
Html.disabled can be undefined

japgolly:
Calling `.disabled` on elements like `HTMLImageElement`, `HTMLDivElement`, `HTMLParagraphElement` type-checks and causes this nastiness (at least in Chrome):

```
Uncaught scala.scalajs.runtime.UndefinedBehaviorError:
An undefined behavior was detected: undefined is not an instance of java.lang.Boolean
```


sjrd:
LGTM

PR #195, opened 2015-12-31
Make File an abstract class

danielwegener:
... such that `foo.isInstanceOf[org.scalajs.dom.raw.File]` and `match ... case file: File ⇒` are supported.


sjrd:
Thanks. Could you please remove the `= ` prefix in the commit message?

danielwegener:
Wrong project, wrong commit convention ;)

sjrd:
LGTM

PR #199, opened 2016-01-18
Fix #198: Make WebGL constants final vals to support Safari.

### LINKED ISSUE #198: Safari WebGL is broken
Unfortunately, while Firefox, IE, Edge, Chrome and Opera allow access to the `GLenum` constants on the `WebGLRenderingContext` constructor directly, this does not work on Safari (iOS or Desktop).

This leaves two options:

* Move the constants defined in `object WebGLRenderingContext` onto `class WebGLRenderingContext`, which works, but personally I find pulling constants out of instance objects weird (especially since you might want to initialize a data structure with a GL constant *before* context initialization), or

* Make `object WebGLRenderingContext` a non-native object containing the literals as `final val` (which are the same across all browsers, and the C OpenGL apis where applicable) - This is confirmed to work in principle, since I have a library that does this which I've been using well for several months, and I only now just noticed this issue when migrating to pure `scala-js-dom`.  I prefer this solution, which appears to work, although it's tedious and error prone so some care must be taken (though I have what I believe to be a fairly correct implementation of such an object already, modulo some WebGL specific constants).

mseddon:
This has been tested against my current code and appears to now behave correctly under Safari as well as Chrome.  Also I found two enums that I'd apparently missed when initially creating the facade, so they've been added, too.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/WebGL.scala
+++ b/src/main/scala/org/scalajs/dom/raw/WebGL.scala
@@ -130,25 +130,31 @@ class WebGLShaderPrecisionFormat private () extends js.Object {
  *
  * MDN
  */
-@js.native
-object WebGLRenderingContext extends js.Object {
+object WebGLRenderingContext {
+  /* Note:
+   *   Normally we'd make this object @js.native extends js.Object
+   *   and declare all these vals as js.native, however Safari does
+   *   not store these constants on WebGLRenderingContext, thus we
+   *   have to explicitly define them as final vals instead.
+   */
+
   /**
    * Specifies the depth buffer should be cleared.
    * @see [[WebGLRenderingContext#clear]]
    */
-  val DEPTH_BUFFER_BIT: Int = js.native
+  final val DEPTH_BUFFER_BIT = 0x00000100
 
   /**
    * Specifies the stencil buffer should be cleared.
    * @see [[WebGLRenderingContext#clear]]
    */
-  val STENCIL_BUFFER_BIT: Int = js.native
+  final val STENCIL_BUFFER_BIT = 0x00000400
 
   /**
    * Specifies the color buffer should be cleared.
    * @see [[WebGLRenderingContext#clear]]
    */
-  val COLOR_BUFFER_BIT: Int = js.native
+  final val COLOR_BUFFER_BIT = 0x00004000
 
   /* BeginMode */
   /**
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/WebGL.scala
```diff
@@ -553,270 +558,274 @@ object WebGLRenderingContext extends js.Object {
...
+  final val VERTEX_ATTRIB_ARRAY_SIZE: Int = 0x8623
+  final val VERTEX_ATTRIB_ARRAY_STRIDE: Int = 0x8624
+  final val VERTEX_ATTRIB_ARRAY_TYPE: Int = 0x8625
+  final val VERTEX_ATTRIB_ARRAY_NORMALIZED: Int = 0x886A
+  final val VERTEX_ATTRIB_ARRAY_POINTER: Int = 0x8645
+  final val VERTEX_ATTRIB_ARRAY_BUFFER_BINDING: Int = 0x889F
+
+  /* Read Format */
+  final val IMPLEMENTATION_COLOR_READ_TYPE: Int = 0x8B9A
+  final val IMPLEMENTATION_COLOR_READ_FORMAT: Int = 0x8B9B
```
mseddon: These two guys were missing.

src/main/scala/org/scalajs/dom/raw/WebGL.scala
```diff
@@ -130,25 +130,30 @@ class WebGLShaderPrecisionFormat private () extends js.Object {
...
+   *   Safari does not store these constants on WebGLRenderingContext, thus we have to explicitly define them as final
+   *   vals instead.
+   */
+
   /**
    * Specifies the depth buffer should be cleared.
    * @see [[WebGLRenderingContext#clear]]
    */
-  val DEPTH_BUFFER_BIT: Int = js.native
+  final val DEPTH_BUFFER_BIT: Int = 0x00000100
```
sjrd: Remove the `: Int`'s so all these things can be constant-folded.

src/main/scala/org/scalajs/dom/raw/WebGL.scala
```diff
@@ -130,25 +130,30 @@ class WebGLShaderPrecisionFormat private () extends js.Object {
...
  * MDN
  */
-@js.native
-object WebGLRenderingContext extends js.Object {
+object WebGLRenderingContext {
+  /* Note:
+   *   Normally we'd make this object @js.native extends js.Object and declare all these vals as js.native, however
+   *   Safari does not store these constants on WebGLRenderingContext, thus we have to explicitly define them as final
+   *   vals instead.
+   */
```
sjrd: Keep the lines under 80 characters.

sjrd:
That's all.

mseddon:
Changed as requested.

sjrd:
LGTM

PR #200, opened 2016-01-18
Add console back to the dom package.

mseddon:
These disappeared from the dom package when we stopped using a js.native package object, so add them back to prevent needlessly breaking code in 0.9.0.

### DIFF
--- a/src/main/scala/org/scalajs/dom/package.scala
+++ b/src/main/scala/org/scalajs/dom/package.scala
@@ -219,4 +219,6 @@ package object dom {
 
   lazy val window: Window = js.Dynamic.global.asInstanceOf[Window]
   lazy val document: html.Document = window.document
+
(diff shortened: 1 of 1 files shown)


mseddon:
Whoops, I've tangled in the webgl commits there, let me fix that.

mseddon:
Corrected now.

sjrd:
`console` makes sense. But wouldn't it make more sense to always access `screen` through `window`?

mseddon:
My decision to add that was simply that I noticed some existing code break moving to 0.9.0.  Since it's a major release anyway, if you think it's worth omitting, I'm happy to do so.

mseddon:
There we go- console only for now.  We can always add `screen` if people complain.

sjrd:
LGTM

PR #202, opened 2016-01-19
window.location.origin should be UndefOr due to IE

### LINKED ISSUE #201: window.location.origin does not exist on IE
Unfortunately, window.location.origin does not exist on "some" versions of IE. Apparently IE 11 on windows 7 and the latest windows 10 have it, but everything I've tried doesn't have it. See https://connect.microsoft.com/IE/feedback/details/1763802/location-origin-is-undefined-in-ie-11-on-windows-10-but-works-on-windows-7

Probably the solution is to include a polyfill. Modernizr has one, is it worth considering just including modernizr?

Otherwise, this is the polyfill needed.
```
if (!window.location.origin) {
  window.location.origin = window.location.protocol + "//" 
    + window.location.hostname 
    + (window.location.port ? ':' + window.location.port : '');
}
```

easel:
Fixes #201

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -4866,15 +4866,18 @@ trait Location extends js.Object {
    */
   var host: String = js.native
   /**
-   * Theorigin read-only property is a String containing the Unicode serialization of the
+   * The origin read-only property is a String containing the Unicode serialization of the
    * origin of the represented URL, that is, for http and https, the scheme followed by
    * '://', followed by the domain, followed by ':', followed by the port (the default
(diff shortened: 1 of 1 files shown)


mseddon:
I think as you said, this does seem to be an edge case. It's in both the [WHATWG spec](https://html.spec.whatwg.org/multipage/browsers.html#the-location-interface)  as well as [HTML 5](https://www.w3.org/TR/html5/browsers.html#the-location-interface), and requiriing an import just for `window.location.origin` is a bit clunky.  This should work well enough with the additional comment about IE11 compatibility.

sjrd:
LGTM

easel:
Thanks guys

PR #208, opened 2016-02-05
Add experimental PointerLock and Fullscreen API

danielwegener:
That's the last one for the moment :) (promised!).

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/Fullscreen.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/Fullscreen.scala
@@ -1,0 +1,83 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom.raw.{Event, Element, Document, Promise}
+
+import scala.language.implicitConversions
+import scala.scalajs.js
+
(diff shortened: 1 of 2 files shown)


danielwegener:
The fullscreen whatwg spec seem to be a little ahead of the browsers. `Document.exitFullscreen` and `Element.requestFullscreen` seem not to return a promise but `Unit` (I tried to express that with the return type `Promise[Unit] | Unit`). 

`onfullscreenchanged`and `onfullscreenerror` do not seem to be implemented at all so I left them out.

sjrd:
> `Promise[Unit] | Unit`

Hum ... That's a very weird and unpractical result type :-s

I guess the least bad option would be `js.UndefOr[Promise[Unit]]`.

sjrd:
Would be nice to document why in the Scaladoc, btw.

danielwegener:
@sjrd Good point. Fixed. Added `onfullscreenchanged` and `onfullscreenerror` but added a note that they may not be available.

sjrd:
LGTM

PR #210, opened 2016-02-06
Fix #209. Add RTCDataChannelEvent. Add more ScalaDoc.

### LINKED ISSUE #209: Issues with RTCDataChannel API
Hi folks,

I just tried to migrate my data-channel playground app to current 0.9.0-SNAPSHOT using WebRTC API provided by #203 and found some small issues:

- `RTCPeerConnection.createDataChannel` is not callable, its missing its `RTCDataChannelInit` parameter
- `RTCPeerConnection.ondatachannel` does not emit an `RTCDataChannelEvent`, so one can not react on data channels requested by the peer.
- some properties on `RTCPeerConnection` are `val` but their value change over time. Wonder if it would make more sense to expose them as `def`.

danielwegener:
Promise broken, here comes another PR :)


sjrd:
LGTM

PR #211, opened 2016-02-12
Upgrade to scalajs-0.6.7

danielwegener:
Remove dom.raw.Promise. Remove invalid js.native annotations.

### DIFF
--- a/project/build.sbt
+++ b/project/build.sbt
@@ -1,3 +1,3 @@
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.5")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.7")
 
 addSbtPlugin("com.lihaoyi" % "scalatex-sbt-plugin" % "0.2.1")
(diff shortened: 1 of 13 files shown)


sjrd:
> Upgrade to scalajs-0.6.7. Remove dom.raw.Promise. Remove invalid js.native annotations.

Those 3 sentences sound very much like 3 commits to me ;)

danielwegener:
Yeah but each committed individually would produce two broken commits :)

sjrd:
1. Remove invalid annotations
2. Upgrade
3. Use `js.Promise`

Doesn't that work?

danielwegener:
Aww but not as it really happened! ;) Sure! If it makes thinks clearer.

sjrd:
There are some `Promise`-related changes in your first commit, which should be in the third commit.
That's all.

mseddon:
Should we perhaps have a type alias where the old Promise was with a `@deprecated` message pointing to the new location?

sjrd:
Yes, good point.

danielwegener:
@mseddon @sjrd :ship: :it: !

sjrd:
LGTM

PR #214, opened 2016-02-16
Make Window.innerWidth and innerHeight Double

mseddon:
See: https://www.w3.org/TR/cssom-view/#extensions-to-the-window-interface

Note- this doesn't appear to be an issue in any current browser (I suppose viewports tend to be integer width in the underlying OS), but the spec does mention `Double`, not `Int`.


sjrd:
LGTM

PR #216, opened 2016-03-07
Add Intl API facade

soc:


src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,290 @@
+/*
+ * Implements the Intl API.
+ *
+ * [[http://tc39.github.io/ecma402/#intl-object The Intl Object]]
```
mseddon: I'd definitely also add the release date here- March 1, 2016, since this is a draft.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,290 @@
...
+  */
+@js.native
+object Intl extends js.Any {
+  /** The Intl.Collator object is a constructor for collators, objects that enable language
+    * sensitive string comparison.
+    *
+    * MDN
+    */
+  @js.native
+  class Collator(locales: js.UndefOr[String | js.Array[String]] = js.undefined, options: js.UndefOr[CollatorOptions] = js.undefined) extends js.Any {
```
mseddon: Break this line up to keep within 80 characters

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,290 @@
...
+    def compare(string1: String, string2: String): Double = js.native
```
mseddon: Can this return type be `Int`?

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,290 @@
...
+    def resolvedOptions(): CollatorOptions = js.native
```
mseddon: Here, and elsewhere, you should have a blank line to separate elements, see https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,290 @@
...
+ */
+
+package org.scalajs.dom.experimental.intl
+
+import scala.scalajs.js
+import scala.scalajs.js.|
+import scala.scalajs.js.annotation.{ScalaJSDefined, JSName}
+
+/** The Intl object is the namespace for the ECMAScript Internationalization API, which provides
+  * language sensitive string comparison, number formatting, and date and time formatting.
```
mseddon: Doc comments should be 'javadoc' style, rather than 'scaladoc' style.  i.e. the * on subsequent lines should be indented 1 space, rather than 2.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,290 @@
...
+    *                          are "upper", "lower", or "false" (use the locale's default); the
+    *                          default is "false". This option can be set through an options
+    *                          property or through a Unicode extension key; if both are provided,
+    *                          the options property takes precedence. Implementations are not
+    *                          required to support this property. 
+    *
+    * MDN
+    */
+  def apply(localeMatcher:     js.UndefOr[String]  = js.undefined,
+  	        usage:             js.UndefOr[String]  = js.undefined,
```
mseddon: There's a tab character here. evil! :)

mseddon:
That's all I have right now- it seems to be correct, but I forgot how horrible ECMAScript specifications were, so I may have missed something.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: Ok, will do.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: Documentation said returns "number" ...

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: You mean between the defs?

soc:
Thanks!

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: Done.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: UARGH. :-)
Ok, I'll uglify that.

soc:
Do the changes look fine? Will squash when you are happy.

soc:
Btw, would it make sense to move this stuff out of the `intl` package, given that the API is itself namespaced already? Instead of `intl.Intl.Collator` we could do `Intl.Collator`?

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,306 @@
...
+  /**
+   * The Intl.Collator object is a constructor for collators, objects that enable language
+   * sensitive string comparison.
+   *
+   * MDN
+   */
+  @js.native
+  class Collator(
+      locales: js.UndefOr[String | js.Array[String]] = js.undefined,
+      options: js.UndefOr[CollatorOptions] = js.undefined) extends js.Any {
```
sjrd: You probably want to `extends js.Object` here (and in most such places).

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,306 @@
...
+    def compare(string1: String, string2: String): Double = js.native
+    def resolvedOptions(): CollatorOptions = js.native
+    def supportedLocalesOf(locales: String | js.Array[String], options: js.Any): js.Array[String] = js.native
```
sjrd: You can break before `options` with 4 spaces of indentation to keep this from being too long.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,306 @@
...
+}
+
+@js.native
+trait CollatorOptions extends js.Object {
+  var localeMatcher:     js.UndefOr[String]
+  var usage:             js.UndefOr[String]
+  var sensivity:         js.UndefOr[String]
+  var ignore­Punctua­tion: js.UndefOr[Boolean]
+  var numeric:           js.UndefOr[Boolean]
+  var caseFirst:         js.UndefOr[String]
```
sjrd: Don't use vertical alignment.

Also, these should probably be `val`s.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -0,0 +1,306 @@
...
+   *
+   * MDN
+   */
+  def apply(
+      localeMatcher:     js.UndefOr[String]  = js.undefined,
+      usage:             js.UndefOr[String]  = js.undefined,
+      sensivity:         js.UndefOr[String]  = js.undefined,
+      ignore­Punctua­tion: js.UndefOr[Boolean] = js.undefined,
+      numeric:           js.UndefOr[Boolean] = js.undefined,
+      caseFirst:         js.UndefOr[String]  = js.undefined) =
```
sjrd: If the right-hand-side of an `=` (such as the body of a method) has more than 1 line, it should have `{}`.

Also, public methods should have an explicit result type.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: Are you sure? I tried changing these fields and it worked.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
soc: Is there a rule of thumb for that?

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
sjrd: Of course, a lot of things *work* in JS. That doesn't mean they're *meant* to work. There are a lot of fields that we define as `def` or `val` instead of `var`s because they're not meant to be modified.

src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
sjrd: Does the class extend the JavaScript class `Object`? Then `extends js.Object`. Otherwise `js.Any`.
Virtually all JS classes extend `Object`, so it's virtually always `extends js.Object`.

soc: Ok, thanks!

soc:
Better?

gzm0:
> Btw, would it make sense to move this stuff out of the intl package, given that the API is itself namespaced already? Instead of intl.Intl.Collator we could just have Intl.Collator?

Why not `intl.Collator` (and use a `@JSName`)? Seems most fluent in Scala.

soc:
@sjrd @gzm0 is this better?

soc:
Use https://github.com/soc/scala-js-dom/commit/779d72402f4a599a54c8c62ae1fadbc82196fc04?w=1 to ignore whitespace changes.

sjrd:
Yep, better. 

My earlier comments on `val`s and (no) vertical alignment still stand.

soc:
@sjrd ok, thanks, will do that!

soc:
@sjrd Done, does this look fine now?

sjrd:
OK could you just squash all the commits, please?

soc:
@sjrd Done!

sjrd:
LGTM, thanks!

PR #218, opened 2016-03-18
un-deprecate AudioBufferSourceNode.buffer

julienrf:
It seems that it is the way to go according to the [spec](https://www.w3.org/TR/webaudio/#widl-AudioBufferSourceNode-buffer).


mseddon:
Huh, you are right.  I'm unsure why that got deprecated.  We also need to move WebAudio eventually into`experimental` at some point because it's not [supported in IE](http://caniuse.com/#feat=audio-api).

sjrd:
LGTM

PR #222, opened 2016-04-15
Configure external links in ScalaDoc

Atry:


sjrd:
Thanks. API mapping is good but ... I don't want to add an external dependency for such a simple thing. I would much rather have the appropriate settings in our `build.sbt`.

We're extremely resistant to external dependencies in scala-js/scala-js and companion repositories, because we sit *very* high in the dependency graph of the Scala eco-system. If we are blocked from releasing a new version because a dependency is not ready yet for some new Scala version, we block the entire Scala eco-system (as was demonstrated last week with 2.12.0-M4 and me being on vacation).

Atry:
I understand your concern. But... would it be acceptable to add dependencies in a sbt plugin? IIRC, It would not break down this library when scala version is changed because the scala version for sbt is still 2.10.x .

sjrd:
Indeed, the scenario I expose would not happen for an sbt plugin. There would be a breakage only when sbt breaks binary compatibility. But still, we try very hard to avoid as many Scala dependencies as we can. In the Scala.js repo, we only have [4 sbt plugins](https://github.com/scala-js/scala-js/blob/master/project/build.sbt#L6-L12), only 1 of which is critical to build and ship (the Bintray one). And all 4 of those take care of a *lot* of work behind the scenes. It's not like they provide a feature that we can reproduce with ~4 LoC of settings.

Every external dependency on something non-Java and non-Node.js is a heavy burden for us. So the feature it provides has to pay off significantly for us to consider it. In this case, the benefits do not outweigh the cost.

Atry:
Fair enough. Feel free to copy sbt configuration from my plugin.

PR #224, opened 2016-04-25
Fix Worker signatures

japgolly:
The doc and second-argument were incorrect.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage


japgolly:
Also fixed `Worker.onmessage` signature as per https://developer.mozilla.org/en-US/docs/Web/API/Worker/onmessage.

mseddon:
https://www.w3.org/TR/workers/#dom-dedicatedworkerglobalscope-postmessage LGTM

sjrd:
LGTM

PR #227, opened 2016-05-17
Add a few missing methods/fields to AudioContext

ramnivas:
Add the close(), suspend(), and resume() methods as well
as the state field to the AudioContext.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/Audio.scala
+++ b/src/main/scala/org/scalajs/dom/raw/Audio.scala
@@ -47,6 +47,14 @@ class AudioContext extends EventTarget {
    */
   val sampleRate: Double = js.native
 
+  /** Returns the current state of the AudioContext. */
+  def state: String = js.native
+
+  /** Closes the audio context, releasing any system audio resources
+   *  that it uses.
+   */
+  def close(): js.Promise[Unit] = js.native
+
   /** Creates an AnalyserNode, which can be used to expose audio time and
    *  frequency data and for example to create data visualisations.
    */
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -47,6 +47,14 @@ class AudioContext extends EventTarget {
    */
   val sampleRate: Double = js.native
 
+  /** Returns the current state of the AudioContext. */
+  val state: String = js.native
```
sjrd: I guess this should be a `def`.

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -47,6 +47,14 @@ class AudioContext extends EventTarget {
...
+
+  /** Closes the audio context, releasing any system audio resources
+   * that it uses.
```
sjrd: missing space before `that`

sjrd:
Could you give a link to the spec of these members?

ramnivas:
@sjrd Link to the spec: https://developer.mozilla.org/en-US/docs/Web/API/AudioContext

src/main/scala/org/scalajs/dom/raw/Audio.scala
```diff
@@ -47,6 +47,14 @@ class AudioContext extends EventTarget {
    */
   val sampleRate: Double = js.native
 
+  /** Returns the current state of the AudioContext. */
+  def state: String = js.native
+
+  /** Closes the audio context, releasing any system audio resources
+   *  that it uses.
+   */
+  def close(): Unit = js.native
```
sjrd: From the spec, this also returns a `js.Promise[Unit]`.

sjrd:
That's all.

mseddon:
Not around atm, but if it corresponds to [this](https://www.w3.org/TR/2015/WD-webaudio-20151208/) (or a previous version) then it should be good.  MDN can be a little out of date at times.

sjrd:
LGTM

PR #228, opened 2016-05-20
Override currentTarget

Atry:
This will avoid unnecessary type cast like `event.currentTarget.asInstanceOf[Input].value`

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -641,7 +641,43 @@ trait NonDocumentTypeChildNode extends js.Object {
 @js.native
 abstract class Element extends Node with NodeSelector with ParentNode with NonDocumentTypeChildNode {
 
-
+  protected type Event = org.scalajs.dom.raw.Event {def currentTarget: this.type}
+  protected type AnimationEvent = org.scalajs.dom.raw.AnimationEvent {def currentTarget: this.type}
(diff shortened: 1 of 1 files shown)


Atry:
@sjrd Could you review this PR, please?

sjrd:
I'm very uneasy about *shadowing* the names of event classes within the definitions of `Element` classes.  This seems to invite a lot of confusion, and does not look like a sane design.

I'm sympathetic to the idea of making scalajs-dom more tightly typed, but IMO this should be done across the board by careful, integrated design. For example, ideally something like `x.addEventListener("mouseup", (e: MouseEvent) => ...)` should typecheck.

Wrt your use case, it seems to me that there is a much easier (and more readable) solution, which is simply to use the `val` of whatever element your attaching the event to. For example:
```scala
val b = document.createElement("input").asInstanceOf[html.Input]
b.addEventListener("keyup", { (e: dom.Event) =>
  val s = b.value // instead of `e.currentTarget.asInstanceOf[html.Input].value`
}
```

Atry:
I heavily used `currentTarget` in Binding.scala's TodoMVC DEMO:

https://github.com/ThoughtWorksInc/todo/blob/master/js/src/main/scala/com/thoughtworks/todo/Main.scala#L109

I think assigning elements to a temporary `val` will make these HTML templates too complicated.

sjrd:
Closing this as part of an effort to close old PRs and be back on a sane basis. Something that deep would require careful analysis, and maybe a larger impact on all sides of the API.

PR #229, opened 2016-05-23
Export AjaxException members to better support Future.toPromise and Ajax Futures.

russwyte:
When using the rich future JSConverter to return a JS Promise from a Future that performs an AJAX request it is possible that the JS caller will be returned an AjaxException to handle in a rejected promise. In these cases is it useful for the JS client to have access to the XHR request itself to better handle the failure. This change simply makes the XHR that caused the exception available to the JS caller.

### DIFF
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -3,15 +3,15 @@ package org.scalajs.dom.ext
 import java.nio.ByteBuffer
 
 import scala.language.implicitConversions
-import scala.concurrent.{Promise, Future}
-
+import scala.concurrent.{Future, Promise}
 import scala.scalajs.js
 import scala.scalajs.js.typedarray._
 import scala.scalajs.js.typedarray.TypedArrayBufferOps._
-
 import org.scalajs.dom
 import org.scalajs.dom.{FormData, html, raw}
 import org.scalajs.dom.raw.Blob
+
+import scala.scalajs.js.annotation.{JSExport, JSExportAll}
 
 
 /**
@@ -234,6 +234,8 @@ object KeyCode {
  * Thrown when `Ajax.get` or `Ajax.post` receives a non-20X response code.
  * Contains the XMLHttpRequest that resulted in that response
  */
+@JSExport
+@JSExportAll
 case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
   def isTimeout = xhr.status == 0 && xhr.readyState == 4
 }
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -11,7 +11,7 @@ import org.scalajs.dom
 import org.scalajs.dom.{FormData, html, raw}
 import org.scalajs.dom.raw.Blob
 
-import scala.scalajs.js.annotation.{JSExport, JSExportAll}
+import scala.scalajs.js.annotation.JSExportAll
 
 
 /**
@@ -234,7 +234,6 @@ object KeyCode {
  * Thrown when `Ajax.get` or `Ajax.post` receives a non-20X response code.
  * Contains the XMLHttpRequest that resulted in that response
  */
-@JSExport
 @JSExportAll
 case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
   def isTimeout = xhr.status == 0 && xhr.readyState == 4


sjrd:
What's the reason for `@JSExport`? I understand from the PR message (which should be in the commit message, btw) that `@JSExportAll` is useful, but not `@JSExport`.

Exports in libraries should be used with great care, as they mean code that will always be included in the emitted JS files of any application using that library, even if they are not used.

russwyte:
Thanks. 

Good points. I will add that kind of detail to future PR commits. 

My mistake on @JSExport I thought it was always required when exporting members.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -3,16 +3,16 @@ package org.scalajs.dom.ext
 import java.nio.ByteBuffer
 
 import scala.language.implicitConversions
-import scala.concurrent.{Promise, Future}
-
+import scala.concurrent.{Future, Promise}
 import scala.scalajs.js
 import scala.scalajs.js.typedarray._
 import scala.scalajs.js.typedarray.TypedArrayBufferOps._
-
```
sjrd: The changes here and just above are useless and should not be part of this commit/PR.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -234,6 +234,7 @@ object KeyCode {
  * Thrown when `Ajax.get` or `Ajax.post` receives a non-20X response code.
  * Contains the XMLHttpRequest that resulted in that response
  */
+@JSExportAll
 case class AjaxException(xhr: dom.XMLHttpRequest) extends Exception {
   def isTimeout = xhr.status == 0 && xhr.readyState == 4
```
sjrd: In fact, I now realize this a `case class`. Case classes have several "hidden" public members such as `copy`, `hashCode`, which we do not really want to be exported. I recommend to only export `xhr` itself. Possibly `isTimeout`, but I would refrain from doing so, in order to minimize what is exported.

sjrd:
I'm wondering how much this is really needed/appropriate for this library. scalajs-dom is not really targeted at providing a nice *JavaScript* API for its things. I'm worried that including this change might set a precedent for exporting stuff, which would then justify exporting more stuff in all kinds of place in this API. Eventually leading to code size bloat for users of the library that do not even care about exporting a JS API.

IMO this kind of features should be left to applications or dedicated libraries. For this particular use case, I would recommend something like the following user-space pimp:
```scala
implicit class AjaxExceptionFutureOps(val self: Future[AjaxException]) extends AnyVal {
  def toXHRPromise: js.Promise[XMLHttpRequest] =
    self.transform(identity, ex => js.JavaScriptException(ex.xhr))
}
```
You can then use `fut.toXHRPromise` to get a `js.Promise[XMLHttpRequest]`, without needing any export. This is also slightly more idiomatic, IMO: it is often better to provide *conversions* from Scala types to JS types, rather than exporting JS APIs in Scala types. Except of course for Scala classes that are *dedicated* to provide a JS API, but that is not the case of `AjaxException`.

russwyte:
I actually like your last suggestion very much. It might be preferable if the default toPromise transformation worked like that - because the AjaxException does not have a lot of value as exported JS - and that is the whole point of toPromise.

sjrd:
> It might be preferable if the default toPromise transformation worked like that

Hum ... no, it wouldn't. `.toJSPromise` converts the `Future` into a `js.Promise`. It does not touch the wrapped value. That's the same as all other conversions in the world, e.g., `seq.toJSArray`. Even if it were desirable, you cannot have `.toJSPromise` deal with all the possible "convenient" deep conversions in the world and beyond.

russwyte:
I only meant when used with a future obtained via Ajax.get .post etc... Not for all Futures.

sjrd:
How would you implement that?

russwyte:
I might change to this impl of "onreadystatechange":

`    req.onreadystatechange = {(e: dom.Event) =>
      if (req.readyState.toInt == 4){
        if ((req.status >= 200 && req.status < 300) || req.status == 304)
          promise.success(req)
        else
          promise.failure(js.JavaScriptException(req))
      }
    }`

Because we know a js.Promise is something only external JS cares about we should not use type that is obtuse to JS like the current AjaxException - it carries no useful information about the failure - but instead something with the XHR accessible.

sjrd:
But most people use `Ajax.get` from *Scala* code and want to see a Scala exception that can be pattern matched against in a `catch` clause. Your use case is comparatively much rarer.

russwyte:
Yes - I actually agree with you. I think for now I will go the implicit conversion route you described earlier - but I think it would be great to expose a bit more (at least the xhr) to the JS clients without hoops if possible. If I can think of a better way to do it I will propose it later.

Thanks for the information and insight into the library strategy.

PR #230, opened 2016-05-24
Expose existing URL trait as window.URL

netzwerg:
While the dom package object defines a URL type, its methods are not actually reachable.

This PR exposes it as `window.URL`

I hope this is the right place and style. Feels weird to use all-caps for a def, but afaik native JS classes may not have inner objects, right?


sjrd:
This received a better fix in ac070dbf6029da41d2d6a464b59aa917b4d939bc.

PR #235, opened 2016-07-10
Add URL constructor

turbolent:
See https://url.spec.whatwg.org/#api and 
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/URL.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/URL.scala
@@ -1,0 +1,93 @@
+package org.scalajs.dom.experimental
+
+import scala.scalajs.js
+
+/**
+  * The URL() constructor returns a newly created URL object representing the URL
+  * defined by the parameters.
+  *
+  * MDN
+  */
+@js.native
+class URL(url: String, base: String = js.native) extends js.Object {
+
+  /**
+    * Returns a DOMString containing the origin of the URL, that is its scheme,
+    * its domain and its port.
+    *
+    * MDN
+    */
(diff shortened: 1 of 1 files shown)


sjrd:
Given that the `URL` constructor does not seem to be supported in IE yet, this should go [in the `experimental` package](https://github.com/scala-js/scala-js-dom/tree/master/src/main/scala/org/scalajs/dom/experimental), in a file `URL.scala`. It's weird that the `object URL` and `class URL` won't live in the same package, but that reflects their respective level of support.

sjrd:
That's all.

turbolent:
Thanks for the feedback! Moved the class to the experimental package as suggested.

src/main/scala/org/scalajs/dom/experimental/URL.scala
```diff
@@ -0,0 +1,93 @@
...
+    */
+  var search: String = js.native
+
+  /**
+    * Is a DOMString containing a '#' followed by the fragment identifier of the URL.
+    *
+    * MDN
+    */
+  var hash: String = js.native
+}
```
sjrd: There's a missing EOL at the end of the file here.

sjrd:
Otherwise good for me. Could you squash your commits into a single commit, please?

turbolent:
Squashed. You can also now do this straight from GitHub BTW: https://help.github.com/articles/about-pull-request-merge-squashing/

sjrd:
LGTM. 

GitHub does not let me squash your commits *and* create a merge commit for merging the PR, unfortunately.

PR #237, opened 2016-07-24
Add DOMParser

soc:


src/main/scala/org/scalajs/dom/experimental/domparser/DOMParser.scala
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * The DOMParser can also be used to parse a SVG document or a HTML document.
+   * There are three different results possible, selected by the MIME type given.
+   * If the MIME type is text/xml, the resulting object will be an XMLDocument,
+   * if the MIME type is image/svg+xml, it will be an SVGDocument and
+   * if the MIME type is text/html, it will be an HTMLDocument.
+   *
+   * MDN
+   */
+  def parseFromString(string: String, supportedType: SupportedType): Document | HTMLDocument = js.native
```
sjrd: Since `HTMLDocument` extends `Document`, the union type in the result type is not helping. It's better to just have `Document`, if you're not going to use phantom types to refine it.

sjrd:
That's all.

soc:
Done!

sjrd:
LGTM

PR #241, opened 2016-08-16
File size as Double

Starzu:
It's possible to select a file bigger than 2GB.


sjrd:
This is most certainly incorrect, as Longs are not JavaScript numbers. If JS numbers larger than `Int.MaxValue` are to be accepted, then the type must be `Double`.

Starzu:
@sjrd Right, my bad. ;)

ritschwumm:
i think i'd still like an Int53 type i could use in these situations...

mseddon:
@ritschwumm see https://github.com/scala-js/scala-js/issues/2170 for a possible implementation.

sjrd:
LGTM

PR #242, opened 2016-09-28
Support assigning a String to a style property

Atry:


src/main/scala/org/scalajs/dom/raw/Html.scala
```diff
@@ -3968,7 +3969,8 @@ abstract class HTMLElement extends Element {
    *
    * MDN
    */
-  var style: CSSStyleDeclaration = js.native
+  def style: CSSStyleDeclaration = js.native
+  def style_=(value: String | CSSStyleDeclaration): Unit = js.native
```
sjrd: I think overloads would be more appropriate, here.

sjrd:
OK, could you squash the three commits, and keep only the first commit's message?

Atry:
@sjrd Done. Waiting for CI now.

sjrd:
LGTM

Atry:
@sjrd Some of Travis CI jobs failed due to network problem: https://travis-ci.org/scala-js/scala-js-dom/jobs/163351090

PR #251, opened 2016-11-23
Fix SVGTextContentElement to use correct return types

mmx900:
Reference : https://www.w3.org/TR/SVG/text.html#InterfaceSVGTextContentElement
MDN : https://developer.mozilla.org/en-US/docs/Web/API/SVGTextContentElement


sjrd:
I believe the second commit shouldn't be part of this PR. Its content seems to have been included as part of #316.

Closing this as part of an effort to close old PRs and be back on a sane basis. Could be brought back to life (with a rebase at least).

PR #255, opened 2016-12-13
Deprecate some non-standard and obsolete types and fields.

danielnixon:
I had to turn off fatal warnings because the type aliases refer to the deprecated types. It might be better to remove the type aliases entirely, or even skip deprecation and just remove the offending types and fields immediately.

See:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bgsound
* https://msdn.microsoft.com/en-us/library/windows/apps/hh965347.aspx
* https://developer.mozilla.org/en-US/docs/Web/API/window/setImmediate
* https://developer.mozilla.org/en-US/docs/Web/API/Window/clearImmediate
* https://developer.mozilla.org/en/docs/Web/API/Event/srcElement
* https://msdn.microsoft.com/en-us/library/hh869721(v=vs.85).aspx
* https://www.w3.org/TR/html5/obsolete.html
* https://developer.mozilla.org/en-US/docs/Web/API/HTMLFontElement

### DIFF
--- a/src/main/scala/org/scalajs/dom/html.scala
+++ b/src/main/scala/org/scalajs/dom/html.scala
@@ -5,12 +5,15 @@ package org.scalajs.dom
  */
 object html {
   type Anchor = raw.HTMLAnchorElement
+  @deprecated("Obsolete.", "HTML 5")
   type Applet = raw.HTMLAppletElement
   type Audio = raw.HTMLAudioElement
   type Area = raw.HTMLAreaElement
   type AreasCollection = raw.HTMLAreasCollection
   type Base = raw.HTMLBaseElement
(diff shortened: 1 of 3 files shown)


sjrd:
You should deprecate the type aliases instead of disabling fatal warnings.

danielnixon:
Done (and squashed).

sjrd:
The first two commits should be squashed together, because the first one doesn't build on its own.

The commit message of the third commit doesn't make sense to me wrt. the contents of the commit.

Maybe it's better to just squash everything together, here, as everything is about deprecations anyway.

danielnixon:
I think the build failure was because Travis couldn't resolve some dependencies.

Everything squashed. Thanks @sjrd !

PR #265, opened 2017-01-21
Move href from CSSStyleSheet to StyleSheet.

danielnixon:
See:
* https://drafts.csswg.org/cssom/#the-stylesheet-interface
* https://drafts.csswg.org/cssom/#the-cssstylesheet-interface


sjrd:
[error] (run-main-0) org.scalafmt.Error$MisformattedFile: /home/travis/build/scala-js/scala-js-dom/src/main/scala/org/scalajs/dom/raw/lib.scala is mis-formatted

PR #271, opened 2017-02-27
Fix type of CustomEvent detail attribute

andersschuller:
See https://dom.spec.whatwg.org/#interface-customevent

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -5091,7 +5091,7 @@ class CustomEvent extends Event {
    *
    * MDN
    */
-  def detail: Object = js.native
+  def detail: Any = js.native
 
   /**
    * Initializes the event in a manner analogous to the similarly-named method in the
@@ -5100,7 +5100,7 @@ class CustomEvent extends Event {
    * MDN
    */
   def initCustomEvent(typeArg: String, canBubbleArg: Boolean,
-      cancelableArg: Boolean, detailArg: Object): Unit = js.native
+      cancelableArg: Boolean, detailArg: Any): Unit = js.native
 }
 
 /**


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -5091,7 +5091,7 @@ class CustomEvent extends Event {
    *
    * MDN
    */
-  def detail: Object = js.native
+  def detail: js.Any = js.native
```
sjrd: It should be `Any` then, not `js.Any`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -5100,7 +5100,7 @@ class CustomEvent extends Event {
    * MDN
    */
   def initCustomEvent(typeArg: String, canBubbleArg: Boolean,
-      cancelableArg: Boolean, detailArg: Object): Unit = js.native
+      cancelableArg: Boolean, detailArg: js.Any): Unit = js.native
```
sjrd: Same.

src/main/scala/org/scalajs/dom/raw/lib.scala
andersschuller: Hm. I guess I assumed it should be `js.Any` since it seemed to be consistent with most other similar attributes/methods. For example, the `History` API has the following:

```
def state: js.Any = js.native
def pushState(statedata: js.Any, title: String): Unit = js.native
```

There only seem to be a handful of uses of `java.lang.Object`, but now that I look more closely, there are quite a few uses of `scala.Any`. There are even a few cases that are not consistent within the same class, for example `MessageEvent`. This uses `scala.Any` for the `data` attribute but `js.Any` for the corresponding argument to `initMessageEvent`.

I'm not sure I understand why all these would be slighly different. Is it just an artifact of having imported them from the TypeScript type definitions? I'm happy to make the change you requested, of course, just thought I would clarify first.

sjrd: Most `js.Any`s are artifacts of importing from TypeScript, and should be `Any`.

andersschuller:
OK, commit amended with the requested changes.

PR #281, opened 2017-04-18
Add HTMLTemplateElement interface

galedric:
Facade for the HTMLTemplateElement interface based on: https://developer.mozilla.org/en-US/docs/Web/API/HTMLTemplateElement

Also a question: why are read-only properties `def`s and not `val`s ?

### DIFF
--- a/src/main/scala/org/scalajs/dom/html.scala
+++ b/src/main/scala/org/scalajs/dom/html.scala
@@ -86,6 +86,7 @@ object html {
   type TableHeaderCell = raw.HTMLTableHeaderCellElement
   type TableRow = raw.HTMLTableRowElement
   type TableSection = raw.HTMLTableSectionElement
+  type Template = raw.HTMLTemplateElement
   type Title = raw.HTMLTitleElement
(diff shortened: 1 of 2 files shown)


ritschwumm:
probably because read-only does not mean the value never changes

galedric:
Well, now that you've said it, it seems pretty obvious.

I also don't understand what is wrong with the formatting of `Html.scala`. The reported error in the Travis log is not very helpful and running `sbt scalafmtTest` doesn't fail locally.

ritschwumm:
true, https://travis-ci.org/scala-js/scala-js-dom/jobs/223201059#L640 is a bit lacking in detail

sjrd:
I don't know why `sbt scalafmtTest` doesn't fail locally, but your issue is that you have the wrong Scaladoc style. Look around in the file to see how Scaladoc comments are formatted.

sjrd:
Closing this as part of an effort to close old PRs and be back on a sane basis. Could be brought back to life (with a rebase at least).

PR #291, opened 2017-07-18
Add window.scrollX and window.scrollY

MaximeKjaer:
While [`window.pageXOffset`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageXOffset) and [`window.pageYOffset`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset) are implemented, MDN describes them as aliases for the main properties [`window.scrollX`](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX) and [`window.scrollY`](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY), which were not implemented.

As a side note, a quick GitHub search shows that [the alias](https://github.com/search?l=JavaScript&q=%22window.pageXOffset%22&type=Code&utf8=%E2%9C%93) is actually more used than the [primary property](https://github.com/search?l=JavaScript&q=%22window.scrollX%22&type=Code&utf8=%E2%9C%93), but given how common these are, they should both be supported.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -1797,11 +1797,19 @@ class Window
 
   /**
    * Returns the number of pixels that the document has already been scrolled
-   * horizontally.
+   * horizontally. An alias for window.scrollX.
    *
    * MDN
    */
   def pageXOffset: Double = js.native
+
+  /**
+    * Returns the number of pixels that the document has already been scrolled
+    * horizontally.
+    *
+    * MDN
+    */
+  def scrollX: Double = js.native
 
   /**
    * The name of the window is used primarily for setting targets for hyperlinks and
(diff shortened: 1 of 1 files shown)


sjrd:
MDN says:

> For cross-browser compatibility, use `window.pageXOffset` instead of `window.scrollX`.

It seems adding those properties is just giving more rope to the developer to hang themselves.

MaximeKjaer:
Hmm, indeed, I missed that in the documentation. That would explain why `pageXOffset` is more used. As follow-up questions then:

- Is it within the scope of this project to redirect `scrollX` to `pageXOffset` in situations like this one, or is it best to mirror the sometimes quirky behavior of browsers, and not to introduce quirks here?
- What's the policy on implementing non-standardized API ([all yellow icons here](https://developer.mozilla.org/en-US/docs/Web/API/Window) for instance)? Is the goal always maximum cross-browser compatibility or should it be up to the developer to figure that out on their own?

sjrd:
> Is it within the scope of this project to redirect scrollX to pageXOffset in situations like this one?

No, we don't do that in this project. Methods and properties stand for themselves, not special redirects.

> What's the policy on implementing non-standardized API?

Only standardized APIs are implemented in scalajs-dom. Sometimes, APIs still in the process of standardization (but well underway and close to finalization) are proposed inside the `experimental` package. So those yellow icons definitely don't make it into scalajs-dom.

Moreover, we usually restrict definitions to standardized APIs already widely supported, unless they have no alternative. So here `scrollX` is standardized, but not completely supported, and it has an alternative, so it would typically not make it into scalajs-dom.

MaximeKjaer:
All right, that clears it up for me. Thank you!

PR #293, opened 2017-07-24
Add HTMLLinkElement.{as,crossOrigin}

japgolly:


sjrd:
`crossOrigin` is marked as experimental [on MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLinkElement), so it shouldn't be here.

`as` has been specified, so it's OK, but it's worth noting that support in Firefox is very recent, and in Edge it is unknown whether it is supported or not.

PR #297, opened 2017-10-09
Fix #295: Change WebSocket.onerror to use Event instead of ErrorEvent

### LINKED ISSUE #295: ErrorEvent sent to WebSocket.onError doesn't have "message" on Firefox
When the `onerror` callback of the following code gets invoked:
```
def onerror(e: ErrorEvent): Unit = {
  global.console.log(e)
  val msg:String = e.message
}
val ws = new WebSocket(someURLthatDoesntWork)
ws.onerror = onerror _
```
The `e.message` expression fails on Firefox at runtime, since its `ErrorEvent` (in this case, apparently) doesn't have a `message` member: (from the firefox console)
```
error { target: WebSocket, isTrusted: true, currentTarget: WebSocket, eventPhase: 2, 
bubbles: false, cancelable: false, defaultPrevented: false, composed: false, 
timeStamp: 2195.899715, cancelBubble: false, originalTarget: WebSocket }
```
It results in an `uncaught exception: scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: undefined is not an instance of java.lang.String`, which however somehow doesn't even get printed until the page is reloaded.

I'm unsure whether [`ErrorEvent`](https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/raw/lib.scala#L6256) itself is modelled incorrectly, whether `WebSocket.onerror` should have a different type, or whether coercing `undefined` to a `String` should have a bit better semantics that having to throw `js.isUndefined` around all over the place...

andersschuller:


andersschuller:
See the spec linked in the original issue, as well as https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#Attributes

> An event listener to be called when an error occurs. This is a simple event named "error".

PR #298, opened 2017-10-22
Fix #287: Add error parameter to window.onerror

### LINKED ISSUE #287: window.onerror error parameter
MDN documentation says:

```
window.onerror = function(messageOrEvent, source, lineno, colno, error) { ... }
```
(see https://developer.mozilla.org/en/docs/Web/API/GlobalEventHandlers/onerror)
but scala-js-dom leaves off the last parameter.

andersschuller:


andersschuller:
Note: the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) suggests that the type of the `error` parameter should be `Error`. However, in my testing I found that this was only the case if the error was thrown specifically using `throw new Error("Some message")`, but not if for example an uncaught exception was thrown from Scala code or if an error was thrown like `throw "Some message"`. That is why I decided to use type `Any` for the new parameter.

PR #299, opened 2017-11-13
Add basic support for the Permissions API

Grogs:
Supported in Chrome and Firefox. https://developer.mozilla.org/en-US/docs/Web/API/Permissions

Not complete, but usable. 

Example usage:
`navigator.permissions.query(js.Dynamic.literal(name = "geolocation")).toFuture`

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
@@ -1,0 +1,42 @@
+package org.scalajs.dom.experimental
+
+import language.implicitConversions
+import org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.ScalaJSDefined
+
+/**
+  * Implements the Permissions API.
+  *
+  * [[https://www.w3.org/TR/permissions/ W3C Working Draft]]
+  */
+package object permissions {
+
+  @js.native
+  trait PermissionState extends js.Any
+
+  final val `granted` = "granted".asInstanceOf[PermissionState]
+  final val `denied` = "denied".asInstanceOf[PermissionState]
+  final val `prompt` = "prompt".asInstanceOf[PermissionState]
+
+  @ScalaJSDefined
+  trait PermissionStatus extends dom.raw.EventTarget {
+    val state: PermissionState
+    val onchange: js.Function1[PermissionState, _]
+  }
+
+  @ScalaJSDefined
+  trait Permissions extends js.Any {
+    def query(permissionDescriptor: js.Object): js.Promise[PermissionStatus]
+  }
+
+  @ScalaJSDefined
+  trait PermissionsNavigator extends js.Any {
+    val permissions: Permissions
+  }
+
+  implicit def toPermissions(navigator: dom.raw.Navigator): PermissionsNavigator =
+    navigator.asInstanceOf[PermissionsNavigator]
+
+}
--- a/src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
@@ -1,10 +1,10 @@
 package org.scalajs.dom.experimental
 
-import language.implicitConversions
+import scala.language.implicitConversions
+
 import org.scalajs.dom
 
 import scala.scalajs.js
-import scala.scalajs.js.annotation.ScalaJSDefined
 
 /**
   * Implements the Permissions API.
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,42 @@
...
+package object permissions {
+
+  @js.native
+  trait PermissionState extends js.Any
+
+  final val `granted` = "granted".asInstanceOf[PermissionState]
+  final val `denied` = "denied".asInstanceOf[PermissionState]
+  final val `prompt` = "prompt".asInstanceOf[PermissionState]
+
+  @ScalaJSDefined
```
sjrd: `@ScalaJSDefined` should be removed. This build has `-P:scalajs:sjsDefinedByDefault`.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,42 @@
...
+  * [[https://www.w3.org/TR/permissions/ W3C Working Draft]]
+  */
+package object permissions {
+
+  @js.native
+  trait PermissionState extends js.Any
+
+  final val `granted` = "granted".asInstanceOf[PermissionState]
+  final val `denied` = "denied".asInstanceOf[PermissionState]
+  final val `prompt` = "prompt".asInstanceOf[PermissionState]
```
sjrd: These vals don't need the backquotes.

sjrd: It would be better to put them in an `object PermissionState`, in order to pollute the top-level namespace.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  @ScalaJSDefined
+  trait PermissionStatus extends dom.raw.EventTarget {
+    val state: PermissionState
+    val onchange: js.Function1[PermissionState, _]
+  }
+
+  @ScalaJSDefined
+  trait Permissions extends js.Any {
+    def query(permissionDescriptor: js.Object): js.Promise[PermissionStatus]
```
sjrd: `js.Object` is way too loose. You should define something like:
```scala
trait PermissionDescriptor extends js.Object {
  val name: PermissionName
}

sealed trait PermissionName extends js.Any

object PermissionName {
  val geolocation: PermissionName = "geolocation".asInstanceOf[PermissionName]
  ...
}
```

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,42 @@
+package org.scalajs.dom.experimental
+
+import language.implicitConversions
```
sjrd: Please don't use imports that are relative to the `scala` package. Use the full name instead:
```scala
```
Also, add a blank line after this one.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.ScalaJSDefined
```
sjrd: This import won't be needed after the `@ScalaJSDefined` annotations have been removed.

Grogs:
Thanks for the code review feedback. Should be addressed now.

I also added an apply method to PermissionDescriptor's companion object to make construction easier.

Example usage now looks like so:
`navigator.permissions.query(PermissionDescriptor(PermissionName.geolocation)).toFuture`

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
Grogs: Updated. This was copied from `org.scalajs.dom.experimental.gamepad`.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
Grogs: Done.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
Grogs: Done. This was originally following the style in `org.scalajs.dom.experimental.gamepad`.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
Grogs: Done.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
Grogs: Done.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  object PermissionDescriptor {
+    @inline
+    def apply(permissionName: PermissionName): PermissionDescriptor = {
+      js.Dynamic
+        .literal(
+            name = permissionName
+        )
+        .asInstanceOf[PermissionDescriptor]
```
sjrd: You should use the typesafe API instead of `js.Dynamic` for the body of this method:
```scala
def apply(permissionName: PermissionName): PermissionDescriptor = {
  new PermissionDescriptor {
    val name = permissionName
  }
}
```

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Implements the Permissions API.
+ *
+ * [[https://www.w3.org/TR/permissions/ W3C Working Draft]]
+ */
+package object permissions {
+
+  @js.native
+  trait PermissionState extends js.Any
```
sjrd: I did not notice this earlier: this trait should be `sealed`.

Grogs: Updated. I was following the codestyle I saw in `mediastream.MediaDeviceInfo`.

I've also changed:
 -  the values of `PermissionState` are no longer final
 - `PermissionName`is sealed too

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
Grogs: Done. I was following the codestyle I saw in mediastream.MediaDeviceInfo.

Outside of this project (where I'm duplicating this until it's merged/released), this gave me an error,  I guess there's a compiler flag in place for this project...

```
Error:(49, 11) A Scala.js-defined JS class cannot directly extend a native JS trait.
      new PermissionDescriptor {
```

sjrd: > I guess there's a compiler flag in place for this project

Yes: `-P:scalajs:sjsDefinedByDefault`. It is recommended to use that flag nowadays, as it switches to the way Scala.js 1.x works.

PR #302, opened 2017-11-21
Add containsNode and extend methods to Selection

kdarkhan:


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1602,6 +1602,22 @@ class Selection extends js.Object {
    * MDN
    */
   def removeRange(range: Range): Unit = js.native
+
+  /**
+   * Indicates if the node is part of the selection
+   *
+   * MDN
+   */
+  def containsNode(node: Node, partialContainment: Boolean): Boolean = js.native
```
sjrd: Since `partialContainment` defaults to `false` in the API, you should write `partialContainment: Boolean = false` to allowing calling that method without an explicit second argument.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1602,6 +1602,22 @@ class Selection extends js.Object {
...
+
+  /**
+   * Moves the focus of the selection to a specified point. The anchor of the selection
+   * does not move. The selection will be from the anchor to the new focus regardless
+   * of direction.
+   *
+   * MDN
+   */
+  def extend(node: Node, offset: Int): Unit = js.native
```
sjrd: Same here with `offset: Int = 0`.

sjrd:
You will also need to run `sbt scalafmt` to format the code so that the CI is happy.

kdarkhan:
@sjrd , thanks! Just updated my branch.

Issue #44, opened 2014-07-23
ChildNode interface


japgolly:
Hey, is there any reason why this isn't typed yet?

>  https://developer.mozilla.org/en-US/docs/Web/API/ChildNode

If not I'll put a patch together.

sjrd:
3 years have passed, and this API is still experimental and not implemented by the majority of browsers. I believe we should close this as won't fix.

PR #306, opened 2017-12-01
add ChildNode interface #44

kdarkhan:


sjrd:
Hum, there seems to be no settled specification for this interface yet. It's only in Living Standard mode, and only 2 browsers implement any of it. See https://developer.mozilla.org/en-US/docs/Web/API/ChildNode

Maybe it's too early to put it in our library? Or we could put it in `experimental`, and not have its subclasses actually extend it, but having an implicit conversion in the companion object that casts them. That would allow people to use it by importing that conversion (on purpose, from the `experimental` package).

kdarkhan:
I see. I saw the hanging issue #44 and there was nothing about the experimental part.
I can move this to experimental or whatever you decide. Was writing a firefox plugin recently and using `currentNode.parentNode.removeChild(currentNode)` was way too verbose compared to `currentNode.remove()`

sjrd:
Yes, sorry, I must not have been very attentive when I commented on #44. It's actually surprising that this was 3 years ago, and other browsers still haven't implemented it. That's a red flag, as there's little hope that it will suddenly be implemented any time.

Maybe it would be best to close this PR and the issue as Won't fix, given the time elapsed. WDYT?

PR #313, opened 2018-02-24
Improve DOM_KEY_LOCATION_* bindings

tindzk:
The current naming of the object `KeyboardEvent` is slightly unfortunate as it only contains constants related to key locations.

Other wrappers call this object `KeyLocation`:

* https://api.dartlang.org/stable/1.24.3/dart-html/KeyLocation-class.html
* https://vaadin.com/api/7.6.5/elemental/events/KeyboardEvent.KeyLocation.html

For compatibility reasons, I kept the object and defined aliases in `ext`.

I was also considering to drop `js.native` since the standard defines the constants as follows: 

```scala
val DOM_KEY_LOCATION_STANDARD = 0x00
val DOM_KEY_LOCATION_LEFT     = 0x01
val DOM_KEY_LOCATION_RIGHT    = 0x02
val DOM_KEY_LOCATION_NUMPAD   = 0x03
```

### DIFF
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -11,7 +11,7 @@ import scala.scalajs.js.typedarray.TypedArrayBufferOps._
 
 import org.scalajs.dom
 import org.scalajs.dom.{FormData, html, raw}
-import org.scalajs.dom.raw.Blob
+import org.scalajs.dom.raw.{Blob, KeyboardEvent}
 
 /**
  * Used to extend out javascript *Collections to make them usable as normal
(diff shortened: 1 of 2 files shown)


tindzk:
Thanks, @sjrd. I rebased the pull request.

What do you think about moving the constants from `KeyboardEvent` to the global scope (`package.scala`)?

sjrd:
> What do you think about moving the constants from KeyboardEvent to the global scope (package.scala)?

Why would we do that?

tindzk:
Thanks for merging!

> Why would we do that?

The constants are also in the global scope in JavaScript.

sjrd:
That doesn't appear to be the case, at least in Firefox: trying `DOM_KEY_LOCATION_STANDARD` in the browser console yields
```
ReferenceError: DOM_KEY_LOCATION_STANDARD is not defined
```
whereas `KeyboardEvent.DOM_KEY_LOCATION_STANDARD` works.

tindzk:
Indeed. In that case, we can keep the object as-is.

PR #315, opened 2018-02-26
Return HTMLCollection in getElementsBy*()

tindzk:
At the moment, all `getElementsBy*` functions return a `NodeList`, but the items are guaranteed to be `Element`s. Therefore, to access an item as an `Element`, a cast is required.

Furthermore, I would propose to remove the function `getElementsByName()` which does not appear in the DOM interfaces and seems to have been superseded by `getElementsByTagName()`. If you agree, I will add a separate commit for it to this pull request. 

See also:
- https://dom.spec.whatwg.org/#document
- https://dom.spec.whatwg.org/#element

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -805,14 +805,34 @@ abstract class Element
   def getAttribute(name: String): String = js.native
 
   /**
+   * Returns a list of elements with the given tag name. The subtree underneath the
+   * specified element is searched, excluding the element itself. The returned list is
+   * live, meaning that it updates itself with the DOM tree automatically.
+   * Consequently, there is no need to call several times
+   * element.getElementsByTagName with the same element and arguments.
+   *
+   * MDN
+   */
+  def getElementsByTagName(name: String): HTMLCollection = js.native
+
+  /**
    * Returns a list of elements with the given tag name belonging to the given namespace.
    *
    * MDN
    */
   def getElementsByTagNameNS(namespaceURI: String,
-      localName: String): NodeList = js.native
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -804,15 +804,35 @@ abstract class Element
...
+  /**
+   * Returns a list of elements with the given tag name. The subtree underneath the
+   * specified element is searched, excluding the element itself. The returned list is
+   * live, meaning that it updates itself with the DOM tree automatically.
+   * Consequently, there is no need to call several times
+   * element.getElementsByTagName with the same element and arguments.
+   *
+   * MDN
+   */
+  def getElementsByTagName(name: String): HTMLCollection = js.native
```
sjrd: Why was this moved from its previous location? It creates a bigger diff than necessary.

tindzk: I moved it for more consistency with `Document` and the DOM specification, which both declare `getElementsByTagName` first and `getElementsByTagNameNS` afterwards. When I originally saw the file, I was under the impression that `getElementsByTagName` was missing and added it. This confusion could have been avoided if it was placed before `getElementsByTagNameNS`.

tindzk:
If you have `svg` nodes in mind, `getElementsByTagName("svg")` would return a node list of `Element`s (which is exactly what an `HTMLCollection` is). The [Typescript's bindings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/blissfuljs/index.d.ts#L534) are actually more expressive and return `NodeListOf<SVGSVGElement>`. In Scala 2.13, we could implement it the same way with literal types.

PR #317, opened 2018-03-11
Add PointerEvent and related event handlers.

Busti:
Pointer events are now available [across all major browsers](https://caniuse.com/#feat=pointer-events), since they provide [several advantages](https://developers.google.com/web/updates/2016/10/pointer-events) over traditional mouse and touch events, it would be handy to be able to use them in scalajs.


Busti:
It seems like everything is now correctly formatted. Unfortunately I am unable to check this since the sbt task does not fail on my own system.

sjrd:
Can't you run `sbt scalafmt` to automatically fix all the formatting?

Busti:
Sorry, I did not know scalafmt was a thing.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1443,6 +1443,107 @@ class MouseEvent extends UIEvent with ModifierKeyEvent {
...
+ *
+ * The events needed to handle generic pointer input are analogous to mouse events
+ * (mousedown/pointerdown, mousemove/pointermove, etc.).
+ * Consequently, pointer event types are intentionally similar to mouse event types.
+ * Additionally, a pointer event contains the usual properties present in mouse events
+ * (client coordinates, target element, button states, etc.)
+ * in addition to new properties for other forms of input: pressure, contact geometry, tilt, etc.
+ * In fact, the PointerEvent interface inherits all of the MouseEvent's properties thus facilitating
+ * migrating content from mouse events to pointer events.
+ */
```
sjrd: Could you confirm that you are the original author of this documentation text?

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1443,6 +1443,107 @@ class MouseEvent extends UIEvent with ModifierKeyEvent {
...
+@js.native
+@JSGlobal
+class PointerEvent extends MouseEvent {
```
sjrd: The constructor should take a `type: String`, and optionally a `init: PointerEventInit`, according to https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/PointerEvent.

`PointerEventInit` should be modeled as a non-native JS trait, with `= js.undefined` for all members.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1443,6 +1443,107 @@ class MouseEvent extends UIEvent with ModifierKeyEvent {
...
+
+  /**
+   * A unique identifier for the pointer causing the event.
+   *
+   * MDN
+   */
+  def pointerId: Int = js.native
```
sjrd: https://www.w3.org/TR/pointerevents/#pointerevent-interface says that `pointerId` is a `long` in the IDL, so it should be a `Double` in Scala.js, not an `Int` (and not a `Long` either).

src/main/scala/org/scalajs/dom/raw/lib.scala
Busti: It is the documentation found on MDN. I added credit below (that must have slipped through).  
MDN documents are licensed under the Creative Commons license found here: https://github.com/mdn/mdn/blob/master/LICENSE

src/main/scala/org/scalajs/dom/raw/lib.scala
Busti: I added the necessary Init trait and constructor parameters.  
Please note that this also applies to MouseEvent, which is currently not implemented.

src/main/scala/org/scalajs/dom/raw/lib.scala
Busti: I changed it accordingly.

Busti:
In the last 2 commits I also added the additional `tangentialPressure` and `twist` properties specified here: https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/PointerEvent

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -1443,6 +1443,213 @@ class MouseEvent extends UIEvent with ModifierKeyEvent {
...
+   */
+  def isPrimary: Boolean = js.native
+}
+
+trait PointerEventInit extends js.Object {
+
+  /**
+   * Sets value of MouseEvent.pointerId. Defaults to 0.
+   */
+  var pointerId: js.undefOr[Double] = js.undefined
```
sjrd: `undefOr` -> `UndefOr` to fix the compile error.

Otherwise good to go.

PR #323, opened 2018-03-18
PermissionState.onchange and PushPermissionDescriptor

cornerman:
Link to [specification](https://w3c.github.io/permissions/#push)

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
@@ -24,7 +24,7 @@ package object permissions {
 
   trait PermissionStatus extends dom.raw.EventTarget {
     val state: PermissionState
-    var onchange: js.Function1[PermissionState, _]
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
```diff
@@ -52,6 +52,20 @@ package object permissions {
     }
   }
 
+  @js.native
+  trait PushPermissionDescriptor extends PermissionDescriptor {
```
sjrd: This should be a non-native trait, like `PermissionDescriptor`.

sjrd:
Also, could you give a link to the specification of `PushPermissionDescriptor` in the PR description, please? Otherwise I cannot check the accuracy of the facade wrt. the spec.

src/main/scala/org/scalajs/dom/experimental/permissions/package.scala
cornerman: done!

PR #324, opened 2018-03-18
Add applicationServerKey to PushSubscriptionOptions

cornerman:
Link to [specification](https://w3c.github.io/push-api/#dom-pushsubscriptionoptions-applicationserverkey)

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
@@ -4,7 +4,7 @@ import org.scalajs.dom.experimental.serviceworkers.ExtendableEvent
 import org.scalajs.dom.raw.Blob
 
 import scala.scalajs.js
-import scala.scalajs.js.typedarray.ArrayBuffer
+import scala.scalajs.js.typedarray.{ArrayBuffer, Uint8Array}
 
 /**
  * The PushManager interface of the Push API provides a way to receive notifications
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
```diff
@@ -183,6 +186,12 @@ object PushSubscriptionOptions {
       .literal("userVisibleOnly" -> userVisibleOnly)
       .asInstanceOf[PushSubscriptionOptions]
   }
+
+  def apply(userVisibleOnly: Boolean, applicationServerKey: Uint8Array): PushSubscriptionOptions = {
+    js.Dynamic
+      .literal("userVisibleOnly" -> userVisibleOnly, "applicationServerKey" -> applicationServerKey)
+      .asInstanceOf[PushSubscriptionOptions]
+  }
```
sjrd: It would be better to turn `PushSubscriptionOptions` into a non-native JS trait with `= js.undefined`, and let users do
```scala
new PushSubscriptionOptions {
  userVisibleOnly = ...
  applicationServerKey = ...
}
```
The `apply` method-based pattern is old; we do not use it anymore.

cornerman: done!

src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
```diff
@@ -166,23 +166,12 @@ trait PushMessageData extends js.Object {
...
-   * Creates a new [[PushSubscriptionOptions]] object.  Default value for userVisibleOnly is false.
-   */
-  def apply(userVisibleOnly: Boolean = false): PushSubscriptionOptions = {
-    js.Dynamic
-      .literal("userVisibleOnly" -> userVisibleOnly)
-      .asInstanceOf[PushSubscriptionOptions]
-  }
+  var userVisibleOnly: js.UndefOr[Boolean] = js.undefined
+  var applicationServerKey: js.UndefOr[Uint8Array] = js.undefined
 }
```
sjrd: Ah, good, but we'll need to keep the existing `PushSubscriptionOptions.apply` for backward binary compatibility. You can deprecate it, though.

cornerman: Oh, that makes sense

sjrd:
It wants some `sbt scalafmt` :)

cornerman:
I need to start remembering this step :)

PR #325, opened 2018-03-18
Add showNotification/getNotifications to ServiceWorkerRegistration

cornerman:
Specification:
- [showNotification](https://notifications.spec.whatwg.org/#dom-serviceworkerregistration-shownotification)
- [getNotifications](https://notifications.spec.whatwg.org/#dom-serviceworkerregistration-getnotifications)

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
@@ -4,7 +4,9 @@ import scala.scalajs.js
 import scala.scalajs.js.annotation._
 import scala.scalajs.js.|
 
-import org.scalajs.dom.experimental.{Request, RequestInfo, Response, Sequence}
+import org.scalajs.dom.experimental.{
+  Notification, NotificationOptions, Request, RequestInfo, Response, Sequence
+}
 import org.scalajs.dom.raw.{WorkerGlobalScope, ErrorEvent}
 import org.scalajs.dom.webgl.RenderingContext
 import org.scalajs.dom.{Event, EventTarget, MessagePort}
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -246,6 +246,42 @@ trait ServiceWorkerRegistration extends EventTarget {
...
+   */
+  def getNotifications(options: GetNotificationOptions = ???): js.Promise[js.Array[Notification]] = js.native
+
+  /**
+   * The showNotification() method of the ServiceWorkerRegistration interface
+   * creates a notification on an active service worker.
+   *
+   * MDN
+   */
+  def showNotification(title: String, options: NotificationOptions = ???): js.Promise[Event] = js.native
```
sjrd: The result type should be `js.Promise[Unit]`, not `js.Promise[Event]`. The spec says that the promise is completed with `undefined`, whose type in Scala.js is `Unit`.

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
```diff
@@ -246,6 +246,42 @@ trait ServiceWorkerRegistration extends EventTarget {
...
+  * returns a list of the notifications in the order that they were created
+  * from the current origin via the current service worker registration.
+  * Origins can have many active but differently-scoped service worker
+  * registrations. Notifications created by one service worker on the same
+  * origin will not be available to other active services workers on that same
+  * origin.
+   *
+   * MDN
+   */
+  def getNotifications(options: GetNotificationOptions = ???): js.Promise[js.Array[Notification]] = js.native
```
sjrd: You should use `Sequence` instead of `js.Array` here. (`Sequence` is defined [here](https://github.com/scala-js/scala-js-dom/blob/2bbe005f563adbdf8e5c089f7443f96cca8b8800/src/main/scala/org/scalajs/dom/experimental/package.scala#L42))

src/main/scala/org/scalajs/dom/experimental/serviceworkers/ServiceWorkers.scala
cornerman: oops, sorry for the oversight

PR #326, opened 2018-04-19
adding html.Code element and type

bbarker:


sjrd:
Travis didn't like it.

bbarker:
@sjrd Ah, I needed to run scalacheck.

sjrd:
With this PR, attempting to match on an `html.Code` will simply throw a `ReferenceError`, because there is no global variable called `HTMLCodeElement` in the browser.

I'm going to close this PR, then, as it is non-conforming to the Web specs.

bbarker:
Ah, that wouldn't be good.  So what would be a good way to model elements like <code> that are part of the HTML standard, but don't have any semantic difference from a bare element other than the tag name?

Maybe model them as opaque types if they are implemented, instead of `type Code = raw.HTMLElement` (which would be fairly useless). Beyond that, some about of reflection would have to be used under the hood to determine which opaque time to emit, if an HTMLElement is being matched on, I guess =/

PR #328, opened 2018-05-13
Add types for AbortController

guymers:
Add AbortController and AbortSignal classes along with adding `signal`
to the Fetch request classes. In addition also add `keepalive` to Fetch
request and fix the names of the `cache` and `redirect` properties.

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/Fetch.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/Fetch.scala
@@ -68,6 +68,10 @@ class Request(input: RequestInfo, init: RequestInit = null) extends js.Object {
   def redirect: RequestRedirect = js.native
 
   def integrity: String = js.native //should be DOMString
+
+  val keepalive: Boolean = js.native
+
+  val signal: js.UndefOr[AbortSignal] = js.native
 }
 
 object RequestInit {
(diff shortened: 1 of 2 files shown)


aappddeevv:
Can we get this merged? I'd like to abort some fetch requests and have that in this library vs my own code.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -95,10 +99,12 @@ object RequestInit {
...
       mode: js.UndefOr[RequestMode] = js.undefined,
       credentials: js.UndefOr[RequestCredentials] = js.undefined,
-      requestCache: js.UndefOr[RequestCache] = js.undefined,
-      requestRedirect: js.UndefOr[RequestRedirect] = js.undefined,
+      cache: js.UndefOr[RequestCache] = js.undefined,
+      redirect: js.UndefOr[RequestRedirect] = js.undefined,
       integrity: js.UndefOr[String] = js.undefined, //see [[https://w3c
       // .github.io/webappsec-subresource-integrity/ integrity spec]]
+      keepalive: js.UndefOr[Boolean] = js.undefined,
+      signal: js.UndefOr[AbortSignal] = js.undefined,
```
sjrd: These two lines will break the binary API of `apply`, so that's not good.

We should deprecate the `def apply` and encourage people to use `new RequestInit { ... }` instead.

aappddeevv:
I'm reading https://fetch.spec.whatwg.org/#body and don't see where keepalive can change. How did you interpret it that way? 

I'm all for deprecating the binary API of apply.

sjrd:
Hum indeed it seems `keepalive` can't change.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -68,6 +68,10 @@ class Request(input: RequestInfo, init: RequestInit = null) extends js.Object {
   def redirect: RequestRedirect = js.native
 
   def integrity: String = js.native //should be DOMString
+
+  val keepalive: Boolean = js.native
```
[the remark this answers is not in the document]
guymers: Done, will make a `def` just in case

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -68,6 +68,10 @@ class Request(input: RequestInfo, init: RequestInit = null) extends js.Object {
...
+
+  val signal: js.UndefOr[AbortSignal] = js.native
```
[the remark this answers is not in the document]
guymers: Done

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -88,6 +92,7 @@ object RequestInit {
    * //todo: it would help a lot if there were a way to make this fully type safe
    */
   @inline
+  @deprecated("0.9.6", "use `new RequestInit {}` instead")
```
sjrd: The arguments go in the other order. First the message, then the `since` version number. Also it should be be `0.9.7` since that's the next version that will be published.

guymers: Whoops, fixed

PR #337, opened 2018-07-23
Fix #327: Add HTMLElement.dataset

### LINKED ISSUE #327: requesting HTMLElement.dataset
I think [HTMLElement.dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) is now supported by most [browsers](https://caniuse.com/#feat=dataset). Can you please add it as part of scala-js-dom facade. 

Stack Overflow: https://stackoverflow.com/questions/34758423/using-element-dataset-in-scalajs

bpholt:
fixes #327


fdietze:
What prevents this PR from being merged?

PR #341, opened 2018-11-26
Fix the result type of a few methods in SVGTextContentElement to Double.

georgeorfanidi:
``def getComputedTextLength(): Int = js.native``
should be 
def getComputedTextLength(): Double = js.native

https://developer.mozilla.org/en-US/docs/Web/API/SVGTextContentElement


georgeorfanidi:
I did sign the CLA.

PR #342, opened 2018-11-29
Add `navigator.mediaDevices.getUserMedia()`, deprecate old method

rtyley:
The `navigator.getUserMedia()` method is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getUserMedia), and in fact during [our hackday](https://twitter.com/rtyley/status/1067815067780087808), I found that I could gain Firefox support just by using the `navigator.mediaDevices.getUserMedia()` method instead, which is [documented](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#Browser_compatibility
) as having good cross-browser support. The deprecation seems to have occurred around [28th May 2014](https://github.com/w3c/mediacapture-main/commit/de3bdaf83d2cc1350cc3be970baf1d11219add1d#diff-ea76d38900f79cfae8f60e5f7cf16dd1R2755).

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/webrtc/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/webrtc/package.scala
@@ -12,6 +12,23 @@ package object webrtc {
   @js.native
   trait NavigatorMediaStream extends js.Object {
 
+    /**
+     * The MediaDevices getUserMedia() method prompts the user for permission to use a media input
+     * which produces a MediaStream with tracks containing the requested types of media. That stream
+     * can include, for example, a video track (produced by either a hardware or virtual video source
+     * such as a camera, video recording device, screen sharing service, and so forth), an audio track
+     * (similarly, produced by a physical or virtual audio source like a microphone, A/D converter, or
+     * the like), and possibly other track types.
+     *
+     * MDN
+     *
+     * @see [[https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia]]
+     */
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/experimental/webrtc/package.scala
```diff
@@ -12,6 +12,23 @@ package object webrtc {
...
+     * can include, for example, a video track (produced by either a hardware or virtual video source
+     * such as a camera, video recording device, screen sharing service, and so forth), an audio track
+     * (similarly, produced by a physical or virtual audio source like a microphone, A/D converter, or
+     * the like), and possibly other track types.
+     *
+     * MDN
+     *
+     * @see [[https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia]]
+     */
+    def mediaDevices: MediaDevices
```
sjrd: The Scaladoc does not correspond to this member. You took the one from `MediaDevices` but it should be the one from [`Navigator.mediaDevices`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaDevices).

sjrd: It would also be best to add `= js.native` to that method, so that it is not considered as abstract. This is what we typically do for interfaces that are actually implemented natively by the DOM.

rtyley:
Thank you - it's a pleasure using scalajs!

PR #343, opened 2019-02-05
Add overloads of {add,remove}EventListener with options objects

Busti:
This Pull Request adds an implementation for passing an options object to `addEventListener`.  
The three options in that object allow for more control about how an event-listener behaves and can be quite useful, especially when working with touchscreen devices.  

The documentation was provided by MDN, details about the standard can be read here:  
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Parameters  

The standard is supported by all *major* browsers, with the only exception being IE11.
https://caniuse.com/#feat=passive-event-listener

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -2693,8 +2693,46 @@ class Window
 }
 
 /**
- * EventTarget is a DOM interface implemented by objects that can receive DOM events
- * and have listeners for them.
+ * An options object that specifies characteristics about the event listener.
+ *
+ * MDN
+ */
+trait EventListenerOptions extends js.Object {
+
+  /**
+   * A Boolean indicating that events of this type
+   * will be dispatched to the registered listener
+   * before being dispatched to any EventTarget
+   * beneath it in the DOM tree.
+   *
+   * MDN
+   */
+  var capture: js.UndefOr[Boolean] = js.undefined
+
+  /**
+   * A Boolean indicating that the listener
+   * should be invoked at most once after being added.
+   * If true, the listener would be automatically removed when invoked.
+   */
+  var once: js.UndefOr[Boolean] = js.undefined
+
+  /**
+   * A Boolean which, if true, indicates
+   * that the function specified by listener
+   * will never call preventDefault().
+   * If a passive listener does call preventDefault(),
+   * the user agent will do nothing other
+   * than generate a console warning.
+   * See Improving scrolling performance with passive listeners to learn more.
+   *
+   * MDN
+   */
+  var passive: js.UndefOr[Boolean] = js.undefined
+}
+
+/**
+ * EventTarget is a DOM interface implemented by objects
+ * that can receive DOM events and have listeners for them.
  *
  * Element, document, and window are the most common event targets, but other
  * objects can be event targets too, for example XMLHttpRequest, AudioNode,
@@ -2730,6 +2768,34 @@ class EventTarget extends js.Object {
   def addEventListener[T <: Event](`type`: String,
       listener: js.Function1[T, _],
       useCapture: Boolean = js.native): Unit = js.native
+
+  /**
+   * Removes the event listener previously registered with
+   * EventTarget.addEventListener.
+   *
+   * This implementation accepts a settings object of type EventListenerOptions.
+   *
+   * MDN
+   */
+  def removeEventListener[T <: Event](`type`: String,
+      listener: js.Function1[T, _],
+      options: EventListenerOptions): Unit = js.native
+
+  /**
+   * The EventTarget.addEventListener() method
+   * registers the specified listener
+   * on the EventTarget it's called on.
+   * The event target may be an Element in a document,
+   * the Document itself, a Window, or any other object that supports events
+   * (such as XMLHttpRequest).
+   *
+   * This implementation accepts a settings object of type EventListenerOptions.
+   *
+   * MDN
+   */
+  def addEventListener[T <: Event](`type`: String,
+      listener: js.Function1[T, _],
+      options: EventListenerOptions): Unit = js.native
 
   /**
    * Dispatches an Event at the specified EventTarget, invoking the affected


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -2692,6 +2692,38 @@ class Window
   var lostpointercapture: js.Function1[PointerEvent, _] = js.native
 }
 
+/**
+ * An options object that specifies characteristics about the event listener.
+ *
+ * MDN
+ */
+@js.native
```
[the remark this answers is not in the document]
Busti: Sorry for that. It compiled on my system and I kinda fell asleep while waiting for Travis

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -2692,6 +2692,38 @@ class Window
...
+  var capture: js.UndefOr[Boolean] = js.undefined
+
+  /**
+   * A Boolean indicating that the listener should be invoked at most once after being added.
+   * If true, the listener would be automatically removed when invoked.
+   */
+  var once: js.UndefOr[Boolean] = js.undefined
+
+  /**
+   * A Boolean which, if true, indicates that the function specified by listener will never call preventDefault().
```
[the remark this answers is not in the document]
Busti: It should be better now. All the lines I touched now have < 80 characters.  
However, there are some lines within the file that still contain > 80 chars.  
But fixing that should be another PR.

Busti:
I do not know what causes the failure here.  
Running the command on my machine is successful:  
```
[IJ]> scalafmtTest
[info] Running org.scalafmt.cli.Cli --test
All files are formatted with scalafmt :)
[success] Total time: 0 s, completed Feb 8, 2019, 11:26:05 PM
```

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -2693,8 +2693,45 @@ class Window
...
 
 /**
- * EventTarget is a DOM interface implemented by objects that can receive DOM events
- * and have listeners for them.
+ * An options object that specifies characteristics about the event listener.
+ *
+ * MDN
+ */
+trait EventListenerOptions extends js.Object {
+  /**
```
sjrd: My `scalafmt` says that there should be a blank line above this line.

Perhaps you have a newer version of the scalafmt sbt plugin enabled globally on your local machine?

Busti: Thank you very much for looking into this.  
I will squash this down into the original commit once the CI build passes.

ramnivas:
After making PR #348, I realized that there is this PR already pending. However, PR #348 also adds removeEventListener version that takes options. So either this PR should be amended with that addition or #348 should be considered as a superset of this.

Busti:
@ramnivas Thanks a lot for the heads up. That would have been confusing if people tried to remove that event listener.  
I have applied the fixes to this PR as well, to make the decision a bit harder for @sjrd.  
Choose whichever you like more. They are mostly identical except for some slight differences in the documentation.

ramnivas:
I have closed #348, since this PR came first and #348 is almost identical, so this should take the precedence.

I think you should change the commit message to include "removeEventListener" as well.

ramnivas:
@Busti BTW, you can run `sbt scalafmt` to fix the formatting errors seen in CI.

Busti:
@ramnivas The problem is that scalafmt does not seem to work correctly for me.  
I have tried running the commands that fail in travis on 3 different machines and in 3 different operating systems now.  
I even created a clean debian installation in a VM and still...

Every time I run `scalafmtTest` it just says `All files are formatted with scalafmt :)`

ramnivas:
@Busti hmm... May be you want to jiggle your code, perhaps by re-formatting comments (they use too many lines, IMO, and could be more readable by merging lines, anyway; see my closed PR for example). Then maybe scalafmt will be awaken!

Busti:
@ramnivas I think the comments should be fine. They were too long before, that's why I have created more line-breaks. See: https://github.com/scala-js/scala-js-dom/pull/343#discussion_r255012554  

I have now tried running `scalafmtTest` on:
- My Personal Computer (Windows IntelliJ)
- My Personal Computer (WSL Debian)
- My Work Computer (Windows IntelliJ)
- My Work Computer (WSL Debian)
- My Laptop (Arch)
- The University Desktop (Debian)

I do not know what else to do here. I am using scalafmt in several other projects and it is working as expected.
From what it sounds like, `scalafmt` does work correctly on your systems @ramnivas & @sjrd .

If it doesn't bother you too much I would like to ask again, for you to check the formatting errors on one of your systems and tell me what's wrong.  
Thanks a lot in advance.

ramnivas:
For me it has been working just fine.

@Busti Probably you already did this, but did you try `sbt scalafmt` (i.e. not `sbt scalafmtTest`).

My suggestion above is to give some ill/differently-formatted code and let `sbt scalafmt` do what it could do. All else fails, copy docs from my PR (which passed all the checks). If even that fails, may be I can resurrect my PR in favor of this to unblock.

sjrd:
`scalafmt` gives the following diff on my computer:
```diff
diff --git a/src/main/scala/org/scalajs/dom/raw/lib.scala b/src/main/scala/org/scalajs/dom/raw/lib.scala
index 6fe8218..cf591ac 100644
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -2737,8 +2737,8 @@ object EventListenerOptions {
    */
   @inline
   def apply(capture: js.UndefOr[Boolean] = js.undefined,
-    once: js.UndefOr[Boolean] = js.undefined,
-    passive: js.UndefOr[Boolean] = js.undefined): EventListenerOptions = {
+      once: js.UndefOr[Boolean] = js.undefined,
+      passive: js.UndefOr[Boolean] = js.undefined): EventListenerOptions = {
     val result = js.Dynamic.literal()
 
     capture.foreach(result.capture = _)
@@ -2797,8 +2797,8 @@ class EventTarget extends js.Object {
    * MDN
    */
   def removeEventListener[T <: Event](`type`: String,
-    listener: js.Function1[T, _],
-    options: EventListenerOptions): Unit = js.native
+      listener: js.Function1[T, _],
+      options: EventListenerOptions): Unit = js.native
 
   /**
    * The EventTarget.addEventListener() method
```

Busti:
Thank you @sjrd again for using scalafmt for me.  
I am sorry for the hold-up and any inconveniences I might have caused.  

I have failed to mention that I have previously used scala, sbt and scalafmt on all of the above systems.
It is about time that I install a fresh OS on my Laptop anyways. So maybe then scalafmt will work correctly.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -2693,8 +2693,65 @@ class Window
...
+   * the user agent will do nothing other
+   * than generate a console warning.
+   * See Improving scrolling performance with passive listeners to learn more.
+   *
+   * MDN
+   */
+  var passive: js.UndefOr[Boolean] = js.undefined
+}
+
+object EventListenerOptions {
```
sjrd: Hum, I only realize now that you added `object EventListenerOptions` since the last time I carefully reviewed.

I recommend not to define that object nor its `apply` method. The recommended way to create an `EventListenerOptions` should be to do something like:
```scala
elem.addEventListener("foo", { e => ... }, new EventListenerOptions {
  capture = true
})
```
`apply` methods in companions are a remnant from before we could do that, and we shouldn't define them for new code.

ramnivas: @sjrd Is there is an advantage of the suggested method besides fewer lines of code to not have the object (less generated code?). If so, we should perhaps deprecate the `apply` methods in objects such as `ClipboardEventInit`.

sjrd: * Less generated code, more efficient
* Resilient to evolution of the library: we can add fields without breaking binary compatibility

Busti: @sjrd I only did not add the `object` in my initial commit because I forgot about it.  

I have to admit that I am a bit torn on this issue.  
On one hand I have to agree with you that these apply methods should be deprecated, but on the other hand I would like to avoid confusion by mixing old and new implementations.  
These `apply` methods should definitely be deprecated, but I would prefer that to be a separate PR / Issue, which deprecates all of them at once.

sjrd: Sure, we should have a separate PR to deprecate all those `apply` methods. In the meantime, I suggest we don't *add* new ones in this PR. There's no point in adding stuff that we already know we're going to deprecate.

Busti: Okay. I will remove the `apply` method and rebase the branch later this evening.

ramnivas: FYI: https://contributors.scala-lang.org/t/expunging-new-from-scala-3/2868/93

Busti:
Sorry that this is so late. I had a rather eventful weekend and forgot about it.

PR #345, opened 2019-02-14
Add Element.matches which has all modern browser support

er1c:
Alternatively, this could be implemented as a polyfill for older browser support, but I don't think scala-js does anything with a polyfill, correct?

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -818,6 +818,15 @@ abstract class Element
   def insertAdjacentHTML(where: String, html: String): Unit = js.native
 
   /**
+   * The `matches()` method of the `Element` interface returns `true` if the
+   * element would be selected by the specified selector string; otherwise, it
+   * returns `false`.
+   *
+   * MDN
+   */
+  def matches(selector: String): Boolean = js.native
+
+  /**
    * getAttribute() returns the value of the named attribute on the specified element.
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -817,6 +817,13 @@ abstract class Element
    */
   def insertAdjacentHTML(where: String, html: String): Unit = js.native
 
+  /**
+   * Supported by FF&gt;34, Opera&gt;21, Chrome&gt;34, IE&gt;9, Safari&gt;4
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Element/matches
```
sjrd: I see that some comments above have this form, but they are the exception, not the norm. The appropriate Scaladoc is to copy-paste the description at the top of the MDN page, and add `MDN` at the end:
```scala
  /**
   * The `matches()` method of the `Element` interface returns `true` if the
   * element would be selected by the specified selector string; otherwise, it
   * returns `false`.
   *
   * MDN
   */
```

sjrd:
> Alternatively, this could be implemented as a polyfill for older browser support, but I don't think scala-js does anything with a polyfill, correct?

Indeed, we don't do polyfills.

src/main/scala/org/scalajs/dom/raw/lib.scala
er1c: 👍  PR updated

PR #347, opened 2019-02-14
Fix #346: Give the correct type to PushSubscription.toJSON

### LINKED ISSUE #346: PushSubscription.toJSON was the wrong return type
The method toJSON on PushSubscription return a Object not a String

https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription/toJSON

https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala#L101

The error:
```
scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: [object Object] is not an instance of java.lang.String
...
Caused by: java.lang.ClassCastException: [object Object] is not an instance of java.lang.String
```

FabioPinheiro:
Update PushSubscription to better reflect w3 standards:
-Change the return type of method toJSON
-Add missing properties (expirationTime, options)

Disclosure I have no idea what I'm doing here or if this is event a problem for anyone else.
(The impression that I have is that if any one is calling this method this will throw a cast extension)

This PR fix my problem but I don't know if this `js.Dictionary[js.Any] ` is the right type, and maybe could be better refine. (https://www.w3.org/TR/2018/WD-push-api-20181026/#idl-def-pushsubscriptionoptionsinit)

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
@@ -14,7 +14,7 @@ import scala.scalajs.js.typedarray.{ArrayBuffer, Uint8Array}
  *
  * MDN
  *
- * The Push API is currently specified here: [[http://www.w3.org/TR/2015/WD-push-api-20151215/]]
+ * The Push API is currently specified here: [[https://www.w3.org/TR/2018/WD-push-api-20181026/]]
  */
 @js.native
 trait PushManager extends js.Object {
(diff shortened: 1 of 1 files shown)


FabioPinheiro:
Fix proposal for https://github.com/scala-js/scala-js-dom/issues/346

src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
```diff
@@ -79,6 +79,23 @@ trait PushSubscription extends js.Object {
...
   val endpoint: String = js.native
 
+  /**
+   * The expirationTime read-only property of the PushSubscription interface returns a
+   * DOMHighResTimeStamp of the subscription expiration time associated with the push
+   * subscription, if there is one, or null otherwise.
+   *
+   * MDN
+   */
+  val expirationTime: Double = js.native
```
sjrd: This should be a `java.lang.Double` so that `null` is an acceptable value.

src/main/scala/org/scalajs/dom/experimental/push/PushManager.scala
```diff
@@ -98,7 +115,7 @@ trait PushSubscription extends js.Object {
    *
    * MDN
    */
-  def toJSON(): String = js.native
+  def toJSON(): js.Dictionary[js.Any] = js.native
```
sjrd: You should use a new type `PushSubscriptionJSON` as specified at https://www.w3.org/TR/2018/WD-push-api-20181026/#pushsubscription-interface. Something like:
```scala
trait PushSubscriptionJSON extends js.Object {
  val endpoint: String
  val expirationTime: java.lang.Double
  val keys: js.Dictionary[String]
}
```

PR #348, opened 2019-02-20
Add a version of add/removeEventLister to take options

ramnivas:
MDN documentaiton reference:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener


ramnivas:
Closing in favor of #343

PR #349, opened 2019-03-10
fix(idb): Transaction mode constants are no longer supported

mprevel:
Hi,

IDB transaction mode constants have been replaced by raw String values since FF25 and Chrome 21.

This PR :
Removes theses constants from native API facade.
Adds an helper in ext package to keep these constants instead of providing raw String to the function.

These constants were not working anymore (undefined behavior exception at runtime when used).
So,  removing them should not be an issue.

More info on https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction#Mode_constants

Regards,


mprevel:
I'll open another PR instead of this one.

PR #350, opened 2019-03-18
feat(idb): Update to newer API

mprevel:
IDB DomString constants have been removed from browser API and replaced by raw String values.

Remove IDB DOMString constants that are no more supported and lead to an UndefinedBehaviorError.
Add these constants in a typesafe way.

### DIFF
--- a/src/main/scala/org/scalajs/dom/idb.scala
+++ b/src/main/scala/org/scalajs/dom/idb.scala
@@ -5,7 +5,7 @@ package org.scalajs.dom
  */
 object idb {
   type Cursor = IDBCursor
-  @inline def Cursor = IDBCursor
+  @inline def CursorDirection = raw.IDBCursorDirection
   type CursorWithValue = IDBCursorWithValue
   type Database = IDBDatabase
   type Factory = IDBFactory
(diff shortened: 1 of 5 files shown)


armanbilge (commented):
At a quick glance, this one looks like it may not be binary compatible. Aside: we should add MiMa to the build if possible.

src/main/scala/org/scalajs/dom/idb.scala
```diff
@@ -5,7 +5,7 @@ package org.scalajs.dom
  */
 object idb {
   type Cursor = raw.IDBCursor
-  @inline def Cursor = raw.IDBCursor
+  @inline def CursorDirection = raw.IDBCursorDirection
```
japgolly, line 8: Hey @sjrd is this binary-compatible? Not sure if this counts as a "member" (of a pkg object) like you were saying in #461 , or if not if it's safe anyone because of `@inline`?

sjrd: It's *not* binary compatible (irrespective of `@inline`), since it is in a Scala object (not a JS facade). So this PR cannot be merged as is for 1.x.

src/main/scala/org/scalajs/dom/raw/Idb.scala
```diff
@@ -334,13 +416,45 @@ class IDBCursor extends js.Object {
 }
 
 @js.native
-@JSGlobal
-object IDBCursor extends js.Object {
```
sjrd, line 338: Removing a top-level object/class/trait is not binary compatible either, even if it's a facade.

sjrd (commented):
In 1.x, you can introduce the new defs and top-level members, but you cannot remove the old ones. You have to deprecate them instead.

japgolly:
@armanbilge I'm thinking to just reschedule this to 2.0 rather than try to massage compatibility. wdyt?

armanbilge:
Yup, agree!

japgolly:
rebased

japgolly (approved):

japgolly:
Thanks so much for the PR @mprevel and sorry for the big turnaround time in getting this merged.

mprevel:
I was a bit busy and did not follow what happened to this PR lately.
Thanks for your work on it and for having it merged :)

PR #360, opened 2019-05-13
Mark deprecated/non-standard HTMLElement interfaces as deprecated

### LINKED ISSUE #272: Remove/deprecate HTMLBlockElement & HTMLPhraseElement
HTMLBlockElement and HTMLPhraseElement don't appear in any of the specs and I can't conjure either of them in Firefox. Searching the web suggests they're IE-specific implementation details that may or may not still exist in recent versions of IE.

exoego:
Closes #272

### DIFF
--- a/src/main/scala/org/scalajs/dom/html.scala
+++ b/src/main/scala/org/scalajs/dom/html.scala
@@ -9,12 +9,14 @@ object html {
   type Applet = raw.HTMLAppletElement
   type Audio = raw.HTMLAudioElement
   type Area = raw.HTMLAreaElement
+  @deprecated("Non-standard", "0.9.8")
   type AreasCollection = raw.HTMLAreasCollection
   type Base = raw.HTMLBaseElement
   @deprecated("Obsolete.", "DOM Level 2")
   type BaseFont = raw.HTMLBaseFontElement
   @deprecated("Non standard.", "forever")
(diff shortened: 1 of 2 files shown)


sjrd:
I'd approve, but the CI didn't pass.

exoego:
Fixed the missing `@deprecated` on type alias and `def`.

src/main/scala/org/scalajs/dom/html.scala
```diff
@@ -9,24 +9,28 @@ object html {
...
+  @deprecated("Non-standard", "0.9.8")
   type BlockElement = raw.HTMLBlockElement
   type Body = raw.HTMLBodyElement
   type Button = raw.HTMLButtonElement
   type BR = raw.HTMLBRElement
   type Canvas = raw.HTMLCanvasElement
   type Collection = raw.HTMLCollection
   type DataList = raw.HTMLDataListElement
+  @deprecated("Non-standard.", "0.9.8")
   type DD = raw.HTMLDDElement
```
damercier: Hello, 
Why are the DD and DT considered non-standard? 

Actually, this is valid HTML5 elements. 
See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dd and https://www.w3.org/TR/html52/grouping-content.html#the-dt-element 

Can we re-introduce those missing elements in the next release?

PR #361, opened 2019-05-13
Add URL.searchParams (URLSearch​Params)

### LINKED ISSUE #339: add URLSearchParams api
This looks like it is in most standards. 

I'm happy to do a PR but I need some guidance where this would actually go in the scala-js-dom project.

Thoughts?

exoego:
Closes #339 

[URLSearch​Params](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) is [widely-adopted](https://caniuse.com/#feat=urlsearchparams) standard.

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/URL.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/URL.scala
@@ -92,4 +92,52 @@ class URL(url: String, base: String = js.native) extends js.Object {
    * MDN
    */
   var hash: String = js.native
+
+  var searchParams: URLSearchParams = js.native
+}
+
+/**
+ * The URLSearchParams defines utility methods to work with the query string of a URL.
+ *
+ * MDN
+ */
+@js.native
+@JSGlobal
+class URLSearchParams
+    extends js.Iterable[js.Tuple2[String, String]] {
+
+  /**
+   * Leading '?' characters are ignored.
+   *
+   * MDN
+   */
+  def this(init: String) = this()
+  def this(init: Sequence[String]) = this()
+  def this(init: js.Dictionary[String]) = this()
+
+  def append(name: String, value: String): Unit = js.native
+
+  def delete(name: String): Unit = js.native
+
+  def get(name: String): js.UndefOr[String] = js.native
+
+  def getAll(name: String): Sequence[String] = js.native
+
+  def has(name: String): Boolean = js.native
+
+  def set(name: String, value: String): Unit = js.native
+
+  def sort(): Unit = js.native
+
+  @JSName(js.Symbol.iterator)
+  override def jsIterator(): js.Iterator[js.Tuple2[String, String]] = js.native
+
+  def entries(): js.Iterator[js.Tuple2[String, String]] = js.native
+
+  def keys(): js.Iterator[String] = js.native
+
+  def values(): js.Iterator[String] = js.native
+
+  def forEach(callback: js.Function2[String, String, Unit]): Unit = js.native
+
 }
--- a/src/main/scala/org/scalajs/dom/ext/Extensions.scala
+++ b/src/main/scala/org/scalajs/dom/ext/Extensions.scala
@@ -276,7 +276,7 @@ object Ajax {
       } else {
         // fall back to copying the data
         val tempBuffer = ByteBuffer.allocateDirect(data.remaining)
-        val origPosition = data.position
+        val origPosition = data.position()
         tempBuffer.put(data)
         data.position(origPosition)
         tempBuffer.typedArray()


src/main/scala/org/scalajs/dom/experimental/URL.scala
```diff
@@ -92,4 +92,59 @@ class URL(url: String, base: String = js.native) extends js.Object {
...
+
+/**
+ * The URLSearchParams defines utility methods to work with the query string of a URL.
+ *
+ * An object implementing URLSearchParams can directly be used in a for...of structure, for example the following two lines are equivalent:
+ *
+ * MDN
+ */
+@js.native
+trait URLSearchParamsInterface extends js.Object {
```
sjrd: There doesn't seem to exist an `URLSearchParamsInterface` even in the IDL definitions. Consider just defining the `class URLSearchParams`.

src/main/scala/org/scalajs/dom/experimental/URL.scala
```diff
@@ -92,4 +92,59 @@ class URL(url: String, base: String = js.native) extends js.Object {
...
+  def entries(): js.Iterator[js.Tuple2[String, String]] = js.native
+
+  @JSName(js.Symbol.iterator)
+  def values(): js.Iterator[String] = js.native
+
+  def forEach(callback: js.Function2[String, String, Unit]): Unit = js.native
+}
+
+/**
+ * The URLSearchParams() constructor creates and returns a new URLSearchParams object. Leading '?' characters are ignored.
```
sjrd: Please wrap lines at 80 characters.

src/main/scala/org/scalajs/dom/experimental/URL.scala
```diff
@@ -92,4 +92,59 @@ class URL(url: String, base: String = js.native) extends js.Object {
...
+ *
+ * MDN
+ */
+@js.native
+@JSGlobal
+class URLSearchParams extends URLSearchParamsInterface {
```
sjrd: The class should extend `js.Iterable[js.Tuple2[String, String]]`.

src/main/scala/org/scalajs/dom/experimental/URL.scala
```diff
@@ -92,4 +92,59 @@ class URL(url: String, base: String = js.native) extends js.Object {
...
+  def getAll(name: String): js.Array[String]
+  def has(name: String): Boolean
+  def set(name: String, value: String): Unit
+  def sort(): Unit
+
+  @JSName(js.Symbol.iterator)
+  def entries(): js.Iterator[js.Tuple2[String, String]] = js.native
+
+  @JSName(js.Symbol.iterator)
+  def values(): js.Iterator[String] = js.native
```
sjrd: Those two methods with different signatures and the same `@JSName` make no sense. I think what you wanted is a single `def iterator(): js.Iterator[js.Tuple2[String, String]]` to conform to the interface of `js.Iterable[js.Tuple2[String, String]]`. The methods `entries()` and `values()` should not have any `@JSName`. It also seems that you're missing the `def keys()` method.

See https://heycam.github.io/webidl/#idl-iterable

src/main/scala/org/scalajs/dom/experimental/URL.scala
exoego: Fixed in https://github.com/scala-js/scala-js-dom/pull/361/commits/1db81ff3a6942fe334eca9f506ea0a789e8d83b3

It was separated as MDN says URLSearchParams has separate ["interface"](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and [constructor](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams).
But, same name between trait and class is not allowed in Scala, so it was suffixed with `~Interface`.
By separating trait, we can have our own class that implements URLSearchParams interface.

However, I change my mind to just define class, since `URL` is defined so.

src/main/scala/org/scalajs/dom/experimental/URL.scala
exoego: Removed this line during merging trait and class.

src/main/scala/org/scalajs/dom/experimental/URL.scala
exoego: Fixed

src/main/scala/org/scalajs/dom/experimental/URL.scala
exoego: Correct.

Implemented `js.Iterable[js.Tuple2[String, String]]` and used `@JSName` to `override def jsIterator`.
`keys` also added.

src/main/scala/org/scalajs/dom/ext/Extensions.scala
```diff
@@ -276,7 +276,7 @@ object Ajax {
       } else {
         // fall back to copying the data
         val tempBuffer = ByteBuffer.allocateDirect(data.remaining)
-        val origPosition = data.position
+        val origPosition = data.position()
```
sjrd: This change is unrelated and should not be in this commit/PR.

exoego: Ah, indeed.
Since it was fixed in another PR, I will rebase and exclude unrelated change.

exoego: Removed !!

PR #366, opened 2019-05-14
Remove invalid HTMLImageElement.create 

### LINKED ISSUE #274: HTMLImageElement's create() is probably wrong
for this `create() ` method here https://github.com/scala-js/scala-js-dom/blob/de25b454f82112406d75cc1a2e3e84db40f20290/src/main/scala/org/scalajs/dom/raw/Html.scala#L1445

there seems to be no corresponding method here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement

discussion on Gitter: https://gitter.im/scala-js/scala-js?at=58ce837c6d7eb18404e37c35

```
object ie extends HTMLImageElement

@JSExport("Main")
  object Main extends js.JSApp {
@JSExport
         def main(): Unit = {
 

  val i2=ie.create()

 println("ie "+ie)
  println("ie "+i2)
...
```

gives:
```
Main.scala:46Uncaught TypeError: Cannot read property 'create' of undefined(…)$c_Lapp_client_Main$.main__V @ Main.scala:46$c_Lapp_client_Main$.$$js$exported$meth$main__O @ Main.scala:45$c_Lapp_client_Main$.main @ Main.scala:45(anonymous function) @ (index):1
```

exoego:
Closes #274

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/Html.scala
+++ b/src/main/scala/org/scalajs/dom/raw/Html.scala
@@ -1546,8 +1546,6 @@ abstract class HTMLImageElement extends HTMLElement {
    */
   def complete: Boolean = js.native
 
-  def create(): HTMLImageElement = js.native
-
   var onload: js.Function1[Event, _] = js.native
 }
 


exoego:
Removed entirely in https://github.com/scala-js/scala-js-dom/pull/366/commits/7abdf751c393503c816860b804610e861f52cbee

PR #370, opened 2019-06-15
Graduate WebGL 1 from experimental

### LINKED ISSUE #352: Should WebGL move from `experimental`?
Since Microsoft has stopped supporting IE10 and below for several years now, and IE11 is feature frozen, I would argue WebGL 1.0 (modulo extensions) and similar APIs are now widely supported by the definition of "last 2 releases" of a browser, assuming we don't want to encourage catering for browsers that are inherently insecure and putting end users at risk. (As corrigendum, the phrasing wrt support should reflect this, since the initial statement assumed IE was at least going to maintain a release cycle.)

Should we consider promoting this, and other APIs into `org.scalajs.dom` and leave deprecation notices to import directly from `dom`? This applies to other APIs in IE11 but not IE10 that are in the same situation (and a slew of iOS Safari updates since iOS 8 since I last tested) but I will need to review what in particular is affected. At minimum it includes various CSS properties that have since become commonplace, I suspect it may include subsets of webrtc.

WebGL extensions are themselves fairly unique since one must probe for their existence a-priori calling them, and I feel given the above, they too should move into the main package, to be replaced by the newer WebGL 2.0 spec in `experimental`.

WDYT?

exoego:
Closes #352 by moving files and definining alias for backward-compat.
[WebGL 1 is widely supported by major browsers including IE11](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API#WebGL_1).

Some of other experimental packages, like [`domparser` (XMLSerializer)](https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer#Browser_compatibility), could be graduated, but in different PR.

Adding WebGL 2 as experimental is huge task, so different PR.


armanbilge:
This seems like a good candidate for v2.

exoego:
I am OK having this only on v2.

armanbilge:
@japgolly I'm leaving this as a PR targetting master.

japgolly:
Rather than dealing with this merge conflict, this can be re-done in #545.

Thanks for the PR @exoego, it's appreciated!

PR #375, opened 2019-08-12
Fix #374: Deprecate DOMError in favor of DOMException.

### LINKED ISSUE #374: Use DOMException instead of DOMError
The current method signature of `FileReader.error` is 
```scala
def error: DOMError = js.native
```
But from [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/error):
> A DOMError containing the relevant error. In Chrome 48+/Firefox 58+ this property returns a DOMException because DOMError has been removed from the DOM standard.

The `DOMError` MDN page also says this DOMError was deprecated.

Also the `DOMError` and the `DOMException` APIs in scala-js-dom is not complete. The MDN `DOMException class` has three methods, but scala-js-dom API only has two, the deprecated message for `DOMException.code` is not incorrect now.

floating-cat:
Fix https://github.com/scala-js/scala-js-dom/issues/374


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -6665,45 +6665,61 @@ abstract class CharacterData extends Node with NonDocumentTypeChildNode {
...
-   * Readonly short. Returns one of the exception code constants (see below).
-   * Deprecated use DOMError.name instead.
+   * Returns a short that contains one of the error code constants one of the
+   * exception code constants (see below) or 0 if none match. This field is used
+   * for historical reasons. New DOM exceptions don't use this anymore: they put
+   * this info in the name attribute.
    *
    * MDN
    */
-  @deprecated("Obsolete.", "DOM4")
```
sjrd: The deprecation should be kept, here. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#Properties) still says it's deprecated.

src/main/scala/org/scalajs/dom/experimental/webrtc/WebRTC.scala
```diff
@@ -1060,6 +1060,7 @@ class RTCPeerConnection(
    * gathered. The answer is delivered to the returned Promise, and should then be sent to
    * the source of the offer to continue the negotiation process.
    *
+   * MDN
```
sjrd: Please include this in the proper commit, rather than fixing the previous commit with a change in this commit. `git rebase --interactive` can help you.

floating-cat:
I have addressed your mentioned problems and rebased code.

floating-cat:
Thanks. I have fixed the typo!

PR #378, opened 2019-09-09
Add experimental Storage API support.

### LINKED ISSUE #377: Feat: Storage API for persistent storage and quota estimates
https://storage.spec.whatwg.org/

Chrome and Firefox have implemented
https://developer.mozilla.org/en-US/docs/Web/API/Storage_API#Browser_compatibility

exoego:
Closes #377 

https://storage.spec.whatwg.org/

Chrome and Firefox have implemented
https://developer.mozilla.org/en-US/docs/Web/API/Storage_API#Browser_compatibility

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/storage/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/storage/package.scala
@@ -1,0 +1,32 @@
+package org.scalajs.dom.experimental
+
+import org.scalajs.dom
+
+import scala.language.implicitConversions
+import scala.scalajs.js
+
+/**
+ * https://storage.spec.whatwg.org/
+ */
+package object storage {
+  implicit def toNavigatorStorage(navigator: dom.Navigator): NavigatorStorage =
+    navigator.asInstanceOf[NavigatorStorage]
+
+  @js.native
+  trait NavigatorStorage extends js.Object {
+    val storage: StorageManager = js.native
+  }
+
+  @js.native
+  trait StorageManager extends js.Any {
+    def persisted(): js.Promise[Boolean] = js.native
+    def persist(): js.Promise[Boolean] = js.native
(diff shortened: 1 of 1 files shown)


exoego:
Is this project no longer maintained ?
If so, I will fork and publish.
Thanks for years of maintenance.

src/main/scala/org/scalajs/dom/experimental/storage/package.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * https://storage.spec.whatwg.org/
+ */
+package object storage {
+  implicit def toNavigatorStorage(navigator: dom.Navigator): NavigatorStorage =
+    navigator.asInstanceOf[NavigatorStorage]
+
+  @js.native
+  trait NavigatorStorage extends js.Object {
+    val storage: js.UndefOr[StorageManager] = js.native
```
sjrd: Why is this a `js.UndefOr`? IIUC the spec, this should always be define, shouldn't it?

sjrd:
> Is this project no longer maintained ?

It's maintained but it's not on the shortlist of projects that I continuously monitor. So if I don't find the time shortly after receiving the notification for some PR/comment/etc., things can get lost. :(

src/main/scala/org/scalajs/dom/experimental/storage/package.scala
exoego: I initially thought that `js.UndefOr` allows user to detect the feature exists on browsers.
However, such feature detection can be achieved in other way.
So now removed `js.UndefOr` in https://github.com/scala-js/scala-js-dom/pull/378/commits/77276de4584bdbf452d6e7302b38ee0c94b728e3 as same as in other experimental API support (e.g. `toWebRTC`).

PR #379, opened 2019-09-09
Fix misuse of ErrorEvent

unarist:
* IndexedDB uses `Event` for `error` events
https://www.w3.org/TR/IndexedDB/#fire-error-event
  * There is no `message` property in the event. You need to obtain details from `IDBRequest.error` property instead, for example.
https://developer.mozilla.org/en-US/docs/Web/API/IDBRequest/onerror
* XMLHttpRequest uses `ProgressEvent` for events except `readystatechange`
https://xhr.spec.whatwg.org/#events


armanbilge:
Another one I'm not sure about bincompat, but definitely good for v2.

japgolly:
Agree @armanbilge . Setting to 2.0.0...

armanbilge:
Actually on second look there's no bincompat concerns here (all the changes are to an erased type parameter).

japgolly:
Ok LGTM as it, merging.

Thanks @unarist for this PR and your patience. This project now has new maintainers, myself and @armanbilge , so we're gonna help out to be responsive to PRs like this going forward.

PR #380, opened 2019-09-09
Remove ServiceWorkerContainer.onerror

unarist:
It looks like not documented in the spec nor implemented in browsers.
see https://github.com/w3c/ServiceWorker/issues/912

The current spec has `messageerror` event instead, but I'm omitting it because it is not described at MDN nor implemented Chrome for now. Tell me if I should include it in this PR.


armanbilge:
@unarist apologies that we are just getting to this, thank you for your contribution!! Have there been any relevant updates to the spec since your PR?

unarist:
> Have there been any relevant updates to the spec since your PR?

The spec seems to not changed about this.

MDN docs had been updated to deprecate [ServiceWorkerContainer.onerror](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/onerror). It also describes [Worker.onmessageerror](https://developer.mozilla.org/en-US/docs/Web/API/Worker/onmessageerror), but Chrome is still not supporting that event.

armanbilge:
@unarist thanks for getting back on this! `onerror` seems long gone, let's proceed with your PR for 1.2.0.

armanbilge (approved):

api-reports/2_12.txt
```diff
@@ -1028,7 +1028,6 @@ experimental/serviceworkers/ServiceWorkerContainer[ST] def controller: ServiceWo
 experimental/serviceworkers/ServiceWorkerContainer[ST] def getRegistration(scope: String = ""): js.Promise[js.UndefOr[ServiceWorkerRegistration]]
 experimental/serviceworkers/ServiceWorkerContainer[ST] def getRegistrations(): js.Promise[js.Array[ServiceWorkerRegistration]]
 experimental/serviceworkers/ServiceWorkerContainer[ST] var oncontrollerchange: js.Function1[Event, _]
-experimental/serviceworkers/ServiceWorkerContainer[ST] var onerror: js.Function1[ErrorEvent, _]
```
armanbilge, line 1031: @japgolly dumb question, why is `ServiceWorkerContainer` annotated as an `ST` scala trait?

japgolly: Great catch! It's a bug, raised: https://github.com/scala-js/scala-js-dom/issues/500

japgolly (approved):

japgolly:
Ok @unarist & @armanbilge , in order to preserve backwards-binary-compatibility, I've restored the field but marked it as deprecated. We'll remove all the deprecated stuff on the 2.x branch

armanbilge:
Wait, why is this binary incompatible?

armanbilge:
This is a facade so it falls under the exceptions: https://github.com/scala-js/scala-js-dom/pull/461#issuecomment-894829615

japgolly:
ARGH! Thanks @armanbilge , I forgot this case is an SJS bincomat exception. D'oh. I'll revert...

armanbilge:
I wonder if we could use your scalafix to generate list of MiMa exceptions, so we can just run MiMa. One day!

japgolly:
Actually yeah, that's a really good idea. I should be able to implement that reasonably quickly. I'll see if I can smash it out today...

armanbilge:
@japgolly You rock!! Your scalafix wizardry has already opened my eyes 🤩 Plus I think this could be super-helpful for a future auto-generation scheme.

japgolly:
Boom. Merged. Thanks for the PR (and your incredible patience) @unarist , and thanks for you attentive reviewing. :)

armanbilge:
That was our last PR! How do you feel about cutting a release @japgolly? The remaining issues look like infrastructure/internal stuff mostly: https://github.com/scala-js/scala-js-dom/milestone/1

PR #381, opened 2019-09-17
Make the fields of RequestInit optional.

zakpatterson:
This way, users do not have to recite every field when creating a `RequestInit`. Also, the spec says the `RequestInit` dictionary must contain a field `window` which is set to null. We might as well include this too as a final variable, since the spec shows omitting it should be a type error.

Prompted by gitter conversation here: https://gitter.im/scala-js/scala-js?at=5d80415105fd3716953afa47

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/Fetch.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/Fetch.scala
@@ -133,30 +133,31 @@ object RequestInit {
  * The underlying object is a dictionary. This trait is here to help encode
  * the types.
  */
+@js.native
 trait RequestInit extends js.Object {
   var method: js.UndefOr[HttpMethod]
 
-  var headers: js.UndefOr[HeadersInit]
+  var headers: js.UndefOr[HeadersInit] = js.undefined
 
   var body: js.UndefOr[BodyInit]
 
-  var referrer: js.UndefOr[String]
+  var referrer: js.UndefOr[String] = js.undefined
 
-  var referrerPolicy: js.UndefOr[ReferrerPolicy]
+  var referrerPolicy: js.UndefOr[ReferrerPolicy] = js.undefined
 
-  var mode: js.UndefOr[RequestMode]
+  var mode: js.UndefOr[RequestMode] = js.undefined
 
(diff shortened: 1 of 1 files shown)


zakpatterson:
I see, however that `@js.native` traits are not supposed to have default definitions: https://www.scala-js.org/doc/interoperability/facade-types.html.  Oddly this works with scala.js 1.0 and 0.6 on scala 2.12 and 2.13, but fails on 2.11 and 2[.yikes].10.

flavomulti:
I think if you pull off the js.native this will work as expected though. Are you going to resubmit?

zakpatterson:
@flavomulti I just removed the js.native as you suggested. I remember now that I added that so that the window field could be set to null explicitly. Perhaps the way I have it now will work?

flavomulti:
I think that as long as its not set, e.g. js.undefined, its ok not to specify it. If you do specify window, it must be null...something like that. It's never clear why funky things like this exist, but it is what it is.

zakpatterson:
Ah, that makes sense. So this will still throw a typeerror if a user tries to set it to something that's not null, which is what the spec wants. Ok thanks @flavomulti !

aappddeevv:
Is there anyway to get this merged? With 1.0 out, we really need js.undefined on the RequestInit trait's properties as it is very unergonomic to use otherwise.

src/main/scala/org/scalajs/dom/experimental/Fetch.scala
```diff
@@ -136,27 +136,27 @@ object RequestInit {
 trait RequestInit extends js.Object {
   var method: js.UndefOr[HttpMethod]
```
sjrd, line 137: `method` should also have `= js.undefined`. It seems optional, according to step 27 of the constructor of `Request` ([spec](https://fetch.spec.whatwg.org/#request-class)).

zakpatterson: 55c1d66

sjrd:
Thank you @zakpatterson

aappddeevv:
Thanks!

PR #383, opened 2019-10-07
Make Blob compliance with documentation

FabioPinheiro:
https://developer.mozilla.org/en-US/docs/Web/API/Blob
Add missing methods on Blob
Update the documentation 
Remove the close method


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -7997,15 +8001,44 @@ class Blob(blobParts: js.Array[js.Any] = js.native,
...
+  /**
+   * A string indicating the MIME type of the data contained in the Blob. If the type
+   * is unknown, this string is empty.
    *
    * MDN
    */
   def slice(start: Double = js.native, end: Double = js.native,
       contentType: String = js.native): Blob = js.native
 
-  def close(): Unit = js.native
```
FabioPinheiro: Where did this method come from?
I don't see it in the documentation and give a error on the browser if a call on a Blob (due to be undefined)

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -7997,15 +8001,44 @@ class Blob(blobParts: js.Array[js.Any] = js.native,
...
+  /**
+   * Returns a ReadableStream that can be used to read the contents of the blob.
+   *
+   * MDN
+   */
+  def stream(): ReadableStream[Byte] = js.native
+
+  /**
+   * Returns a promise that resolves with a USVString containing the entire
```
FabioPinheiro: Shouldn't we have some place where type USVString is defined?
ex: `type USVString = String`

src/main/scala/org/scalajs/dom/raw/lib.scala
FabioPinheiro: I don't know how to check the previous versions of [MDN-Blob-Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob)

Should we just annotate with @deprecated

FabioPinheiro:
@sjrd Can I have feedback on the PR?
Would be possible to fix this for the next released

src/main/scala/org/scalajs/dom/raw/lib.scala
japgolly: Ideally yes, but let's leave that for another time and get this merged for now :)

api-reports/2_12.txt
```diff
@@ -2534,9 +2534,12 @@ raw/BiquadFilterNode[JT] val numberOfOutputs: Int
 raw/BiquadFilterNode[JT] def removeEventListener[T <: Event](`type`: String, listener: js.Function1[T, _], options: EventListenerOptions): Unit
 raw/BiquadFilterNode[JT] def removeEventListener[T <: Event](`type`: String, listener: js.Function1[T, _], useCapture: Boolean?): Unit
 raw/BiquadFilterNode[JT] var `type`: String
-raw/Blob[JC] def close(): Unit
```
japgolly, line 2537: Not actually removed, just moved down a line

armanbilge (approved):

japgolly:
@armanbilge Ready for review and merge.

@FabioPinheiro I know it's been two years since you submitted this and sorry for such a long delay. @armanbilge and I are helping out to improve response times going forward. Thanks so much for your patience!

PR #386, opened 2019-10-26
Fix wrong parameter names in CollatorOptions

exoego:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
@@ -65,8 +65,8 @@ class NumberFormat(locales: String | js.Array[String],
 trait CollatorOptions extends js.Object {
   val localeMatcher: js.UndefOr[String]
   val usage: js.UndefOr[String]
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/experimental/intl/Intl.scala
```diff
@@ -65,8 +65,8 @@ class NumberFormat(locales: String | js.Array[String],
 trait CollatorOptions extends js.Object {
   val localeMatcher: js.UndefOr[String]
   val usage: js.UndefOr[String]
-  val sensivity: js.UndefOr[String]
-  val ignore­Punctua­tion: js.UndefOr[Boolean]
+  val sensitivity: js.UndefOr[String]
+  val ignorePunctuation: js.UndefOr[Boolean]
```
exoego: This is weird.
Before this change, `ignore­-Punctua­-tion` shown in Intellij IDEA, so I changed it to `ignorePunctuation`.
GitHub recognizes the change, but `-` is not shown.

Issue #382, opened 2019-09-30
Navigator.languages missing


OndrejSpanel:
While method `Navigator.language` is present (implemented in `org.scalajs.dom.raw.NavigatorLanguage`), [Navigator.languages](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages) is not.

PR #394, opened 2019-12-20
Added `NavigatorLanguage.languages`

catap:
`NavigatorLanguage.languages` is the way to get value that is equiualent to
`Accept-Language` HTTP header in every HTTP request from the user's browser.

Docs: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -5352,6 +5352,15 @@ trait NavigatorLanguage extends js.Object {
...
   def language: String = js.native
+
+  /**
+   * Returns a Array of DOMStrings representing the the user's preferred languages.
+   * The language is described using BCP 47 language tags.
+   * The null value is returned when this is unknown.
+   *
+   * MDN
+   */
+  def languages: js.Array[String] = js.native
```
armanbilge, line 5363: Should the return type be `js.Array[String] | Null` ?

armanbilge: I take it back, doesn't seem to be how this library does things.

armanbilge (approved):

armanbilge:
Fixes #382.

japgolly:
Ok @catap we're merging this now. Thanks so much for your patience and sorry about such a long delay. @armanbilge and I will be helping out as maintainers going forward and we hope to be much more responsive next time. Thanks for the PR!

PR #395, opened 2020-02-06
Upgrade to Scala.js 1.0.0 and make it the default.

exoego:


.travis.yml
```diff
@@ -11,15 +11,15 @@ scala:
 jdk:
   - openjdk8
 env:
-  - SCALAJS_VERSION=0.6.28
-  - SCALAJS_VERSION=1.0.0-RC2
+  - SCALAJS_VERSION=0.6.32
```
sjrd: Please don't change the version of 0.6.x. We intentionally build and publish for older versions as long as possible, to allow people using older versions of Scala.js 0.6.x to use scalajs-dom.

exoego: Ok, updated only 1.0.0.

sjrd:
Thanks :)

PR #397, opened 2020-02-07
Prepare for version 1.0.0.

sjrd:
* Bump the version number to 1.0.0-SNAPSHOT
* Remove all the deprecated things

### DIFF
--- a/src/main/scala/org/scalajs/dom/html.scala
+++ b/src/main/scala/org/scalajs/dom/html.scala
@@ -5,44 +5,22 @@ package org.scalajs.dom
  */
 object html {
   type Anchor = raw.HTMLAnchorElement
-  @deprecated("Obsolete.", "HTML 5")
-  type Applet = raw.HTMLAppletElement
   type Audio = raw.HTMLAudioElement
   type Area = raw.HTMLAreaElement
-  @deprecated("Non-standard", "0.9.8")
-  type AreasCollection = raw.HTMLAreasCollection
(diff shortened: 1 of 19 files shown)


src/main/scala/org/scalajs/dom/html.scala
```diff
@@ -5,76 +5,46 @@ package org.scalajs.dom
...
   type Div = raw.HTMLDivElement
   type DList = raw.HTMLDListElement
-  @deprecated("Non-standard.", "0.9.8")
-  type DT = raw.HTMLDTElement
   type Document = raw.HTMLDocument
   type Element = raw.HTMLElement
   type Embed = raw.HTMLEmbedElement
   type FieldSet = raw.HTMLFieldSetElement
-  @deprecated("Obsolete.", "HTML 4.01")
-  type Font = raw.HTMLFontElement
```
gzm0: Do we actually want to remove these?

(Unsure about how scala-js-dom has handled this in the past, but just seems potentially problematic).

sjrd:
> Do we actually want to remove things that are deprecated in the standards but scala-js-dom does not provide replacements for?

I think we do, yes. Those things might not work in browsers, and if you *really* want them, you can use `js.Dynamic` or use your own facades. You wouldn't want to live with deprecation warnings in your codebase if you can't get rid of them anyway.

gzm0:
OK, fair enough.

PR #401, opened 2020-03-13
Add Clipboard API

exoego:
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard

This PR adds `Clipboard` trait.
It's instance can be obtained via `Navigator.clipboard` property.

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -487,7 +487,22 @@ trait WindowTimers extends js.Object {
 @JSGlobal
 class Navigator
     extends NavigatorID with NavigatorOnLine with NavigatorContentUtils
-    with NavigatorGeolocation with NavigatorStorageUtils with NavigatorLanguage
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -7958,3 +7959,76 @@ object VisibilityState {
...
+ * The NavigatorClipboard interface contains a constructor method allowing
+ * objects implementing it to obtain a Clipboard instance.
+ *
+ * There is no object of type NavigatorClipboard, but some, like Navigator
+ * implements it.
+ *
+ * MDN
+ */
+@js.native
+trait NavigatorClipboard extends js.Object {
```
sjrd: I don't find any evidence that this interface exists on its own. It seems that `def clipboard` should just be directly added in `Navigator`.

exoego: OK, added `def clipboard` directly in `Navigator` class.

PR #402, opened 2020-03-23
Fix #376: Add the Fullscreen API

### LINKED ISSUE #376: Fullscreen API
https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API

exoego:
Closes #376

### DIFF
--- a/src/main/scala/org/scalajs/dom/raw/lib.scala
+++ b/src/main/scala/org/scalajs/dom/raw/lib.scala
@@ -975,7 +975,51 @@ abstract class Element
    */
   def removeAttributeNS(namespaceURI: String,
       localName: String): Unit = js.native
+
+  /**
+   * The Element.requestFullscreen() method issues an asynchronous request to
+   * make the element be displayed in full-screen mode.
+   *
+   * It's not guaranteed that the element will be put into full screen mode.
+   * If permission to enter full screen mode is granted, the returned Promise
+   * will resolve and the element will receive a fullscreenchange event to let
+   * it know that it's now in full screen mode. If permission is denied, the
+   * promise is rejected and the element receives a fullscreenerror event
+   * instead. If the element has been detached from the original document, then
+   * the document receives these events instead.
+   *
+   * Earlier implementations of the Fullscreen API would always send these
+   * events to the document rather than the element, and you may need to be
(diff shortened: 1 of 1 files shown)


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -975,8 +975,52 @@ abstract class Element
...
+   * The Element interface's onfullscreenerror property is an event handler
+   * for the fullscreenerror event which is sent to the element when an error
+   * occurs while attempting to transition into or out of full-screen mode.
+   *
+   * MDN
+   */
+  var onfullscreenerror: js.Function1[Event, _] = js.native
 }
 
+class FullscreenOptions(
```
sjrd: According to MDN, this should be a `trait` with a `var navigationUI` instead of a `class`.

src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -975,8 +975,52 @@ abstract class Element
...
+   *
+   * Earlier implementations of the Fullscreen API would always send these
+   * events to the document rather than the element, and you may need to be
+   * able to handle that situation. Check Browser compatibility in fullscreen
+   * for specifics on when each browser made this change.
+   *
+   * MDN
+   */
+  def requestFullscreen(
+      options: FullscreenOptions = ???): js.Promise[Unit] = js.native
```
sjrd: Prefer `js.native` instead of `???` for default values in JS APIs:
```suggestion
      options: FullscreenOptions = js.native): js.Promise[Unit] = js.native
```

src/main/scala/org/scalajs/dom/raw/lib.scala
exoego: Fixed in https://github.com/scala-js/scala-js-dom/pull/402/commits/5147c02aabf4f3793c33d801c1385d96cd50d98c

src/main/scala/org/scalajs/dom/raw/lib.scala
exoego: Fixed in https://github.com/scala-js/scala-js-dom/pull/402/commits/075d58a381576ca63f41205c6cccf238cddffc19

PR #405, opened 2020-04-01
Add the CSSStyleRule.style property.

shaneorama:


src/main/scala/org/scalajs/dom/raw/Css.scala
```diff
@@ -308,6 +308,13 @@ class CSSStyleRule extends CSSRule {
    */
   var selectorText: String = js.native
   var readOnly: Boolean = js.native
+
+  /**
+   * Returns the CSSStyleDeclaration object for the rule. (Read only)
+   *
+   * MDN
+   */
+  var style: CSSStyleDeclaration = js.native
```
sjrd, line 317: According to [the specification](https://drafts.csswg.org/cssom/#the-cssstylerule-interface), `style` is `readonly` and `SameObject`, so it should be a `val` in Scala.js.

Once it's a `val`, there is no point in mentioning `(Read only)` in the Scaladoc. This is already better explained by the `val` keyword.

shaneorama: That's my opinion as well, I'll change it.

PR #410, opened 2020-05-23
Added ImageCapture

RoelofRuis:
- Added part of the ImageCapture API as per: https://developer.mozilla.org/en-US/docs/Web/API/MediaStream_Image_Capture_API

Does not include `ImageCapture.getPhotoSettings()` as its return value depends on a yet undocumented class.

Does not include `ImageCapture.getPhotoCapabilities()`, might be added later.

- Added ImageBitmap, required as return type for `ImageCapture.grabFrame()`: https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap


src/main/scala/org/scalajs/dom/raw/lib.scala
```diff
@@ -6045,6 +6045,41 @@ class ImageData extends js.Object {
...
+   * MDN
+   */
+  def height: Int = js.native
+
+  /**
+   * Is an unsigned long representing the width, in CSS pixels, of the ImageBitmap.
+   *
+   * MDN
+   */
+  def width: Int = js.native
```
armanbilge, line 6072: Should these be `Long`?

sjrd: No, `Long`s are [opaque to JavaScript](https://stackoverflow.com/q/27821841/1829647). We can never use `Long`s in facade types of JS libraries. If it is an integer value that can exceed the range of `Int`, it must be a `Double` instead.

armanbilge: Hmm, so then this should be `Double`?

sjrd: Yes.

japgolly:
This isn't supported in Firefox or Safari.
https://developer.mozilla.org/en-US/docs/Web/API/MediaStream_Image_Capture_API#imagecapture

Let's postpone this to 2.0 and see how we're gonna handle non-standard stuff.

japgolly:
> This isn't supported in Firefox or Safari.

So IIRC I think we said we're not gonna bother distinguishing between full and partial browser support, and that the onus would be on the user to know the (dynamic) status of the API they're calling. @armanbilge can you confirm? That was the plan right?

armanbilge:
Yep! Decided in https://github.com/scala-js/scala-js-dom/issues/514#issuecomment-913044880. So, we can bring this back to life!

japgolly:
Awesome, thanks! On it....

japgolly (approved):

api-reports/2_12.txt
```diff
@@ -23627,6 +23628,9 @@ experimental/mediastream/AudioTrackList[JT] var onchange: js.Function1[Event, An
 experimental/mediastream/AudioTrackList[JT] var onremovetrack: js.Function1[Event, Any]
 experimental/mediastream/AudioTrackList[JT] def removeEventListener[T <: Event](`type`: String, listener: js.Function1[T, _], options: EventListenerOptions): Unit
 experimental/mediastream/AudioTrackList[JT] def removeEventListener[T <: Event](`type`: String, listener: js.Function1[T, _], useCapture: Boolean?): Unit
+experimental/mediastream/ImageCapture[JC] def grabFrame(): js.Promise[ImageBitmap]
+experimental/mediastream/ImageCapture[JC] def takePhoto(): js.Promise[Blob]
+experimental/mediastream/ImageCapture[JC] val track: MediaStreamTrack
```
armanbilge, line 23633: @japgolly do we want to use `val`s or `def`s for read-only properties? It would be nice to be consistent.

src/main/scala/org/scalajs/dom/WindowOrWorkerGlobalScope.scala
```diff
@@ -96,6 +96,10 @@ trait ImageBitmap extends js.Object {
   /** An unsigned long representing the width, in CSS pixels, of the ImageData.
     */
   def width: Double = js.native
+
+  /** Dispose of all graphical resources associated with an ImageBitmap.
+    */
```
armanbilge, line 101:
```suggestion
  /** Dispose of all graphical resources associated with an ImageBitmap.
    *
    * MDN
    */
```

api-reports/2_12.txt
japgolly: I think we should raise an issue for that rather than catching it here in this PR. The whole codebase currently has this inconsistency.

armanbilge (approved):
I can't believe it!! We made it through every PR.

api-reports/2_12.txt
japgolly: To be continued in #552

japgolly:
:boom: !

PR #412, opened 2020-06-10
Fix #411: console doesn't work in web workers

### LINKED ISSUE #411: console doesn't work in web workers
Using `console` from a web worker results in an exception:

```
Uncaught TypeError: Cannot read property 'console' of undefined"
```

Probably because `window` isn't defined. `console` does exist in the global scope (and also under `self` in that `self.console` exists).

japgolly:


exoego:
@sjrd 
Could you publish new version for this change?
It helps Scala.js users to use `console.log` instead of heavier `println`.

sjrd:
Done (pending sync on Maven Central).

PR #416, opened 2020-06-25
Add WindowOrWorkerGlobalScope

exoego:
This PR adds `WindowOrWorkerGlobalScope` trait, and let `Window` and `WorkerGlobalScope` extends the new trait.
Please refer https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope
Related class/traits, e.g. `OffscreenCanva` and `ImageBitmap`, are also added.

Background: I am using these JS APIs in Worker thread in several projects, via my own quick-and-dirty type facade, so supports in official `scala-js-dom` is highly appreciated.


exoego:
Closing in favor of #430

PR #424, opened 2020-09-30
Added FileReaderSync to dom.experimental

catap:


armanbilge (approved):

src/main/scala/org/scalajs/dom/experimental/FileReaderSync.scala
```diff
@@ -0,0 +1,48 @@
...
+  def readAsDataURL(blob: Blob): URL = js.native
+
+  /**
+   * The readAsText method is used to read the contents of the specified Blob or File.
+   * When the read operation is complete, the readyState is changed to DONE, the loadend
+   * is triggered, and the result attribute contains the contents of the file as a text string.
+   *
+   * MDN
+   */
+  def readAsText(blob: Blob, encoding: String = "UTF-8"): String = js.native
```
japgolly, line 47: `encoding: String = "UTF-8"` ← Does this really work in a facade? @sjrd ?

japgolly:
I'm not up to speed with how experimental stuff is handled in this library yet. Revisit soon'ish....

src/main/scala/org/scalajs/dom/experimental/FileReaderSync.scala
armanbilge: Good catch! Doesn't do anything.

> Methods can have parameters with default values, to mark them as optional. However, the actual value is irrelevant and never used. Instead, the parameter is omitted entirely (or set to undefined). The value is only indicative, as implicit documentation.

https://www.scala-js.org/doc/interoperability/facade-types.html

catap:
@japgolly I'm using for more than a year a fork which is contains this branch ;)

catap:
Ah, I'm wrong. I'm using only `readAsArrayBuffer`.

armanbilge:
> I'm not up to speed with how experimental stuff is handled in this library yet.

@japgolly I think it's fine for this to live in the `experimental` package for 1.x, won't be lonely anyway :) unless we want to put it in the `sharedworker` package or something, because this is specifically a worker API.

For 2.x, is this even still considered experimental? I didn't see any indication of such at https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync.

src/main/scala/org/scalajs/dom/experimental/FileReaderSync.scala
```diff
@@ -44,5 +44,5 @@ class FileReaderSync() extends js.Object {
    *
    * MDN
    */
-  def readAsText(blob: Blob, encoding: String = "UTF-8"): String = js.native
+  def readAsText(blob: Blob, encoding: String = js.native): String = js.native
```
armanbilge: Turns out UTF-8 is _not_ the default:
> The optional parameter specifies encoding to be used (e.g., iso-8859-1 or UTF-8). If not present, the method will apply a detection algorithm for it.

https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync/readAsText#parameters

japgolly:
> @japgolly I think it's fine for this to live in the experimental package for 1.x, won't be lonely anyway :) unless we want to put it in the sharedworker package or something, because this is specifically a worker API.

@armanbilge Ok if you're cool with it I'm cool with it. Let's keep it in `experimental` for now and we'll do a proper reorg in 2.0

japgolly:
Merged, thanks for the PR @catap !

PR #426, opened 2020-10-07
Fix docs examples

sake92:
Improvements:
- openweathermap didnt work (needs an API key), so I used another free API :D
- websockets didnt work when using https, it has to be wss (secure socket)..
- 2 cosmetic changes

### DIFF
--- a/example/src/main/scala/example/Example.scala
+++ b/example/src/main/scala/example/Example.scala
@@ -123,9 +123,7 @@ object XMLHttpRequest{
   @JSExport
   def main(pre: html.Pre) = {
     val xhr = new dom.XMLHttpRequest()
-    xhr.open("GET",
-      "https://www.boredapi.com/api/activity?type=recreational"
-    )
+    xhr.open("GET", "https://www.boredapi.com/api/activity")
     xhr.onload = { (e: dom.Event) =>
       if (xhr.status == 200) {
         pre.textContent = xhr.responseText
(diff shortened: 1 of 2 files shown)


example/src/main/scala/example/Example.scala
```diff
@@ -123,14 +123,10 @@ object XMLHttpRequest{
   @JSExport
   def main(pre: html.Pre) = {
     val xhr = new dom.XMLHttpRequest()
-    xhr.open("GET",
-      "http://api.openweathermap.org/" +
-      "data/2.5/weather?q=Singapore"
-    )
+    xhr.open("GET", "https://www.boredapi.com/api/activity")
```
sjrd, line 126: These examples show up in the final document next to their output, and hence they only have ~40 characters a line. We should try to keep lines as short as possible:
```suggestion
    xhr.open("GET",
      "https://www.boredapi.com/api/activity")
```

example/src/main/scala/example/Example.scala
```diff
@@ -123,14 +123,10 @@ object XMLHttpRequest{
...
     xhr.onload = { (e: dom.Event) =>
       if (xhr.status == 200) {
-        pre.textContent =
-          xhr.responseText
+        pre.textContent = xhr.responseText
```
sjrd, line 129: Same problem here. The previous formatting was better for the short line width.

example/src/main/scala/example/Example.scala
```diff
@@ -166,9 +162,7 @@ object AjaxExtension {
                 .ExecutionContext
                 .Implicits
                 .global
-    val url =
-      "http://api.openweathermap.org/" +
-      "data/2.5/weather?q=Singapore"
+    val url = "https://www.boredapi.com/api/activity"
```
sjrd, line 165:
```suggestion
    val url =
      "https://www.boredapi.com/api/activity"
```

sjrd (changes requested):
I have a few formatting comments, but otherwise LGTM.

sake92:
Makes sense. All addressed now! @sjrd

sjrd (approved):

PR #429, opened 2020-11-13
IDBVersionChangeEvent.newVersion can be null

japgolly:


japgolly:
Spec: https://www.w3.org/TR/IndexedDB/#events

src/main/scala/org/scalajs/dom/raw/Idb.scala
```diff
@@ -123,7 +124,7 @@ class IDBObjectStore extends js.Object {
 }
 
 trait IDBVersionChangeEventInit extends EventInit {
-  var newVersion: js.UndefOr[Int] = js.undefined
+  var newVersion: js.UndefOr[Int | Null] = js.undefined
```
sjrd, line 127: The convention so far has been to use the boxed classes when we have primitive-or-null. So this should be
```suggestion
  var newVersion: js.UndefOr[Integer] = js.undefined
```
In a future where Scala has explicit nulls, that will change. But we're not there yet.

src/main/scala/org/scalajs/dom/raw/Idb.scala
```diff
@@ -142,9 +143,11 @@ class IDBVersionChangeEvent(typeArg: String,
   /**
    * Returns the new version of the database.
    *
+   * This is null when the database is being deleted.
+   *
    * MDN
    */
-  def newVersion: Int = js.native
+  def newVersion: Int | Null = js.native
```
sjrd, line 150:
```suggestion
  def newVersion: Integer = js.native
```

sjrd (changes requested):
Thanks for the PR.

src/main/scala/org/scalajs/dom/raw/Idb.scala
japgolly: Np. Updated.

armanbilge (approved):

PR #430, opened 2020-11-14
Add WindowOrWorkerGlobalScope

japgolly:
Supersedes #416.

Additionally, this fixes the following:

* IDBEnvironment removed because it's officially obsolete and "although
  it may still work in some browsers, its use is discouraged since it
  could be removed at any time".

* indexedDB now wrapped in js.UndefOr because it's not always available

* indexedDB added to WorkerGlobalScope

* caches return type changed from js.Any to js.UndefOr[CacheStorage]

* caches added to Window

* CacheStorage moved into experiment.cachestorage as it's available from
  Window too

* Add missing methods to window and web-worker scopes:
  * crossOriginIsolated
  * isSecureContext
  * origin

Ref: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope


japgolly:
Oh no. I now see two things.

1. There is overlap with PR #416. I didn't even see it until I raised this PR.
2. I only added the properties and not the methods. (*sigh*)

Regarding the surface area of this PR that intersects with #416, this PR is more accurately typed. The only concern with this PR is that it is incomplete. We can either merge this first and then follow up with a revised #416, or combine this and #416 to form a new PR. @sjrd please let us know which you'll allow and we'll make it happen.

@exoego Happy to work together to combine our PRs, or if you couldn't be bothered anymore, I'm happy to integrate your PR into mine. Yours did come first after all :D よろしくお願いします

exoego:
@japgolly I am fine if you make this PR complete by bringing some missing parts from #416 😄

japgolly:
np @exoego I'll combine our PRs in the next few days. :+1: And thanks for your work there. I've got so much other work to be doing so I think I'd try if I myself had to write all the new facades in your PR :sweat_smile:

japgolly:
> I'll combine our PRs in the next few days

A few days, a month, they're the same thing right? Hmmm.... Moving on. I've merged #416 into this PR and made a few little fixes. This is good to go now.

exoego:
@japgolly Could you update this PR so that @sjrd may review ?

japgolly:
Yeah I have already. I did on 15th Dec 2020. There's a comment right above yours @exoego :smile:

japgolly:
Did a rebase.

sjrd:
For scalajs-dom 2.x, which doesn't have to support Scala.js 0.6.x, a much better alternative to defining `WindowOrWorkerGlobalScope` is to simply define its members as `@js.native` members of the `dom` package object. For example, like how `fetch` is defined in
https://github.com/scala-js/scala-js-dom/pull/458/files#diff-492b8ef836683ba613a558252a12fa68f30e3fd2a220fc21a1082a1d7319fe2e

armanbilge:
@japgolly is this binary compatible? I don't think this change is:

> CacheStorage moved into experiment.cachestorage as it's available from Window too

I think the rest would be okay though. So either undo just the `CacheStorage` change and merge or put off to 2.0?

japgolly:
@armanbilge Nah it's not. I'll update this soon (not today though) for compat.

japgolly:
Ready to go @armanbilge !

armanbilge:
This is a huge one so I'm just going to check changes with bincompat implications and assume you got the types right and/or they can be fixed compatibility in the future.

scalafix/src/main/scala/org/scalajs/dom/scalafix/MutableState.scala
```diff
@@ -47,10 +52,20 @@ final class MutableState {
 
     val b = SortedSet.newBuilder[Result]
 
+    def deprecationSuffix(ver: Option[String]): String =
+      ver match {
+        case None    => ""
+        case Some(v) => s"  (@deprecated in $v)"
```
armanbilge, line 58:
```suggestion
        case Some(v) => s" (@deprecated in $v)"
```
Unless it was intentional? :)

armanbilge (approved):

scalafix/src/main/scala/org/scalajs/dom/scalafix/MutableState.scala
japgolly: Yeah, double-space is intentional cos when I first tried single space it was a bit too cluttered for easy reading. This is already on the `series/v1.x` branch btw, here I'm just showing additional deprecations without changing the format.

japgolly:
Thanks @armanbilge ! :rocket:

armanbilge:
Thanks to you!! This was an important one :)

Do you think we might get 1.2 out today?

japgolly:
Ooooo do you want to? Are we ready? Anything else you want to add before we do? The only thing left on my list is to get #424 merged in for 1.2.0. I'm not sure if I'll get to it day though, feel free to take that one over if you want. Apart from the experimental question, I was also planning to just test `arg = "UTF-8"` instead of `arg = js.native` in something that actually runs and just see what happens.

armanbilge:
@japgolly did you see my comment https://github.com/scala-js/scala-js-dom/pull/424#discussion_r692607746? It's explained in the Scala.js docs (it's essentially equivalent to `arg = js.native`).

> Ooooo do you want to? Are we ready? Anything else you want to add before we do?

I think so! The most important question for me is did we _break_ anything, which I don't think so. If we _forgot_ something or there are more nice-to-haves, theoretically we could do a 1.2.1 or 1.3 etc. I really really hope that getting a release out with all these merged PRs will encourage the community to become more engaged :)

PR #431, opened 2020-11-15
Make crypto.subtle.importKey return a js.Promise[CryptoKey]

japgolly:
See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey


armanbilge (approved):

japgolly:
Merged as this is already approved, thanks @armanbilge

PR #432, opened 2020-11-15
Avoid forcing evaluation of crypto

japgolly:
Having `val crypto` in the package object means that when it's undefined
an exception is thrown immediately, which makes it impossible to use the
WebCrypto facade in such a case.

An example situation where this is preventative is using the WebCrypto
API from Node where its available as a library and not globally
available.


armanbilge:
+1! I had this exact problem in https://github.com/http4s/http4s/pull/4938.

FWIW, I think the WebCrypto facade should be its own module without any reference to DOM if possible, to maximize cross-compatibility. But not important here :)

armanbilge (approved):

sjrd:
For scalajs-dom 2.x, even better is to use an `@js.native @JSGlobal val`, like is done at
https://github.com/scala-js/scala-js-dom/pull/458/files#diff-c31f4c3c58ea375faca860d35fc697e45da565e346fc33a17bedbb9d9f92c2ff

japgolly:
Merged as this is already approved, thanks @armanbilge

PR #434, opened 2020-11-23
Use IDBVersionChangeEvent for IDBOpenDBRequest.onblocked

unarist:
An event type of IDBOpenDBRequest.onblocked should be `IDBOpenDBRequest`.

ref: https://developer.mozilla.org/en-US/docs/Web/API/IDBOpenDBRequest/blocked_event


japgolly:
LGTM, thanks both for the PR and your patience, @unarist . This project has additional maintainers now so we'll try to be much more responsive in the future. Cheers

japgolly (approved):

PR #436, opened 2020-12-22
Add onwheel

vhiairrassary:
Scala-js-dom currently supports [onmousewheel](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onmousewheel) but as written on MDN:
> Do not use this wheel event. This interface is non-standard and deprecated. It was used in non-Gecko browsers only. Instead use the standard wheel event.

As mentioned above it is [not supported on Firefox](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onmousewheel#Browser_compatibility). This PR adds [onwheel](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onwheel).

You can test with the following code:
```scala
import org.scalajs.dom
import org.scalajs.dom.html.Div

object Main {
  def main(args: Array[String]): Unit = {
    val onwheelChild = dom.document.createElement("div").asInstanceOf[Div]
    onwheelChild.textContent = "onwheelChild"
    onwheelChild.onwheel = event => {
      println("onwheel: " + event)
    }
    dom.document.body.appendChild(onwheelChild)
    
    val onmousewheelChild = dom.document.createElement("div").asInstanceOf[Div]
    onmousewheelChild.textContent = "onmousewheelChild"
    onmousewheelChild.onmousewheel = event => {
      println("onmousewheel: " + event)
    }
    dom.document.body.appendChild(onmousewheelChild)
  }
}
```

⚠️ [MouseWheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseWheelEvent) is different than the standard [WheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent), and Scala-js-dom currently uses `WheelEvent` for the deprecated `onmousewheel`. I didn't want to change the Scala-js-dom code as it would break the compatibility for the current users. Note: on my computer, Chrome returns a `WheelEvent` for both events. Here is a screenshot showing the output of the above snippet:
Screenshot 2020-12-22 at 18 55 20

Also I didn't add documentation. Feel free to ask me to do it if required


armanbilge (approved):

japgolly:
Thanks for the PR and your patience @vhiairrassary!

PR #437, opened 2020-12-30
Fix XMLHttpRequest.onerror param type: ErrorEvent -> ProgressEvent

raquo:
Apparently `xhr.onerror` listener does not actually receive `ErrorEvent`-s. Confirmed with [MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/error_event) and the [spec](https://xhr.spec.whatwg.org/#events), and tested in Chrome – `xhr.onerror` listener is getting `ProgressEvent`-s.

MDN & spec actually say that other xhr event listeners get `ProgressEvent`-s too, whereas in scala-js-dom `onabort`, `onloadstart` et al get `js.Any` instead. Unless anyone knows why this is the case (IE misbehaving?), should we fix all of these too?


japgolly:
This same change was recently merged as part of another PR #379

Thanks though for the PR, and thanks for your patience!

PR #452, opened 2021-07-14
Bump to 2.0.0-SNAPSHOT.

### LINKED ISSUE #451: What to do about the Scala 3-specific artifact?
We merged support for Scala 3 in #450. The question is now: what do we do about the Scala 3-specific artifacts of scalajs-dom?

If we release a version 1.2.0 with a new artifacts for `_sjs1_3`, we can create problems in the binary ecosystem:

* There can be a library A compiled against the _2.13 artifact of scalajs-dom 1.1.0
* And a library a B compiled against the _3 artifact of scalajs-dom 1.2.0

Mixing A and B in the same project will cause both scalajs-dom_2.13 v1.1.0 and scalajs-dom_3 v1.2.0 to be on the classpath. This can cause serious issues.

There are two solutions, IMO:
* *Not* release the _3 artifact. After all, it does not bring any actual value compared to the _2.13 artifact at the moment. Users will still have to declare their dependency on scalajs-dom using `.cross(CrossVersion.for3Use2_13)`, as they currently do.
* Bump to scalajs-dom 2.0.0, clearly advertising that there is binary conflict with the previous versions of scalajs-dom.

@gzm0 Any opinion?

Anyone else?

sjrd:
So that we can release v2.0.0 with the Scala 3 artifact.

Fixes #451.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,7 @@ ThisBuild / crossScalaVersions := {
 ThisBuild / scalaVersion := crossScalaVersions.value.head
 
 val commonSettings = Seq(
-  version := "1.2.0-SNAPSHOT",
+  version := "2.0.0-SNAPSHOT",
   organization := "org.scala-js",
   scalacOptions ++= Seq("-deprecation", "-feature", "-Xfatal-warnings")
 )


gzm0:
I have commented on #451 regarding this.

gzm0:
So, I'm still unsure about what the community consensus is regarding that. But the major bump in itself is correct. So I'm going to approve this.

@sjrd if you want my opinion on what to do with regards to this, I'll need significantly more time, because I need to learn how the entire dotty compat thing is supposed to work and how it works in practice. I'm not able to follow the conversation in #451.

gzm0 (approved):

sjrd:
Given the reactions at https://github.com/scala-js/scala-js-dom/issues/451#issuecomment-891141873, I'm going to merge this, and proceed with a bunch of cleanups.

armanbilge:
@sjrd Thanks for going forward for this! :) Can I help with any/all of the cleanup tasks? I ask, just to make sure that there aren't any duplicate efforts.

sjrd:
I have started doing the cleanups today. If I have to move to other things, I'll let you know.

PR #458, opened 2021-08-04
Cleanups, part 1.

sjrd:
This is a first wave of cleanups:

* Move everything from `raw.*` to `dom.*`
* Remove all extensions in `ext.*`
* Introduce `Seq` support directly on `DOMList` through its companion object
* Likewise `Map` support for `NamedNodeMap`
* And some moves from `experimental.*` to `dom.*`, most notably the `fetch` API, which is now used in the readme instead of `XMLHttpRequest`/`ext.Ajax`


src/main/scala/org/scalajs/dom/experimental/package.scala
```diff
@@ -1,51 +1,179 @@
...
+  lazy val Response: dom.Response.type = dom.Response
+
+  @deprecated("use dom.ResponseInit instead", "2.0.0")
+  type ResponseInit = dom.ResponseInit
+
+  @deprecated("use new dom.ResponseInit { ... } instead", "2.0.0")
+  object ResponseInit {
+    def apply(_status: Int = 200, _statusText: ByteString = "OK",
+        _headers: HeadersInit = js.Dictionary[String]()): ResponseInit = {
+      new ResponseInit {
```
armanbilge, line 68: A pedagogical question: why is `new` preferable to an `apply` method on a companion object?

armanbilge (approved):
Nice cleanups! With the caveat that I've only used a small fraction of this API, specifically, Fetch and Web/Service Workers.

src/main/scala/org/scalajs/dom/experimental/package.scala
sjrd: For this case in isolation, it's not very different. But for many config objects, which have at least one completely optional field, the `new` idiom is much better. Therefore, it is better to use that idiom *everywhere* to design a more uniform API.

Also, in Scala 3, you can just write `new { ... }` without the type name, which you cannot do with the `apply` idiom. That means you don't have to remember the type name or look it up when translating a documentation snippet in JavaScript.

armanbilge: I see, that is much nicer, on both fronts! Thanks for explaining :)

japgolly:
I'll give this a proper look tomorrow but a question for now: we'll need an automated migration strategy (be it full on scalafix and just regex magic like I often do with scalajs-react [example](https://github.com/japgolly/scalajs-react/blob/master/doc/changelog/2.0.0.md#migration)). Any thoughts on this yet? If not, this is me planting the seed.

I'm thinking I'll create an issue and start a discussion tomorrow. We might want to make sure that all PRs that break compatibility with 1.x also include whatever we decide is appropriate for migration.

japgolly:
Also this PR targets `master`, I feel like we should open a 2.x branch and target that until we're ready for 2.0.0. An example is that we might want to introduce an additional 1.x release that eases migration.

Also! I think we should be merged all the outstanding PRs first before a big cleanup. Unfortunately this creates more work for you @sjrd now that you've already started but I think it's pretty unfair to push that burden to people who've raised PRs already. We should be making PRs as easy as possible for casual contributors to encourage more collaboration.

armanbilge:
+1 to working on a separate branch. At the very least we should run the workflows on outstanding PRs and/or ping the contributors to see what's realistic to merge.

sjrd:
> I'll give this a proper look tomorrow but a question for now: we'll need an automated migration strategy (be it full on scalafix and just regex magic like I often do with scalajs-react [example](https://github.com/japgolly/scalajs-react/blob/master/doc/changelog/2.0.0.md#migration)). Any thoughts on this yet? If not, this is me planting the seed.
> 
> I'm thinking I'll create an issue and start a discussion tomorrow. We might want to make sure that all PRs that break compatibility with 1.x also include whatever we decide is appropriate for migration.

So far, all the changes in this PR are actually very compatible. Most things are only moved around, and I left deprecated aliases everywhere. So nothing is really necessary to migrate per se. It's only after if you want to get rid of the deprecations. And even here, in most cases it should just be changing some imports.

The only real breaking changes are the removal of `Color` and other obscure stuff in `dom.ext.*`. I suspect those have very low usage anyway.

sjrd:
Regarding existing PRs, it's up to you, but:

* the most recent open PR is from January, so it's not clear that the authors will be interested/responsive to address rebase and other comments right now
* this PR makes sure not to break file *renames*, as interpreted by git; so the existing PRs shouldn't be hard to rebase after this PR has been merged, if there is still interest in them

armanbilge:
> the most recent open PR is from January, so it's not clear that the authors will be interested/responsive to address rebase and other comments right now

I merged in main and re-ran CI on these, it looks like they all pass CI actually and many seem in decent shape. See also https://github.com/scala-js/scala-js-dom/pull/461 to enable MiMa.

sjrd:
> I merged in main and re-ran CI on these, it looks like they all pass CI actually and many seem in decent shape.

In general, it's much better to *rebase* PR branches, then force-push them, instead of merging master into PR branches. Merge commits in PR branches create weird log graphs. Most main Scala repos demand clean PR branches without merge commits. I think it would be good to keep that property.

sjrd:
It's also possible to use "Squash and merge" at the time of merging instead, although it is then important not to forget to use that option instead of "Merge".

armanbilge:
> I think it would be good to keep that property.

After having contributed to a couple of your projects, I know this is your preferred style, and for the record I do think it's a very good one for the reasons you mentioned. Especially for important projects such as Scala.js itself. But, imo this requirement makes casual contributing much much more difficult, and I agree with @japgolly that we want to do as much as we can to support this here.

> Most main Scala repos demand clean PR branches without merge commits.

Eh, define "most" :) E.g. none of the [typelevel](https://github.com/typelevel/) projects I've contributed to do this. So this is evidence to me that the merge-way works fine too.

sjrd:
Yeah I meant most scala/* projects. I don't know about other organizations.

armanbilge:
Also this seems to be @japgolly's style of working too, e.g. https://github.com/japgolly/scalajs-react/pull/938

lihaoyi:
Using the squash-merge button all the time seems like it would be a reasonable compromise to keeping clean histories without inconvenience; IIRC it can be enforced in the repo settings too.

In my experience, nobody reviews the individual commits (messages + diffs) as carefully as the entire PR, someone digging through git history 6 months down the line does not care for either merge commits or the individual commits that went into each pull request. Squash merge ensures that the thing that is actually reviewed and useful is what makes it into the master git history

sjrd:
> In my experience, nobody reviews the individual commits (messages + diffs) as carefully as the entire PR

Well, to counter that with a data point: in scala/scala and in scala-js/scala-js, as well as most other scala-js/*, we *do* review individual commits, and they must make integral sense. The composition of what goes into a commit, and what commits make up a PR, are very much taken into account when reviewing.

sjrd:
Regarding the separate branch, the `master` branch already contains several changes that are 2.x-only, notably dropping Scala.js 0.6.x support.

If there is a need for a 1.2.0, we could make another branch 1.x off 22e4a70994844d1fc4e7bed10b562355f3b4591c, which is the last commit before the changes for Scala 3 and scalajs-dom 2.x in general.

armanbilge:
I think the question is whether there should be one more 1.x release with some of the open PRs merged in. Personally, I think this would be both an easy and a nice thing to do, for those who cannot hop to 2.x immediately. And secondly, whether any sort of migration support can be baked in as @japgolly suggested, although from @sjrd's [comment](https://github.com/scala-js/scala-js-dom/pull/458#issuecomment-894780078) it seems this probably won't be necessary.

> Well, to counter that with a data point: in scala/scala and in scala-js/scala-js, as well as most other scala-js/*, we do review individual commits, and they must make integral sense.

And I'm very very glad you do!! A regression in one of those projects is a very different beast than a regression in a facade, imo.

japgolly:
### Git Commit policy

My personal opinion on keeping git histories pristine is a bit more relaxed. For mainline branches (eg. master/main, series/v1.x, etc) I find it unhelpful to have a bunch of random commits and I like commits to be atomic/complete; to be incomplete would mean a mainline branch is broken and that should be avoided, and as we all know, things like when a feature breaks it's easy to revert, readers can read a PR associated with a single intent, blah blah blah. What I don't agree with is the every topic branch / PR needs to adhere to the same standards, in fact I think that's quite detrimental for a few reasons:

1) information is lost; if a bug persists it's possible to later go back and see which intermediate commit may have accidentally broke another aspect of the same feature. It also preserves alternate attempts at solving problems which may become quite valuable later to either re-read or reinstate entirely once new information comes to light.

2) Git has the power to hide all of the intermediate commits and just show the aggregated merge commit on a mainline (which is what you get with a rebase, squash and ff). In other words you can simulate a squash later without losing information.

3) Asking contributors to rebase and squash themselves all the time is a burden that need not exist. I want to see maximum involvement and contribution from the community rather than a minimal subset of pedants (myself included! I'm very pedantic by nature and have to consciously remind myself to focus on the bigger picture).

4) I'm older and busier and (at least in some small ways) wiser than I was before. I'm happy to help maintain this repo because frankly it wasn't being maintained at all with PRs completely ignored for 6+ months. Maintenance is a burden and I've seen @sjrd  working on so many things including the Scala 3 compiler. So I'm happy to take on this maintenance to help myself and the community. Faffing around trying to get the best commit messages, the best looking git graphs, the best whitespace formatting, etc is something I do not want to spend my time on because I honestly believe it has no value to myself and the community (my personal goals here), and is in fact detrimental by wasting time and energy that could be spend elsewhere, and wasting good will with potential contributors. Sometimes those concerns are justified (especially in a vacuum because they genuinely have merit), but in this case I don't see any argument for how a strict focus on those aspects is going to improve the life of us maintainers or the community.

japgolly:
Ok everyone, it's 10:30am over here in Australia and I just want to smash out work on this repo and get as much done as possible (without breaching our quality standards of course). I'm not quite sure what authority looks like in this new world but because @sjrd mentioned that his plan is to leave maintenance completely up to myself and @armanbilge , and that @armanbilge put his hand up as a deputy, I'm going to assume for today that makes me the sole and main maintainer. Which this new privilege I'm just gonna do my best today and run it the way I do my other repos.

If I'm overstepping in any way, it's not my intent and please let's have that discussion. Until I hear back, I'll just get to work :)

japgolly:
Oh and finally, just in case I'm not being as clear as I think I am, @sjrd I've always appreciated your attention to detail and the time you put it to adhere to your own standards. The individualised commits in this PR alone demonstrate that really awesome attribute of yours and I want to clarify that not only do I not mean to dismiss it (in case my words above accidentally convey that), but I also appreciate it. What I'm mainly trying to communicate is that I don't want to hold everyone, especially casual maintainers to that standard because "big picture tradeoffs, see above". I always think I'm being pretty clear when I type away on the internet but whether it's something about the medium or just me being non-neurotypical, there always seems to be misunderstanding. So I apologise for being so verbose but I'm trying to prevent a potential misunderstanding here too. No ill will is intented :)

armanbilge:
@japgolly sounds like a plan! Feel free to assign any stuff to me so you can focus on the decision-making aspects of maintainership.

japgolly:
@armanbilge you're online! I stalked you a little and found out your in Seattle so I assume it's ? o'clock in the morning over there. I found you on Twitter, I'll ping you there instead of spamming this PR

japgolly:
@sjrd FYI we're going to be coming back to this PR a bit later. If you want to work on Cleanups, part 2 and 3 and base them on this PR, go for it and don't let us block you.

sjrd:
Rebased. This is ready for a new review.

armanbilge:
@japgolly are things settled enough on main that we can re-review and get this effort merged? Or do you still have some things to do before?

japgolly (changes requested):
Looks great to me! Thanks for rebasing @sjrd . I pushed two commits of my own:

1. merge `master` in which both brings in the test support, and fixes the merge conflicts
2. added a test case for `NodeList` ops

Anyway, there are two issues before we merge:
1. `object rawold` exists but isn't used anywhere. Rename back to `raw`
2. Adding `@deprecated` to an object doesn't cascade to its type members. Eg. I can write `val x: css.FontFaceRule = ???` without getting a deprecation warning. The annotation will unfortunately, need to be applied to all type members.

sjrd:
Rebased and addressed the review. I integrated the changes of the merge commit directly in the relevant commit during the rebase.

src/main/scala/org/scalajs/dom/lib.scala
```diff
@@ -6375,14 +6398,40 @@ class StyleSheet extends js.Object {
 }
 
 @js.native
-trait DOMList[T] extends js.Object {
+trait DOMList[+T] extends js.Object {
```
japgolly, line 6401: Very nice btw

japgolly (approved):

japgolly:
Fantastic, thank you @sjrd !!

Issue #427, opened 2020-11-09
SharedWorker onconnect has wrong type


japgolly:
`onconnect` is currently defined as being a Scala function instead of a JS function, which is basically the equivalent of

```scala
onconnect = {}
```

which doesn't work.

PR #461, opened 2021-08-08
Enable MiMa

armanbilge:
This adds MiMa to the build, checking against 1.1.0. A bit silly for v2 of course, but helpful if we want to release another 1.x series.

The only problem found since 1.1.0 was introduced in https://github.com/scala-js/scala-js-dom/issues/427.

My 2 cents: since we are releasing a binary-incompatible v2 anyway, we should freeze binary compatibility in 1.x. Maybe we should make a series/1.x branch (possibly without #427), merge the approved/compatible PRs into there, and then forward merge them into main (aka series/2.x).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,8 @@ jobs:
         run: sbt "++${{ matrix.scalaversion }}" example/compile
       - name: scalafmt
         run: sbt "++${{ matrix.scalaversion }}" scalafmtCheck
+      - name: mima
+        run: sbt "++${{ matrix.scalaversion }}" mimaReportBinaryIssues
   readme:
     runs-on: ubuntu-latest
     steps:
--- a/build.sbt
+++ b/build.sbt
@@ -54,6 +54,13 @@ scalacOptions ++= {
   else Nil
 }
 
+import com.typesafe.tools.mima.core._
+mimaPreviousArtifacts := Set("org.scala-js" %%% "scalajs-dom" % "1.1.0")
+mimaBinaryIssueFilters ++= Seq(
+  // #427 SharedWorker onconnect has wrong type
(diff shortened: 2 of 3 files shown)


sjrd:
Usually, I'm all for MiMa, and I enable it in all the repos. But it is really not good at checking JS facade types. It has no idea what is or isn't binary compatible from the point of view of facade types. Since scalajs-dom is almost exclusively filled with facade types, I think enabling MiMa will cause much more pain than benefit.

armanbilge:
I see, so if I understand correctly, binary compatibility _is_ important here, it's just that MiMa flags up a lot of false positives?

sjrd:
Yes, that's right.

armanbilge:
Then at least for me, MiMa is still a useful tool for catching bincompat issues even if there will be a lot of false positives ... maybe I'll change my mind after trying it. But if @japgolly isn't worried, then that's fine.

> It has no idea what is or isn't binary compatible from the point of view of facade types.

Do you mind expanding on this a little bit i.e. what makes a facade different? I worry there's something here I'm not understanding.

sjrd:
MiMa checks binary compatibility at the class file level, following the JVM linking requirements. For Scala classes, this is a very good model for Scala.js compatibility as well, because sjsir file compatibility is very close to class file compatibility. However, for facades, sjsir compatibility is a lot more permissive. Here is a non-exhaustive list of changes that would be binary incompatible for Scala classes, but that are compatible for JS facade types:

* Removing a member
* Changing the type or signature of a member
* Adding a field in a trait
* Adding an abstract member in a trait

Most changes in the DOM facades belong to one of those, since it is often fixing types and adding members in classes/traits. They would be flagged as false positives.

The following are changes that are indeed incompatible in both formats:

* Removing a trait/class/object
* Changing a class into a trait or vice versa

armanbilge:
Aha, thank you for explaining in detail. That is a much more permissive than I realized. I agree, we don't need MiMa here.

PR #473, opened 2021-08-11
Fix #445 return type for Header#get

### LINKED ISSUE #445: Headers#get won't return undefined but typed as js.UndefOr

armanbilge:


japgolly:
1. Are you absolutely sure it will never return `undefined` in any circumstance/browser?
2. Breaks compat. Will have to target 2.x

armanbilge:
Good questions 🤔 

1. Not according to the spec. I have no idea how to answer that question in general 🙃 
2. I'm still confused about what exactly breaks compat. I assume this is breaking source compat but not binary compat?

japgolly:
1. Gotta just google it I'm afraid. I usually start with mdn, then if needed read through the linked specs. 
2. Hmmm interesting, actually my understanding of binary compat when it comes to SJS facades is full of unverified assumptions so I'm not much of an authority there. But source compat definitely breaks. Imagine `.get(???).isDefined`

armanbilge:
1. Yup, I checked in MDN and it didn't say anything about `undefined`, just possibly returns `null`.
2. I got an excellent primer on binary compat for facades here: https://github.com/scala-js/scala-js-dom/pull/461#issuecomment-894829615

Regarding source compat: we should have clarified what freezing backwards compat looks like for the next 1.x. I assumed we'd freeze bincompat (semantic versioning) but not source compat. If we also freeze source compat, then many PRs are no longer applicable to 1.x. I'm fine either way, let's just be really clear on this :)

sjrd:
I can confirm that this PR does not break binary compatibility. (In a Scala class, it would, but not in a JS facade.)

japgolly (approved):

japgolly:
Thanks @armanbilge :+1:

PR #480, opened 2021-08-13
Set version scheme for publishing

armanbilge:
I'm not sure what was used in the `0.x` stage of this project, so I'm specifying `semver-spec` and not `early-semver` to be cautious, doubt it'll matter to anyone.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,6 +73,8 @@ scalacOptions ++= {
   else Nil
 }
 
+versionScheme := Some("semver")
+
 pomExtra := (
     <developers>
       <developer>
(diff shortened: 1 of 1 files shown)


japgolly:
WDYT @sjrd ? How strict were you with the 0.x releases?

armanbilge:
If I had to guess (and knowing @sjrd!) almost definitely `early-semver`, that's what Scala.js itself uses.

sjrd:
It's been `early-semver`, yes, indeed.

japgolly:
Thanks!

PR #483, opened 2021-08-13
Sort API members by name

armanbilge:
@japgolly is this what you are looking for?


japgolly:
Close! Types still need to come first though. Sorting key should be `(prefix, name, body)`

japgolly:
Wait, no no no. Yeah I didn't think it would be this easy. `name` here is the scope name, not the member name. The member name is currently lost from `MutableState`'s pov. You'd have to change the `.add` methods in `MutableState.Scope` to also accept name and then plumb it through

armanbilge:
Hmm ok maybe this is more one for you 😅

japgolly:
hehe np

japgolly:
Superseded by #484

PR #491, opened 2021-08-15
Fix #488 `HTMLMediaElement.play()` should return `Promise`

### LINKED ISSUE #488: HTMLAudioElement.play should return Promise, not Unit
This should be a backwards-compatible change.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

The HTMLMediaElement play() method attempts to begin playback of the media. It returns a Promise which is resolved when playback has been successfully started. Failure to begin playback for any reason, such as permission issues, result in the promise being rejected.

armanbilge:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

The spec says:

> Browsers released before 2019 may not return a value from `play()`

So I made the return value `js.UndefOr[js.Promise[Unit]]`.


japgolly:
Awesome @armanbilge  LGTM

PR #498, opened 2021-08-22
Use GitHub Action: japgolly/setup-everything-scala

japgolly:
Which makes use of the brand new GitHub Actions feature: action
composition. This allows us to just have a single GHA for everything
Scala and Scala.JS, and centralise and hide boring details like caching.


armanbilge (approved):
Awesome! 🚀 bye-bye boilerplate!

Issue #479, opened 2021-08-13
Change auto-formatting settings


japgolly:
Once our PR queue is cleared, on the 2.x branch lets:

* [x] Add import normalisation and sorting to auto-formatting (I use it everywhere and it's GLORIOUS)
* [x] I wouldn't mind increasing the column width from 79 to 120
* [x] Format ScalaDoc in a more concise way - it takes up so much vertical space atm
* [ ] Force use of `override`
* [ ] sort members
* [ ] More?

armanbilge:
Yes PLEASE! Formatting for `build.sbt` as well.

japgolly:
I think we're good enough. I don't see a rule for the other features (and I'm not gonna write myself this time) so I'll just close this hey

PR #518, opened 2021-08-29
Auto-format imports

japgolly:
This is part of #479. I'd love to do all of #479 on the 1.x branch but it would probably just make it harder to merge the outstanding PRs. Doing just the import side now though shouldn't be very disruptive and it lays a little more groundwork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,16 +32,16 @@ jobs:
         run: ./prepareForScala210.sh
 
       - name: Build
-        run: sbt "++${{ matrix.scalaversion }}" package
+        run: sbt -DCI=1 "++${{ matrix.scalaversion }}" package
 
       - name: Test generate documentation
-        run: sbt "++${{ matrix.scalaversion }}" doc
+        run: sbt -DCI=1 "++${{ matrix.scalaversion }}" doc
 
       - name: Build examples
-        run: sbt "++${{ matrix.scalaversion }}" example/compile
+        run: sbt -DCI=1 "++${{ matrix.scalaversion }}" example/compile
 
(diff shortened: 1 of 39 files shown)


.github/workflows/ci.yml
```diff
@@ -32,16 +32,16 @@ jobs:
         run: ./prepareForScala210.sh
 
       - name: Build
-        run: sbt "++${{ matrix.scalaversion }}" package
+        run: sbt -DCI=1 "++${{ matrix.scalaversion }}" package
```
armanbilge, line 35: What does `DCI=1` do?

armanbilge (approved):

.github/workflows/ci.yml
japgolly: Facilitates this: https://github.com/scala-js/scala-js-dom/blob/b0fa8f9f06b8aa8223688ce425f6a786fc7a0a81/build.sbt#L31-L41

So `-Xfatal-warnings` isn't enabled locally, we need actual warnings so that scalafix can identify and remove unnecessary imports. Where as in CI, we expect that scalafix has already been run (as part of `prePR`) so we turn those warnings back into errors.

armanbilge: Wait a minute, doesn't `-D` set system properties and not environment variables?

japgolly:
Oh I see you already approved, cool, merging

armanbilge:
@japgolly Oops, maybe I should have revoked my approval until you clarified https://github.com/scala-js/scala-js-dom/pull/518#discussion_r698094655 😅

japgolly:
faaaaaaaaaaaark, good catch

PR #522, opened 2021-08-30
Fix CI detection

japgolly:
Follow up to #518

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,7 +28,7 @@ ThisBuild / crossScalaVersions := {
 }
 ThisBuild / scalaVersion := crossScalaVersions.value.find(_.startsWith("2.13.")).get
 
-val inCI = Option(System.getProperty("CI")).exists(_.contains("1"))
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -28,7 +28,7 @@ ThisBuild / crossScalaVersions := {
 }
 ThisBuild / scalaVersion := crossScalaVersions.value.find(_.startsWith("2.13.")).get
 
-val inCI = Option(System.getenv("CI")).exists(_.contains("1"))
+val inCI = Option(System.getProperty("CI")).exists(_.contains("1"))
```
armanbilge, line 31: why not Scala's `sys.props` btw?

japgolly: Why not `System.getProperty` lol? Same thing either way

armanbilge: Because, why scala? 😝 lol nbd, infrastructure is ugly anyway 🤣  

```suggestion
val inCI = sys.props.get("CI").exists(_.contains("1"))
```

armanbilge (approved):

Issue #514, opened 2021-08-28
Document policy for APIs that are experimental and/or not supported by all browsers


armanbilge:
This comes up time and time again e.g. in #469. If we can determine some guidelines for this, should be easy enough to follow.

armanbilge:
See https://github.com/scala-js/scala-js-dom/issues/469#issuecomment-907695865, this seems like the way forward.

japgolly:
@sjrd says:

> I would suggest leaving `experimental` in the past, and represent instead the full spec, independently of browser support. Otherwise, there's never a good time nor a good way to have things "graduate" out of `experimental` anyway.

Yeah ok makes sense. So `experimental` goes in the bin and our API is just everything available, with the onus on users to know the availability status of the API they choose to call.

Semi-separately, I think it would be a good idea to have browser-specific packages for browser-specific API too. Eg. a `chrome` package with Chrome-only API. Thoughts?

armanbilge:
> Semi-separately, I think it would be a good idea to have browser-specific packages for browser-specific API too.

Interesting! What's an example of this?

japgolly:
Also, it would be good if we could somehow indicate browser support somehow though, and in a backwards-compatible way. Ideas:

* mention in ScalaDoc
* annotations
* encode in term/type names (eg. `@JSName("x") def x_experimental` which we then deprecate and replace over time if necessary)
* ???

None of these options seem amazing. I kind of like the annotations idea because

1. a quick look at the source is all that's needed, don't even have to really read unlike ScalaDoc
2. could be used be IDE plugins
3. super-low cost on maintenance

But is it even worth it? I weakly feel that it might be a good tradeoff. Thoughts?

armanbilge:
If I had to choose one, I'd say annotations as well. But it seems like a lot of work to keep the statuses up-to-date?

armanbilge:
Thinking about automation: could we query www.caniuse.com via an API to add this information programatically?

japgolly:
Hmmm, now that I've let this simmer a bit, I think it's not worth it. Mainly because we aren't getting paid to work on this (and I personally don't have any income lol!) so although it's a nice idea, I personally don't want to spend the time on it. You wanna take this on @armanbilge or shall we close it?

(This would be good to reopen if we end up generating everything from ts.)

armanbilge:
Agree, let's roll that idea into the mythical #487.

japgolly:
[...]

Issue #476, opened 2021-08-12
Split huge files into smaller units


japgolly:
Atm we have huge files like `lib.scala` with tons of stuff in them. It would be nice to split this into a more traditional structure.

armanbilge:
@japgolly Do you still want to do this?

japgolly:
@armanbilge Oh no way lol. I absolutely do not want to do this but I was resolved to because it'll be valuable and make future maintenance easier. If you want to do though, feel free! I would be overjoyed lol

armanbilge:
Lol! @sjrd also strongly recommended against in https://github.com/scala-js/scala-js-dom/issues/545#issuecomment-912187479. I think we should consider closing.

japgolly:
My understanding of that comment is that he was against doing both changes at the same time. `git blame` isn't a compelling reason to me anyway for not improving the codebase. We should do the best we can by git but the value of the codebase far outweights the value of git, especially on a project like this where it's 99% facades. Plus do keep in mind that `git blame` is already useless due to reformatting. If we really want to know the history, `git blame` isn't the best way and it can still be achieved via git.

armanbilge:
Ah, makes sense now. Ok, well I'll mark this help-wanted since basically anyone can do it.

sjrd:
> My understanding of that comment is that he was against doing both changes at the same time.

That's indeed what I was saying.

japgolly:
My understanding is that he was against doing both changes at the same
time. `git blame` isn't compelling cos we're moving everything around and
have reformatted it anyway.

japgolly:
Automation coming to save the day................. :eyes:

japgolly:
Hey @armanbilge just a an FYI, this is like 85% done but something came up, I got interrupted and it looks like I wont online for 4 days or so. Not even for reviewing your PRs unfortunately. Do with that info whatever you think is best, I'll be back next week, and sorry for any inconvenience :)

armanbilge:
No worries, have a nice long weekend I hope!

japgolly:
> looks like I wont online for 4 days or so

Ah yeah, sorry but I've had a bit of a medical emergency that blew out way beyond 4 days. It's going to take quite a while to resolve but slowly, slowly I'll starting coming back bit by bit. Sorry about the inconvenience and feel free to maintain away without my involvement. Don't want to hold anyone back.

armanbilge:
[...]

PR #523, opened 2021-08-30
Document coding standards and conventions

### LINKED ISSUE #519: Document coding standards and conventions
## Purpose:

* Consistency
* Keep current and new maintainers on the same (explicit) page
* Provide a doc for new contributors to read and adhere to
* Avoid needless back-and-forth between maintainers and contributors

## Content:

* package structure
* file organisation (#476)
* browser support (#514)
* non-facades (eg. implicit ops or conversions like `DOMNodeList => SeqLike`)
  * location
  * naming conventions
  * exports (eg. `@inline def css = raw.css`)
* facades
  * `def` vs `val`
  * null handling (i.e. just document that we don't put it in types atm, hey maybe annotations would be a good idea?)
  * when/where/how to use `js.Any`
* PRs
  * `prePR`
  * changelog? or is release-drafter enough? (probably right?)

japgolly:
Closes #519

Pending #514


japgolly:
@armanbilge @lihaoyi @sjrd Would you all mind having a review of this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+* Files
+* Facades
+* Non-Facades
+* Partially-Supported DOM API
+* Binary Compatibility
+* Submitting a PR
+
+
+Packages
+========
```
sjrd, line 16: I haven't made up a clear mind about this, but IMO the default should be to put everything inside `dom`, under their normal DOM name. The DOM is not namespaced itself, so it probably does not make sense to namespace things in scalajs-dom, especially for API subsets with a small surface.

Potential exceptions could be made for very large subspecifications. Currently, I can think of WebGL and Intl as the two most likely candidates for specific packages.

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+===========
+
+* Implicit conversions should go in companion objects so that they are always in scope without the
+  need for imports. There's no need to group by feature, the types already specify the feature.
+
+* Add Scala-only utilities that pertain to a specific facade, in the facades companion object
+  Eg: helper constructors, legal facade values.
+
+* Add Scala-only utilities that don't pertain to a specific facade, or shouldn't be universally
+  available (subjective judgement here) in the `ext` package.
```
sjrd, line 85: In my vision, the ideal outcome would be that we don't have anything left that belongs to this category, and hence that the `ext` package disappears. But that's only IMO.

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+Partially-Supported DOM API
+===========================
+
+TODO: Pending discussion in #514
+
+
+Binary Compatibility
+====================
+
+Binary compatibility for Scala.JS facades is different than standard Scala.
```
sjrd, line 97:
```suggestion
Binary compatibility for Scala.js facades is different than standard Scala.
```

CONTRIBUTING.md
armanbilge: > The DOM is not namespaced itself, so it probably does not make sense to namespace things in scalajs-dom

Thinking towards an auto-generated future, this is also good. Since generators wouldn't know about our namespaces without manual interventions.

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+Files
+=====
+
+* Use `package.scala` for a package object and nothing else.
+  Also don't include traits/classes/objects.
+
+* Match the filename to the trait/class/object in it; don't add multiple top-level types.
+  This is effectively Java-style.
```
armanbilge, line 56: 👍

CONTRIBUTING.md
armanbilge: Agree, but what to do with the things currently in there? 🤔

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+Submitting a PR
+===============
+
+Once you're done making your changes...
+
+1. Run `sbt prePR`
+2. Run `git diff api-reports` and ensure that you aren't breaking backwards-binary-compatibility
+   (see above). We'll probably automate this step in the future (See #503)
```
armanbilge, line 121: I think in the section above we should include some examples of what compatible/incompatible changes look like in the API reports. Specifically what all the `[ST]` and `[JT]` etc. means, I know they are documented at the top of the reports but nobody will see that when they check the diff.

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+* Match the filename to the trait/class/object in it; don't add multiple top-level types.
+  This is effectively Java-style.
+
+
+Facades
+=======
+
+If a feature has many types that don't share a common unambiguous prefix (like `crypto`),
+  * create a feature package
+  * put all of its types in its package
```
armanbilge, line 64: We should mention, if you are adding something this large/significant probably good to raise an issue for any necessary discussion before doing all this work!

CONTRIBUTING.md
sjrd: In my cleanup PR I am moving, deprecating, or outright removing everything in `ext`, so there's nothing left there if it gets accepted.

armanbilge: Apologies, it's been too long since I reviewed it and I was much less familiar with the library then.

japgolly: Yep now that #458 is merged everything in `ext` is deprecated. *chef's kiss*

CONTRIBUTING.md
japgolly: Noted. Committing to my long-term RAM...

CONTRIBUTING.md
japgolly: I've raised #545 to follow this thread.

japgolly:
Addressed PR feedback. Just need to sort out packages in #545

japgolly:
Ok this should be good to go now, @armanbilge can you please give this a final review and hopefully a merge :)

CONTRIBUTING.md
armanbilge: This is pending #476, unless we grandparent in all existing files but enforce for new files. Also, without namespaces this could get messy. Or not.

CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
+
+If a feature has only a few types that don't share a common unambiguous prefix (like `DOMParser`),
+  * create a facade for the main feature
+  * create a companion object for the facade
+  * put all of its types in the companion object
+
+If a feature has types that all share a common unambiguous prefix (like `Gamepad*`),
+  * put it all in `dom`
```
armanbilge, line 50: We can remove all of this now right? I assume we're not even doing the companion-object thing, because that's also a form of namespacing.

armanbilge: Oh, this is where the `def` / `val` thing can go.

CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
+* We currently don't see the need for Scala-only utilities that don't pertain to a specific facade,
+  or shouldn't be universally available (subjective judgement here).
+  If you believe you've got a compelling use case please raise an issue first to discuss.
+
+
+Partially-Supported DOM API
+===========================
+
+TODO: Pending discussion in #514
+
```
armanbilge, line 71: We can just drop this section. Let's add an intro section to this that says we accept PRs for anything in the spec.

CONTRIBUTING.md
armanbilge: Did I miss where you added this?

armanbilge (changes requested):
Sorry, this is a moving target with all the other changes/decisions going on 🎯 I'll write up a friendly intro to add to this :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
+Submitting a PR
+===============
+
+Once you're done making your changes...
+
+1. Run `sbt prePR`
+
+2. Run `git diff api-reports` and ensure that you aren't breaking backwards-binary-compatibility
+   (see above). The api reports are auto-generated by `prePR` and provide a concise summary of the
+   entire API. Make sure to look at the top of the file for a description of format.
```
armanbilge, line 102: Oh, I see this now. Hmm, I still think some examples would be helpful. Anyway, let's not block on this and I'll do it in a follow-up PR, sound good? :)

japgolly: Yeah great! Thank you!

CONTRIBUTING.md
japgolly: Yeah great idea. Can I leave this to you then in your follow up PR? Or else I can come back to this tomorrow

CONTRIBUTING.md
japgolly: Nah I reckon let's just state it now and we'll do #476 next. wdyt?

CONTRIBUTING.md
armanbilge: Oh yeah, that's fine I'll do it in the follow-up. If you address everything else we can merge!

CONTRIBUTING.md
armanbilge: Ok, so we're going ahead with #476? I wasn't sure because https://github.com/scala-js/scala-js-dom/issues/545#issuecomment-912187479.

armanbilge:
Add intro, api report ex, other fixes

armanbilge (approved):
@japgolly I pushed my changes here. Apologies, for expediency I'm going to merge pro-actively (hope it's okay) so that I can make a call for PRs and point to this guide. Please review when you next have a chance!

CONTRIBUTING.md
```diff
@@ -0,0 +1,133 @@
...
+```diff
+-experimental/Fetch[JO] def fetch(info: RequestInfo, init: RequestInit = null): js.Promise[Response]
++Fetch[JO] def fetch(info: RequestInfo, init: RequestInit = null): js.Promise[Response]
+```
+
+Anything annotated `[SC]`, `[ST]`, or `[SO]` in an API report is an ordinary Scala class/trait/object,
+for which standard binary compatibility rules apply.
+
+If the above doesn't make sense to you, don't worry!
+The majority of useful changes to scala-js-dom are indeed binary compatible.
```
armanbilge, line 116: I'm afraid this section just creates confusion esp. since most changes really are binary compatible anyway. IDK if we should put this burden on users. It's fine to have a section like "if you're interested" but asking that they understand this to make a PR might be intimidating.

japgolly:
Fantastic additions!! Very nice!!

PR #540, opened 2021-08-31
Add tests for {node + jsdom, chrome, firefox, webworkers}

### LINKED ISSUE #520: Add tests for various environments
## Why?

There are some things that are not immediately obvious to me that they'll work as I'm expected in various environments. I want automated tests for something, and I also want to ability to request new tests from contributors.

## Envs:

* browsers
* web workers
* node + jsdom
* node (without jsdom)

japgolly:
Closes #520

### DIFF
--- a/prePR.sbt
+++ b/prePR.sbt
@@ -13,5 +13,11 @@ ThisBuild / prePR_nonCross := Def.sequential(
     else
       Def.task[Unit]((dom / Compile / compile).value)
   },
-  root / Compile / compile,
+  root / Test / compile,
+  Def.taskDyn {
+    if (scalaVersion.value.startsWith("2.12."))
+      Def.task[Unit]((readme / Compile / compile).value)
+    else
+      Def.task(())
+  },
 ).value
--- a/prePR.sbt
+++ b/prePR.sbt
@@ -4,8 +4,21 @@ addCommandAlias("prePR", "+prePR_nonCross")
 
 val prePR_nonCross = taskKey[Unit]("Performs all necessary work required before submitting a PR, for a single version of Scala.")
 
+// Unfortunately we can't just call `root/Test/compile` because it doesn't take aggregation into account :(
 ThisBuild / prePR_nonCross := Def.sequential(
-  root / clean,
+
+  Def.task {
+    (root           / clean).value
+    (scalafixRules  / clean).value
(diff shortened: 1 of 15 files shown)


prePR.sbt
```diff
@@ -4,14 +4,43 @@ addCommandAlias("prePR", "+prePR_nonCross")
...
+    (testsNodeJsdom / Test / compile).value
+    (example        / Test / compile).value
+  },
+
+  Def.taskDyn {
+    if (scalaVersion.value.startsWith("2.12."))
+      Def.task[Unit]((readme / Compile / compile).value)
+    else
+      Def.task(())
+  },
```
armanbilge, line 44: 😊

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/AsyncTesting.scala
```diff
@@ -0,0 +1,37 @@
...
+import scala.scalajs.js.timers._
+
+object AsyncTesting {
+
+  type AsyncResult = Future[Try[Unit]]
+
+  implicit def global: ExecutionContext =
+    ExecutionContext.global
+
+  def async(run: => Future[Any]): AsyncResult = {
```
armanbilge, line 15: So, like please excuse my ignorance, but how do async tests work with Junit? How does the runner know to "wait" until they're completed?

armanbilge (approved):
Thanks for getting lift off on this, this is a really really important one.

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/AsyncTesting.scala
japgolly: No idea how/why it all works, just know that it does. I found out by looking at Future-related tests in scala-js itself and they have `type AsyncResult = Future[Try[Unit]]` and return `AsyncResult` all over the place and it Just Works so I followed the pattern over here too. Also tried `Future[Unit]` and it complained and failed the test.

armanbilge: Hmm, thanks, I'll try and look into this a bit myself to try and figure out how it works. But since it works and has precedence, let's go ahead with this!

PR #543, opened 2021-09-02
Remove hardcoded Scala versions from CI

japgolly:


.github/workflows/ci.yml
```diff
@@ -13,28 +13,32 @@ jobs:
...
 
       - uses: actions/checkout@v2
 
       - name: Setup Scala
-        uses: japgolly/setup-everything-scala@v1.0
+        uses: japgolly/setup-everything-scala@v2.0
 
       - name: Build and test
-        run: sbt -DCI=1 "++${{ matrix.scalaversion }}" test package doc
+        run: sbt++field ${{ matrix.scala }} -DCI=1 test package doc
```
armanbilge, line 29: What is `++field`?

armanbilge (commented):
This got a lot more confusing, it's great stuff but I'm worried if the build/ci is becoming too non-standard?

.github/workflows/ci.yml
```diff
@@ -13,28 +13,32 @@ jobs:
...
 
       - name: Validate formatting
-        run: sbt -DCI=1 "++${{ matrix.scalaversion }}" dom/scalafmtCheck
+        run: sbt++field ${{ matrix.scala }} -DCI=1 dom/scalafmtCheck
 
       - name: Validate api report
-        if: matrix.scalaversion != '2.11.12' && matrix.scalaversion != '3.0.1'
-        run: ./api-reports/validate "${{ matrix.scalaversion }}"
+        if: matrix.scala != 'scala211' && matrix.scala != 'scala3'
+        run: ./api-reports/validate $(sbt-const ${{ matrix.scala }})
```
armanbilge, line 36: What is `sbt-const`?

.github/workflows/ci.yml
japgolly: See https://github.com/japgolly/setup-everything-scala#dynamic-scala-versions

.github/workflows/ci.yml
japgolly: See https://github.com/japgolly/setup-scala-util

.github/workflows/ci.yml
armanbilge: Personal taste here, but I'd much rather use sbt-gh-actions for this. IMO there's nothing surprising about the generated workflows, and there is a 1-to-1 correspondence with the workflow DSL used to set it up in sbt itself.

armanbilge: > When Scala Steward proposes a Scala upgrade it will tested through CI, where as with sbt-gh-actions the github-actions check command fails immediately and a user has to manually upgrade Scala and regen all the workflows.

Nope, just works. Example: https://github.com/typelevel/cats-effect/pull/2290

> I also have to say as a non-user of sbt-gh-actions I'm plenty surprised by all the stuff if generates and I don't find reading it pleasant at all.

We can turn the defaults off, and literally just directly encode the current `ci.yml` using the GH actions DSL and fill in the blanks with the variables.

> It's no more or less custom than sbt-gh-actions

I disagree, because it produces a workflow than can be read by someone who knows nothing about sbt-gh-actions. Whereas someone reading your workflow has to know about all these custom features.

Again, maybe this is easier to discuss if you have a concrete example to compare against?

japgolly: > Nope, just works. Example: typelevel/cats-effect#2290

Oh wow! Scala Steward now does a separate `Regenerate workflow with sbt-github-actions` commit too. That's both amusing and amazing. Good on Scala Steward!

> We can turn the defaults off, and literally just directly encode the current ci.yml using the GH actions DSL and fill in the versions with the settings defined in the sbt build.

Yeah but for literally no benefit imo. As far as I can tell the only benefit is syncronising Scala versions and this PR achieves that without an unnecessary custom Scala layer. I presented a bunch of arguments last time we discussed. I don't really want to restate them all but none of them have been assuaged, they all still remain. (With the exception of the newer Scala Steward argument which turned out to be invalid.)

> I disagree, because it produces a workflow than can be read by someone who knows nothing about sbt-gh-actions. Whereas someone reading your workflow has to know about all these custom features.

Let's not exaggerate. `all these custom features` is literally two (and would usually be one) and there's a simple bit of doc that can be grokked in a few seconds. Where as from my perspective, there's a ton of stuff to have to learn for `sbt-gh-actions`, heaps of keys and custom, more verbose code to generate something so simple and small. To me `sbt-gh-actions` has a significantly higher learning curve  / pool of new custom concepts to learn, compared to what I'm proposing here. My goal is as little code as required whist maximising all the things that (imo) are valuable.

> Maybe this is easier to discuss if you have a concrete example to compare against?

I think scalajs-dom is the perfect concrete example :grin: 

Let's try something! I don't think either of us lack any new information, we may just have different values so how about we try to state each others and we can see if we're really understanding each other properly. So if I were to summarise your arguments and values, it's that

* avoiding duplication is good (i.e. Scala versions specified once)
* `sbt-gh-actions` is a big hit and lots of people will use it
* it's important that someone be able to look at and comprehend the yml (`githubWorkflowCheck` excepted I guess)
* the ability for someone to comprehend the scala used to generate the yaml isn't very important
* people don't want to learn how github yaml works
* the more config in scala the better
* `sbt-gh-actions` is easy to learn

Is that right? Not really as profound as I'd hoped so maybe I'm still missing some of your values / arguments?

japgolly: FYI: This is what I was referring to when I said Scala Steward and Scala upgrades don't work well together (until now): https://github.com/japgolly/test-state/pull/227. Moot for `sbt-gh-actions` users, but was a problem for everyone else.

armanbilge: > I think scalajs-dom is the perfect concrete example 😁

I think so too, I was proposing I'll make a parallel PR using sbt GH actions so we can have a look at what it gets right/wrong.

* _avoiding duplication is good (i.e. Scala versions specified once)_ Yes! 😁 I think this we both agree.
* _`sbt-gh-actions` is a big hit and lots of people will use it_ I would summarize this more as the bus factor on sbt-gh-actions is much better than the bus factor on your actions (which to be fair are brand new). I think also you really value uniformity here across all your projects which I totally get, and I am also seeking the same :)
* _it's important that someone be able to look at and comprehend the yml_ Also yes, also this we both agree I hope.
* _the ability for someone to comprehend the scala used to generate the yaml isn't very important_ I'm sad if you think I really believe that, this is important too. That's why I mentioned you can encode it as a 1-to-1 mapping of sbt-gh-action DSL to yaml. I do have the feeling you rate sbt complexity higher than GH action complexity maybe?
* _the more config in scala the better_ kind of. More like, the best place to access sbt settings is inside of sbt, because that's where they are used 99% of the time.
* _sbt-gh-actions is easy to learn_ maybe not easy to start from scratch, but I think if we set it up here people would feel comfortable modifying it due to its DSL closely matching the YAML. I see your point that yours is easy to learn by a quick look to the docs for the couple things (and all credit to your for fantastic docs and in general all your writing/organization skills), but I still feel this is one step further away. 

Another value that seems to differ between us are which tools we reach for to solve problems. For example the 2.10 hack script on series/1.x ... I don't think I'd ever be brave enough to do something like that. But to be fair, it did have somewhat unforseen consequences in at least a couple instances (I'd argue precisely because it was built outside of the standard scala build/release workflow). Not a criticism, it got us out of a very sticky situation and it was a temporary thing, but those are the kind of things that scare me tbh.

Also, not quite sure how to phrase this one, but maybe in a sense you are underestimating your overwhelming contribution to this project? What I mean by this is that the more amazing tools you contribute, things get much much dangerous in the bus factor sense (what would we ever do without you!!). So I know it may be suboptimal for you personally, but it may be optimal in the long-term in that this project can be more easily maintained without you. As an example, consider the scalatex thing the docs are written in that maybe made lots of sense at the time given the then-maintainers but now is maybe becoming cumbersome for you and me who don't really know what's going on there.

With all this said: I'm just a deputy and you have _much_ more experience in these things, so I defer to your judgement. I also think it would be perfectly reasonable to stick to the current (fairly vanilla) workflow with its caveats.

japgolly: Good morning! Ok well that idea of getting people in a disagreement to try to describe each others points has worked out really well for me in my personal life, but I'm actually pretty inexperienced at applying it to solving tech value problems when we're both working remotely. It didn't quite work out as well as I'd hoped but thanks to your responses I believe I understand where you're coming from and what you're thinking about more now, so in that sense it did provide value. It was a bit hard for me to accurately articulate so where you said *"I'm sad if you think I really believe that"*, yeah sorry pls don't take it the wrong way, I was more just trying to emphasise and exaggerate underlying values rather than be super-precise, I didn't mean to imply anything negative and no I didn't mean it literally :)

But ok cool, I understand you perspectives more now. If we were having this discussion over food or drinks or even a nice walk I'd happy respond back and I'm sure we'd get further, plus it would be pretty fun and interesting, but because this is over text and because our disagreement isn't high-impact, let's just leave this as they are, which is already a well-balanced compromise. Thanks for the discussion!

armanbilge: Morning! I see, apologies if I didn't quite grasp the concept 🤔 but I agree, unfortunately not so easy to discuss these things remotely. Always glad we discuss anyway and thanks for considering my opinions!

> If we were having this discussion over food or drinks or even a nice walk

Sounds nice 😀 Once upon a time I lived in your neck of the woods, I'm sure I'll find my way back someday 😆

japgolly: > Once upon a time I lived in your neck of the woods, I'm sure I'll find my way back someday

Oh wow, that's a fun coincidence! Definitely give me a shout if you're back around again!!

PR #548, opened 2021-09-04
Change auto-formatting settings

japgolly:
Part of #479. Not the whole thing and it's still pretty far from perfect, but it's an improvement :)


armanbilge (approved):
It is!! So, just confirming, from this point forward no more merges from series/1.x?

japgolly:
Yeah I don't think so, especially if we get 2.0.0 out soon

armanbilge:
That's great!! Yes let's keep pushing to 2.0 🚀

PR #554, opened 2021-09-05
Remove MDN from scaladocs

### LINKED ISSUE #553: Can we get rid of `MDN` in all the ScalaDoc?
It's annoying and useless IMO. Do we really need it? A link to the actual mdn page would be great, but just stating `MDN` seems pretty useless to me.

If it's for licencing issues,

1. we could just make a licencing statement upfront, once, about comments and MDN.
2. descriptions of API aren't copyrightable anyway. Even API itself isn't copyright-able according to Google vs Oracle.

armanbilge:
Fixes #553.

Also it looks like a bunch of other formatting snuck in somehow 🤔


src/main/scala/org/scalajs/dom/Fetch.scala
```diff
@@ -173,12 +173,12 @@ trait ResponseInit extends js.Object {
 @js.native
 trait Body extends js.Object {
 
-  /** MDN: Contains a Boolean that indicates whether the body has been read.
-    */
+  /*: Contains a Boolean that indicates whether the body has been read.
```
japgolly, line 176:
```suggestion
  /* Contains a Boolean that indicates whether the body has been read.
```

src/main/scala/org/scalajs/dom/Fetch.scala
```diff
@@ -173,12 +173,12 @@ trait ResponseInit extends js.Object {
...
+   */
   def bodyUsed: Boolean = js.native
 
-  /** MDN: Takes a Response stream and reads it to completion. It returns a promise that resolves with an ArrayBuffer.
-    */
+  /*: Takes a Response stream and reads it to completion. It returns a promise that resolves with an ArrayBuffer.
```
japgolly, line 180:
```suggestion
  /* Takes a Response stream and reads it to completion. It returns a promise that resolves with an ArrayBuffer.
```

src/main/scala/org/scalajs/dom/SVGTypes.scala
```diff
@@ -2756,38 +2194,30 @@ object SVGPreserveAspectRatio extends js.Object {
   /*
    * Corresponds to value meet for attribute preserveAspectRatio.
    *
-   * MDN
+
```
japgolly, line 2197: woah. Needs another `dom/scalafmt`

src/main/scala/org/scalajs/dom/lib.scala
```diff
@@ -2877,68 +2229,48 @@ abstract class Document extends Node with NodeSelector with DocumentEvent with P
   /** Creates an XPathNSResolver which resolves namespaces with respect to the definitions in scope for a specified
     * node.
     *
-    * MDN.
+    * .
```
japgolly, line 2232: remove

japgolly (changes requested):

japgolly:
Awesome work man!

armanbilge:
LOL it was just find replace really. Good catches, let me fix those.

src/main/scala/org/scalajs/dom/SVGTypes.scala
armanbilge: Ah, it's because the comment starts with `/*` 😡

japgolly: lol good catch!

japgolly (approved):

PR #557, opened 2021-09-05
Remove 1.x deprecatia

armanbilge:
Not sure if it's too soon for this since 1.2.0 is still fresh. But if not now, then when? :)

### DIFF
--- a/src/main/scala/org/scalajs/dom/lib.scala
+++ b/src/main/scala/org/scalajs/dom/lib.scala
@@ -5182,9 +5182,6 @@ object BlobPropertyBag {
 @JSGlobal
 class Blob(blobParts: js.Array[js.Any] = js.native, options: BlobPropertyBag = js.native) extends js.Object {
 
-  @deprecated("This method seems to have been added in error and not actually exist.", "1.2.0")
-  def close(): Unit = js.native
-
   /** The size, in bytes, of the data contained in the Blob object.
     */
   def size: Double = js.native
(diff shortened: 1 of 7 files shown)


japgolly:
Yeah idk, I was wondering about this too. It might be nicer for the migration to leave all in the 1.x deprecated stuff around in 2.x just so that it's really easy to upgrade. OTOH we could instruct everybody to upgrade to 1.2.0 first and make sure they aren't depending on anything deprecated, before upgrading. I don't think that would be unreasonable, but we could go either way. Option 2 (i.e. this PR) would be nicer for us....

armanbilge:
We can revisit this before we release or feel free to close!

armanbilge:
Also maybe we could do some of this. Like I think one of these is a method that was added in error so IDK how anyone could actually use that code, I'd love to purge it.

src/main/scala/org/scalajs/dom/lib.scala
```diff
@@ -5182,9 +5182,6 @@ object BlobPropertyBag {
 @JSGlobal
 class Blob(blobParts: js.Array[js.Any] = js.native, options: BlobPropertyBag = js.native) extends js.Object {
 
-  @deprecated("This method seems to have been added in error and not actually exist.", "1.2.0")
-  def close(): Unit = js.native
-
```
armanbilge, line 5187: Yep. This should definitely go IMO.

japgolly:
> We can revisit this before we release or feel free to close!

I'll take you up on the *feel free to close* part. Let's revisit later

PR #563, opened 2021-09-06
Revise all `postMessage` methods

### LINKED ISSUE #238: Worker.postMessage's second parameter type should be array of Transferables
As stated in https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage

Right now the parameter type is `js.Any` which of course works, but is not type safe.

The parameter type is correct in `DedicatedWorkerGlobalScope.postMessage` but I suppose it has not been updated to the `Worker` trait.

japgolly:
Fixes #238

### DIFF
--- a/api-reports/2_13.txt
+++ b/api-reports/2_13.txt
@@ -1195,7 +1195,7 @@ DedicatedWorkerGlobalScope[JT] var onmessage: js.Function1[MessageEvent, _]
 DedicatedWorkerGlobalScope[JT] var onoffline: js.Function1[Event, _]
 DedicatedWorkerGlobalScope[JT] var ononline: js.Function1[Event, _]
 DedicatedWorkerGlobalScope[JT] def origin: String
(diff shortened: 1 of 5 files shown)


api-reports/2_13.txt
```diff
@@ -13964,7 +13964,7 @@ MessagePort[JT] def addEventListener[T <: Event](`type`: String, listener: js.Fu
 MessagePort[JT] def close(): Unit
 MessagePort[JT] def dispatchEvent(evt: Event): Boolean
 MessagePort[JT] var onmessage: js.Function1[MessageEvent, _]
-MessagePort[JT] def postMessage(message: js.Any, ports: js.Any?): Unit
+MessagePort[JT] def postMessage(message: js.Any, transferList: js.UndefOr[js.Array[Transferable]]?): Unit
```
japgolly, line 13967: Deliberate. `ports` is incorrect. See https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/postMessage

armanbilge (approved):
Nice job with the docs.

Issue #545, opened 2021-09-03
Move packages into `dom`


japgolly:
Move the following into `dom`:

* `beacon`
* `crypto`
* `deviceorientation`
* `domparser`
* `gamepad`
* `mediastream`
* `permissions`
* `push`
* `serviceworkers`
* `sharedworkers`
* `storage`
* `webrtc`

This should leave us with the following packages:
* `intl`
* `webgl`

japgolly:
Decided to count LoC:
```sh
for f in \
  src/main/scala/org/scalajs/dom/crypto \
  src/main/scala/org/scalajs/dom/experimental/beacon \
  src/main/scala/org/scalajs/dom/experimental/cachestorage \
  src/main/scala/org/scalajs/dom/experimental/deviceorientation \
  src/main/scala/org/scalajs/dom/experimental/domparser \
  src/main/scala/org/scalajs/dom/experimental/gamepad \
  src/main/scala/org/scalajs/dom/experimental/intl \
  src/main/scala/org/scalajs/dom/experimental/mediastream \
  src/main/scala/org/scalajs/dom/experimental/permissions \
  src/main/scala/org/scalajs/dom/experimental/push \
  src/main/scala/org/scalajs/dom/experimental/serviceworkers \
  src/main/scala/org/scalajs/dom/experimental/sharedworkers \
  src/main/scala/org/scalajs/dom/experimental/storage \
  src/main/scala/org/scalajs/dom/experimental/webgl \
  src/main/scala/org/scalajs/dom/experimental/webrtc \
  ; do
  l="$(cloc $f | tail -2 | head -1 | awk '{print $5}')"
  echo "$f  -- $l"
done
```

which yielded unexpected results:

```
src/main/scala/org/scalajs/dom/crypto  -- 569
src/main/scala/org/scalajs/dom/experimental/beacon  -- 20
src/main/scala/org/scalajs/dom/experimental/cachestorage  -- 40
src/main/scala/org/scalajs/dom/experimental/deviceorientation  -- 50
src/main/scala/org/scalajs/dom/experimental/domparser  -- 21
src/main/scala/org/scalajs/dom/experimental/gamepad  -- 43
src/main/scala/org/scalajs/dom/experimental/intl  -- 161
src/main/scala/org/scalajs/dom/experimental/mediastream  -- 376
src/main/scala/org/scalajs/dom/experimental/permissions  -- 60
src/main/scala/org/scalajs/dom/experimental/push  -- 81
[...]

PR #564, opened 2021-09-06
Migrate beacon to dom

armanbilge:
Towards #545. Tried my hand at this but kind of tricky to do while maintaining source compatibility, let me know what you think, thanks!

### DIFF
--- a/src/main/scala/org/scalajs/dom/beacon.scala
+++ b/src/main/scala/org/scalajs/dom/beacon.scala
@@ -1,0 +1,52 @@
+package org.scalajs.dom
+
+import org.scalajs.dom
+import scala.scalajs.js
+
+/** The Beacon interface is used to schedule an asynchronous and non-blocking request to a web server. Beacon requests
+  * use the HTTP PUT method and requests typically do not require a response. Requests are guaranteed to be initiated
+  * before a page is unloaded and they are run to completion without requiring a blocking request (for example
+  * XMLHttpRequest).
+  *
+  * @see
+  *   [[https://www.w3.org/TR/2016/WD-beacon-20160204/ Beacon W3C Working Draft]]
+  * @see
+  *   [[https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API MDN Beacon API]]
+  */
+@js.native
+trait BeaconNavigator extends js.Object {
+
+  /** The navigator.sendBeacon() method can be used to asynchronously transfer small HTTP data from the User Agent to a
+    * web server.
+    *
+    * @param url
+    *   The url parameter indicates the resolved URL where the data is to be transmitted.
+    * @param data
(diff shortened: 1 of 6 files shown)


src/main/scala/org/scalajs/dom/lib.scala
```diff
@@ -317,6 +317,16 @@ class Navigator
   def clipboard: Clipboard = js.native
 }
 
+object Navigator {
+
+  implicit def toBeaconNavigator(n: Navigator): BeaconNavigator =
+    n.asInstanceOf[BeaconNavigator]
+
+  implicit def toBeaconWorkerNavigator(n: WorkerNavigator): BeaconWorkerNavigator =
```
japgolly, line 325: I think this might need to go in `object WorkerNavigator`

armanbilge: Oops, good catch! My bad.

japgolly (approved):
Looks great!!

armanbilge:
Thanks! Hopefully got the hang of it, going to try and slowly slog away at #545. Not looking forward to the big ones... (ahem crypto..)

japgolly:
Yeah I know, I appreciate the slog through tedium hehe and I'll be joining you soon I promise :laughing:

armanbilge:
Haha can't scalafix us out of this one eh? 😛

src/main/scala/org/scalajs/dom/beacon.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  /** The navigator.sendBeacon() method can be used to asynchronously transfer small HTTP data from the User Agent to a
+    * web server.
+    *
+    * @param url
+    *   The url parameter indicates the resolved URL where the data is to be transmitted.
+    * @param data
+    *   The data parameter is the ArrayBufferView, Blob, DOMString, or FormData data that is to be transmitted.
+    */
+  def sendBeacon(url: String, data: dom.BodyInit = null): Boolean = js.native
```
sjrd, line 27: It would be much simpler and cleaner to directly integrate this method in `Navigator` instead of using those implicit conversions. See for example https://github.com/scala-js/scala-js-dom/commit/d7933f9264b1ae5cd267baa7c18c0861edfaef8f where I did this for the Vibration API.

armanbilge: Good point! I guess the reason they were separated before was to quarantine the experimental methods?

sjrd: Yes, that was it. But now it doesn't make sense anymore, and it just obscures the API.

japgolly:
> Haha can't scalafix us out of this one eh? :stuck_out_tongue:

Lol I'm literally considering it hahahaha

PR #566, opened 2021-09-06
Migrate DOMParser to dom

armanbilge:
Towards #545.


src/main/scala/org/scalajs/dom/experimental/domparser/package.scala
```diff
@@ -0,0 +1,14 @@
...
+package object domparser {
+
+  @deprecated("use dom.DOMParser instead", "2.0.0")
+  type DOMParser = dom.DOMParser
+
+  @deprecated("use dom.MIMEType instead", "2.0.0")
+  type SupportedType = dom.MIMEType
+
+  @deprecated("use dom.MIMEType instead", "2.0.0")
+  val SupportedType = dom.MIMEType
```
armanbilge, line 13: Thoughts about renaming, to better match the spec? https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString

japgolly: Not sure what you mean (wrong link?) but `s/MIMEType/SupportedType` on this live and above

armanbilge: Huh? I meant that I renamed `SupportedType` to `MIMEType` since that's what they call it.

japgolly: Ah I'm an idiot, I was reading it the other way around.

japgolly (approved):

PR #568, opened 2021-09-06
Move Beacon API into Navigator

armanbilge:
As suggested in https://github.com/scala-js/scala-js-dom/pull/564#discussion_r702575875. Thanks @sjrd!


src/main/scala/org/scalajs/dom/WebWorkerTypes.scala
```diff
@@ -168,12 +167,10 @@ object DedicatedWorkerGlobalScope extends js.Object {
...
 @js.native
-trait WorkerNavigator extends NavigatorID with NavigatorOnLine with NavigatorLanguage
+trait WorkerNavigator extends NavigatorID with NavigatorOnLine with NavigatorLanguage {
 
-object WorkerNavigator {
-
-  implicit def toBeaconWorkerNavigator(n: WorkerNavigator): BeaconWorkerNavigator =
-    n.asInstanceOf[BeaconWorkerNavigator]
+  @deprecated("sendBeacon is not supported in web workers", "2.0.0")
+  def sendBeacon(url: String, data: BodyInit = js.native): Boolean = js.native
```
armanbilge, line 173: See https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API

> Note: This API is not available in Web Workers (not exposed via WorkerNavigator).

japgolly: Great catch!!!!

japgolly (approved):

PR #569, opened 2021-09-06
Improve precision of IDB API and add tests

japgolly:


japgolly:
wait up i'll rebase

japgolly:
Ok cool, rebased, tweaked, and ready for review

src/main/scala/org/scalajs/dom/IDBTypes.scala
```diff
@@ -10,6 +10,18 @@ import scala.scalajs.js
...
+class IDBEvent[+A](typeArg: String, init: js.UndefOr[EventInit] = js.undefined) extends Event(typeArg, init) {
+  override def target: IDBEventTarget[A] = js.native
+}
+
+@js.native
+@JSGlobal("EventTarget")
+class IDBEventTarget[+A] extends EventTarget {
+  def result: A = js.native
+}
+
```
armanbilge, line 24: Just a random thought, should `Event` and `EventTarget` themselves take type arguments?

src/main/scala/org/scalajs/dom/IDBTypes.scala
```diff
@@ -52,36 +64,51 @@ class IDBObjectStore extends js.Object {
     */
   def keyPath: String = js.native
 
-  def count(key: js.Any = js.native): IDBRequest = js.native
+  def count(key: Any = js.native): IDBRequest[Double] = js.native
```
armanbilge, line 67: Pedagogical question: when to use `js.Any` vs `Any`?

src/main/scala/org/scalajs/dom/IDBTypes.scala
```diff
@@ -52,36 +64,51 @@ class IDBObjectStore extends js.Object {
...
   /** To determine if the add operation has completed successfully, listen for the transaction’s complete event in
     * addition to the IDBObjectStore.add request’s success event, because the transaction may still fail after the
     * success event fires. In other words, the success event is only triggered when the transaction has been
     * successfully queued.
+    *
+    * @return
+    *   [[IDBRequest]] with the key as the `target` value
     */
-  def add(value: js.Any, key: js.Any = js.native): IDBRequest = js.native
+  def add(value: js.Any, key: Any = js.native): IDBRequest[Any] = js.native
```
armanbilge, line 77: I'm not really familiar with IDB so apologies if stupid: is there no way we can improve the type situation here? e.g.
```scala
def add[A](value: js.Any, key: A = js.native): IDBRequest[A] = js.native
```

armanbilge (commented):
This looks pretty awesome, adding type parameters to this API is obviously a big win. Most just questions since I'm not so familiar with IDB.

src/main/scala/org/scalajs/dom/IDBTypes.scala
```diff
@@ -488,15 +547,15 @@ class IDBFactory extends js.Object {
...
+class IDBRequest[A] extends EventTarget {
 
   /** The source of the request, such as an Index or a ObjectStore. If no source exists (such as when calling
     * IDBFactory.open), it returns null.
     */
   def source: js.Any = js.native
 
   /** The event handler for the success event. */
-  var onsuccess: js.Function1[Event, _] = js.native
+  var onsuccess: js.Function1[IDBEvent[A], _] = js.native
```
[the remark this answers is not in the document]
japgolly: Ahh yeah I'd say don't worry about that. The `this` is never new information in the IDB events, and as far as I know, not never in any of the standard dom api. We already have huge precedent all over this codebase to not type `this`.

src/main/scala/org/scalajs/dom/IDBTypes.scala
japgolly: I thought about that too but it would be a hugely disruptive change, and I don't want to take on that much work (not to mention that I can't guarantee that it wouldn't have adverse side effects that I'm not able to predict). How about we reconsider it again in future if it comes back up?

src/main/scala/org/scalajs/dom/IDBTypes.scala
japgolly: Good question and it would be nice to enforce and/or migrate to some consistent pattern. @sjrd would you mind sharing some advise here?

In terms of this specific change, I make all `IDBKey`s `Any` so that I can use `String` as a key directly without needing an ugly `.asInstanceOf[js.Any]`

armanbilge: Wait, there's no implicit conversion for string??

src/main/scala/org/scalajs/dom/IDBTypes.scala
japgolly: I actually made a branch to do just this and it worked very nicely! ...until I tried to compile it against Scala 3. There doesn't seem to be a way to manage deconsturction of unions via implicits in Scala 3 so I've parked it for now sadly. This reason it's not so straight forward to type IDB{Key,Value} is that they require recursive type aliases, ideally we'd just have `type IDBKey = String | ... | js.Array[IDBKey]` but there are some problems here: 1) js.Array isn't covariant, 2) trickery and implicits are required to have a self-recursive facade type, 3) implicits need to work with unions which is fine in Scala 2 but doesn't work in Scala 3 (yet)

src/main/scala/org/scalajs/dom/IDBTypes.scala
sjrd: The default should always be to use `Any`. Only use `js.Any` if you have a very good reason to use `js.Any`.

Use cases for typing things as `js.Any` are about as infrequent as those for `AnyVal`.

japgolly: Cool, I've replaced all of the unnecessary use of `js.Any` now.

src/main/scala/org/scalajs/dom/IDBTypes.scala
japgolly: I've added type aliases for `IDB{Key,KeyPath,Value}` now that just point to `Any`, but 1) clarify the API when reading, 2) allows a simple upgrade path for improving those types in future

japgolly:
Ok @armanbilge I've updated this again, and addressed all the feedback. You can read the expanded commit messages for the changes.

src/main/scala/org/scalajs/dom/IDBTypes.scala
armanbilge: Ahh, very clever!! Thanks for thanking a second pass at this, those Scala 3 issues seemed insurmountable.

armanbilge (approved):
Thanks for your massive efforts on this!! Learned a lot too.

japgolly:
Awesome! Thanks for the lengthy review too

PR #570, opened 2021-09-06
Fix api report validation

japgolly:
It stopped working because we stopped regenerating the api reports from CI.

Follow up to #567


armanbilge (approved):
Yikes! Thanks for looking into this.

PR #572, opened 2021-09-08
Migrate storage to dom

armanbilge:
Towards #545.

### DIFF
--- a/src/main/scala/org/scalajs/dom/experimental/storage/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/storage/package.scala
@@ -7,7 +7,7 @@ import scala.language.implicitConversions
 package object storage {
 
   @deprecated("use dom.Navigator directly instead", "2.0.0")
-  implicit def toNavigatorStorage(navigator: dom.Navigator): dom.Navigator =
+  def toNavigatorStorage(navigator: dom.Navigator): dom.Navigator =
     navigator
 
   @deprecated("use dom.Navigator directly instead", "2.0.0")
--- a/src/main/scala/org/scalajs/dom/experimental/storage/package.scala
+++ b/src/main/scala/org/scalajs/dom/experimental/storage/package.scala
@@ -7,24 +7,18 @@ import scala.scalajs.js
 /** https://storage.spec.whatwg.org/ */
 package object storage {
 
-  implicit def toNavigatorStorage(navigator: dom.Navigator): NavigatorStorage =
-    navigator.asInstanceOf[NavigatorStorage]
+  @deprecated("use dom.Navigator directly instead", "2.0.0")
+  implicit def toNavigatorStorage(navigator: dom.Navigator): dom.Navigator =
+    navigator
 
+  @deprecated("use dom.Navigator directly instead", "2.0.0")
   @js.native
-  trait NavigatorStorage extends js.Object {
-    val storage: StorageManager = js.native
-  }
+  trait NavigatorStorage extends dom.Navigator
 
-  @js.native
-  trait StorageManager extends js.Any {
-    def persisted(): js.Promise[Boolean] = js.native
-    def persist(): js.Promise[Boolean] = js.native
-    def estimate(): js.Promise[StorageEstimate] = js.native
-  }
+  @deprecated("use dom.StorageManager instead", "2.0.0")
+  type StorageManager = dom.StorageManager
 
-  @js.native
-  trait StorageEstimate extends js.Any {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/scalajs/dom/experimental/storage/package.scala
```diff
@@ -1,30 +1,22 @@
...
 import scala.language.implicitConversions
-import scala.scalajs.js
 
 /** https://storage.spec.whatwg.org/ */
 package object storage {
 
-  implicit def toNavigatorStorage(navigator: dom.Navigator): NavigatorStorage =
-    navigator.asInstanceOf[NavigatorStorage]
+  @deprecated("use dom.Navigator directly instead", "2.0.0")
+  implicit def toNavigatorStorage(navigator: dom.Navigator): dom.Navigator =
```
japgolly, line 10: This is `identity. Either it shouldn't exist at all, or if it's for compat, it shouldn't be implicit.

src/main/scala/org/scalajs/dom/storage.scala
```diff
@@ -0,0 +1,16 @@
...
+
+@js.native
+trait StorageManager extends js.Any {
+  def persisted(): js.Promise[Boolean] = js.native
+  def persist(): js.Promise[Boolean] = js.native
+  def estimate(): js.Promise[StorageEstimate] = js.native
+}
+
+@js.native
+trait StorageEstimate extends js.Any {
```
japgolly, line 13: This (and above) should extends `js.Object`, not `js.Any`

src/main/scala/org/scalajs/dom/lib.scala
```diff
@@ -325,6 +325,9 @@ class Navigator
     *   The data parameter is the ArrayBufferView, Blob, DOMString, or FormData data that is to be transmitted.
     */
   def sendBeacon(url: String, data: BodyInit = js.native): Boolean = js.native
+
+  /** https://storage.spec.whatwg.org/ */
+  val storage: StorageManager = js.native
```
japgolly, line 330: I had bad experience before with direct `val`s like this in facades when they aren't available in the target environment. Can you just please add a quick test to `SharedTests` to confirm that it works as expected in all environments, and if it isn't available in `JSDOMNode` can you pls make sure that another method of Navigator works as expected?

japgolly (changes requested):

src/main/scala/org/scalajs/dom/experimental/storage/package.scala
armanbilge: Yes, you're right, this is for compat. I think I need to fix this in a couple other places as well.

src/main/scala/org/scalajs/dom/storage.scala
armanbilge: I think I need to fix this in several places too, sorry. So in general I should prefer `js.Object` to `js.Any`?

src/main/scala/org/scalajs/dom/lib.scala
armanbilge: Hmm, maybe safer to just make a `def`? That's what we decided in the contributing guidelines. This caught my eye too, but I erred on the side of not changing the code I was migrating.

src/main/scala/org/scalajs/dom/experimental/storage/package.scala
armanbilge:
```suggestion
  def toNavigatorStorage(navigator: dom.Navigator): dom.Navigator =
```

src/main/scala/org/scalajs/dom/storage.scala
armanbilge: Ahh, silly me, I didn't write this facade, just migrated it 😅 I'll try and fix this when I encounter it.

japgolly: Yeah, with the facades have fields it should be `js.Object`. Normally I only extend `js.Any` for new enum types like `MediaStreamTrackState`

src/main/scala/org/scalajs/dom/lib.scala
japgolly: Yeah that's fine too

src/main/scala/org/scalajs/dom/storage.scala
armanbilge: That's an easy rule of thumb, thanks!

japgolly (approved):
Awesome! LGTM

PR #578, opened 2021-09-09
Migrate DeviceOrientation to dom

armanbilge:
Towards #545.

@japgolly apologies, I very stupidly pushed this commit directly to master at first (reverted now) ... I thought branch protections would prevent this?


japgolly:
LGTM.

Not sure about branch protections, I thought they were on too. I cbf atm, feel free to investigate :)

PR #579, opened 2021-09-10
Clean up config traits

sjrd:
And deprecate all the `def apply`-based constructors, in favor of the `new Foo { ... }` syntax.

### DIFF
--- a/api-reports/2_12.txt
+++ b/api-reports/2_12.txt
@@ -317,9 +317,9 @@ Blob[JC] def stream(): ReadableStream[Uint8Array]
 Blob[JC] def text(): js.Promise[String]
 Blob[JC] def `type`: String
 Blob[JO]
-BlobPropertyBag[JT] def endings: String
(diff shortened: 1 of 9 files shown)


armanbilge (approved):
This is really great (thanks for continuing to contribute!!) and very huge ... honestly I skimmed most of the changes. A lot of this code I'll be revisiting anyway as we de-namespace.

@japgolly seems to be on a bit of a hiatus atm but hopefully he'll be back soon :)

japgolly (approved):

japgolly:
Fantastic changes, thanks @sjrd . LGTM

armanbilge:
@sjrd do you have any more planned/suggested cleanups for 2.0? I'm trying to get an idea of remaining blockers. Thanks!

sjrd:
I'd like to have separate definitions of `sealed traits` for Scala 3 that would be `opaque type`s instead.

I think that's the only thing I have left.

PR #586, opened 2021-09-27
Restructure Java-style

### LINKED ISSUE #476: Split huge files into smaller units

japgolly:
Closes #476

Performed automatically using https://github.com/japgolly/scala-restructure

Changes:
* Move files into directories that match their packages
* Split and move files so that their filename matches the top-level type

Notice that the API reports have no modifications which is a good sign.


armanbilge:
Holy. Crap. Talk about coming back with a bang!! 😮 

Hope everything is well btw ❤️

src/main/scala/org/scalajs/dom/Location.scala
```diff
@@ -0,0 +1,73 @@
...
+  * and available under the Creative Commons Attribution-ShareAlike v2.5 or later.
+  * http://creativecommons.org/licenses/by-sa/2.5/
+  *
+  * Everything else is under the MIT License http://opensource.org/licenses/MIT
+  */
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** Vibrate the device for the specified number of milliseconds. */
```
japgolly, line 11: It's a very new tool. It still has bugs like this... hmmm......

japgolly: (This being ScalaDoc for a class that was moved, still hanging around.)

japgolly:
Note: this PR is more for feedback. I'll fix the issue I mentioned above and re-run this.

japgolly:
Ok I think this is ready to go now. I did a quick skim of some of the huge 44k patch and didn't notice anything wrong.

Also thanks for your previous comment @armanbilge

armanbilge (approved):
I skimmed through too and honestly LGTM. Like seriously major yak-shaving good. We can keep an eye out for any other weirdness but I think you nailed it 😄 

On a separate note, maybe we should add an auto-file-header plugin to the build.

PR #588, opened 2021-09-27
The great migration to `dom`

armanbilge:
Trying to wrap up #545.


armanbilge:
Phew 😅 that's the whole lot. I'd be shocked if this is 100% perfect, but if any issues do come up we should be able to fix them binary-compatibly.

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -63,6 +63,11 @@ package object dom {
   @deprecated("directly use the dom.CSS* types and values instead", "2.0.0")
   lazy val css: DeprecatedCSSAliases.type = DeprecatedCSSAliases
 
+  @js.native
+  @JSGlobal("crypto")
+  // `webCrypto` instead of just `crypto` to avoid conflict with deprecated package :(
+  def webCrypto: Crypto = js.native
```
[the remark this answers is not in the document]
sjrd: You can cheat:

1. Make `package object crypto` an `object crypto` instead.
2. Declare `subtle` and `getRandomValues()` as `@js.native val/def`s members of the `object crypto`, with `@JSGlobal("crypto.subtle")` and `@JSGlobal("crypto.getRandomValues")`.
3. Remove this `webCrypto`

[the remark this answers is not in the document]
armanbilge: Nice! Thanks!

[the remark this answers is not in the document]
armanbilge: @sjrd Hmm, I'm getting this warning:
```
Generated class org.scalajs.dom.crypto differs only in case from org.scalajs.dom.Crypto (defined in Crypto.scala).
  Such classes will overwrite one another on case-insensitive filesystems
```
Is this something to be concerned about and/or is there a workaround?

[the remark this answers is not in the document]
armanbilge: Maybe if we rename the object to e.g. `webcrypto` and then inside the `dom` package object we put `val crypto = webcrypto`?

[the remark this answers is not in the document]
sjrd: Hum... Remove `Crypto`? It's not like there's any use case for it anyway.

Or turn `crypto` back into a package object.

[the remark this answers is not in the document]
armanbilge: I think we still need `Crypto` for source compatibility. So back to a package object then.

src/main/scala/org/scalajs/dom/crypto/package.scala
```diff
@@ -1,33 +1,277 @@
...
 
+  /** Returns a SubtleCrypto object providing access to common cryptographic primitives, like hashing, signing,
+    * encryption or decryption.
+    */
+  @JSGlobal("crypto.subtle")
+  @js.native
+  val subtle: dom.SubtleCrypto = js.native
+
+  /** Fills the passed TypedArray with cryptographically sound random values. */
+  @JSGlobal("crypto.subtle")
```
sjrd, line 18:
```suggestion
  @JSGlobal("crypto.getRandomValues")
```

armanbilge: My bad, thanks!

src/main/scala/org/scalajs/dom/package.scala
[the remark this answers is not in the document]
sjrd: You don't need `Crypto` in `dom` for compatibility. You only need it in `crypto`. In `dom` you could either not have it at all, or have it under a different name, like `WebCrypto`. It's more important for `crypto` to have its original name because it is the common use case. `Crypto` is way less important.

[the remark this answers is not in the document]
armanbilge: Ah, you're right. Although there's an argument to be made for migration ("change imports from `dom.crypto.*` to `dom.*` and It Just Works").

Is there anything suboptimal about using a `package object`?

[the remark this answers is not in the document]
sjrd: Not per se. But it means that even in a later version where we can remove the deprecated things, we still won't be able to turn `crypto` into a regular native `val`.

To preserve imports, you could add a type alias
```scala
type Crypto = WebCrypto
```
as that won't generate a class file, and hence will not have any issue on case insensitive file systems.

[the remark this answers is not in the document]
armanbilge: > we still won't be able to turn `crypto` into a regular native `val`.

Wait ... if `crypto` is an vanilla (not-package) `object`, we can make it a `val` binary-compatibly? I thought that removing objects would cause problems.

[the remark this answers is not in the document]
sjrd: Not binary-compatibly, no. But source-compatibly, in a hypothetical scalajs-dom v3 one day.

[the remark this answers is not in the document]
armanbilge: Right, thanks. Then I guess I'm confused why a `package object` can't become a native `val` source-compatibly?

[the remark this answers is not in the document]
sjrd: It can, but you'll still be stuck by the `Crypto` trait.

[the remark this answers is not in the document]
armanbilge: I'm sorry, I've completely lost the thread of discussion here 😕 in summary:
1. We currently have `Crypto` and `package object crypto`,
2. In SJS-dom v3 we can source-compatibly make this `Crypto` and `val crypto: Crypto = js.native`.

What's the problem?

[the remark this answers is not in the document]
sjrd: Ah yes, you're completely right. I was confused.

So it's good as is. :)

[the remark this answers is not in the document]
armanbilge: No worries! Thanks for all your help 😁

api-reports/2_13.txt
```diff
@@ -271,8 +318,9 @@ AudioTrackList[JT] @JSBracketAccess def apply(index: Int): T
 AudioTrackList[JT] def dispatchEvent(evt: Event): Boolean
 AudioTrackList[JT] def getTrackById(id: String): AudioTrack
 AudioTrackList[JT] def length: Int
-AudioTrackList[JT] var onaddtrack: js.Function1[TrackEvent, _]
-AudioTrackList[JT] var onchange: js.Function1[js.Any, _]
+AudioTrackList[JT] var onaddtrack: js.Function1[Event, Any]
+AudioTrackList[JT] var onchange: js.Function1[Event, Any]
+AudioTrackList[JT] var onremovetrack: js.Function1[Event, Any]
```
japgolly, line 323: Can you confirm the above 5 lines? Types have changed and a new method appeared.

src/main/scala/org/scalajs/dom/MediaSource.scala
```diff
@@ -14,7 +13,7 @@ import scala.scalajs.js.annotation._
 class MediaSource extends EventTarget {
   def sourceBuffers: SourceBufferList = js.native
   def activeSourceBuffers: SourceBufferList = js.native
-  def readyState: ReadyState = js.native
+  def readyState: String = js.native
```
japgolly, line 16: Looks like a mistake

src/main/scala/org/scalajs/dom/MediaSource.scala
```diff
@@ -28,7 +27,7 @@ class MediaSource extends EventTarget {
 
   def addSourceBuffer(mimeType: String): SourceBuffer = js.native
   def removeSourceBuffer(sourceBuffer: SourceBuffer): Unit = js.native
-  def endOfStream(error: EndOfStreamError = js.native): Unit = js.native
+  def endOfStream(error: String = js.native): Unit = js.native
```
japgolly, line 30: Should be `EndOfStreamError` right?

src/main/scala/org/scalajs/dom/SourceBuffer.scala
```diff
@@ -1,12 +1,11 @@
...
-package org.scalajs.dom.experimental.mediastream
+package org.scalajs.dom
 
-import org.scalajs.dom._
 import scala.scalajs.js
 
 @js.native
 trait SourceBuffer extends EventTarget {
-  var mode: AppendMode = js.native
+  var mode: String = js.native
```
japgolly, line 8: another seeming regression

src/main/scala/org/scalajs/dom/TextTrack.scala
```diff
@@ -12,7 +12,7 @@ import scala.scalajs.js.annotation._
 @js.native
 trait TextTrack extends EventTarget {
   var language: String = js.native
-  var mode: js.Any = js.native
+  var mode: String = js.native
```
japgolly, line 15: beep boop

src/main/scala/org/scalajs/dom/TextTrackList.scala
```diff
@@ -9,4 +9,10 @@ package org.scalajs.dom
 import scala.scalajs.js
 
 @js.native
-trait TextTrackList extends DOMList[TextTrack]
+trait TextTrackList extends DOMList[TextTrack] {
+  def getTrackById(id: String): TextTrack = js.native
+
+  var onchange: js.Function1[Event, Any] = js.native
+  var onaddtrack: js.Function1[Event, Any] = js.native
+  var onremovetrack: js.Function1[Event, Any] = js.native
```
japgolly, line 17: These look like the same events I commented on above. Can you confirm the types?

src/main/scala/org/scalajs/dom/WebGLRenderingContext.scala
```diff
@@ -701,6 +701,21 @@ object WebGLRenderingContext {
   final val CONTEXT_LOST_WEBGL = 0x9242
   final val UNPACK_COLORSPACE_CONVERSION_WEBGL = 0x9243
   final val BROWSER_DEFAULT_WEBGL = 0x9244
+
+  implicit class WebGLRenderingContextOps(val webGL: WebGLRenderingContext) extends AnyVal {
```
japgolly, line 705: private val

src/main/scala/org/scalajs/dom/package.scala
```diff
@@ -14,7 +14,7 @@ package object dom {
     * property: it is merely a tag indicating objects that can be used in specific conditions, like to be transfered to
     * a Worker using the Worker.postMessage() method.
     */
-  type Transferable = ArrayBuffer | MessagePort
+  type Transferable = ArrayBuffer | MessagePort | CanvasProxy
```
japgolly, line 17: erm? Pls confirm? Doesn't seem very migratory...

src/main/scala/org/scalajs/dom/crypto/GlobalCrypto.scala
```diff
@@ -3,7 +3,7 @@ package org.scalajs.dom.crypto
 import scala.scalajs.js
 import scala.scalajs.js.annotation._
 
-@deprecated("use dom.crypto.crypto instead", "2.0.0")
+@deprecated("use dom.crypto instead", "2.0.0")
```
japgolly, line 6: This doesn't seem right. `dom.crypto` is now is the package object, where as this deprecated `dom.crypto.crypto` method returns an instance of `Crypto` which extends `js.Object`. 

Also because `crypto` isn't always available, shouldn't `dom.crypto` be an instance of `js.UndefOf[Crypto]`?

japgolly (changes requested):
Few comments but fantastic work getting this done @armanbilge ! It would've involved quite a bit of tedium.

api-reports/2_13.txt
armanbilge: That's because I merged duplicated facades.
https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/AudioTrackList.scala
https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/experimental/mediastream/AudioTrackList.scala

src/main/scala/org/scalajs/dom/MediaSource.scala
armanbilge: These were just straight aliases i.e. `type ReadyState = String` so I think I dropped these.
https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/experimental/mediastream/package.scala

What should happen, is these should be modeled as `sealed trait`s on Scala 2 and `opaque type` on Scala 3. Can that be a followup PR?

src/main/scala/org/scalajs/dom/MediaSource.scala
armanbilge: Same as above.

src/main/scala/org/scalajs/dom/SourceBuffer.scala
armanbilge: Same as above (thanks for the careful eye btw :)

src/main/scala/org/scalajs/dom/TextTrack.scala
armanbilge: same

src/main/scala/org/scalajs/dom/WebGLRenderingContext.scala
armanbilge:
```suggestion
  implicit class WebGLRenderingContextOps(private val webGL: WebGLRenderingContext) extends AnyVal {
```

src/main/scala/org/scalajs/dom/package.scala
armanbilge: I merged what was already there with
https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/experimental/serviceworkers/package.scala

src/main/scala/org/scalajs/dom/crypto/GlobalCrypto.scala
armanbilge: Yes, but to keep source compatibility we are in a bit of a hole. See my back-and-forth with @sjrd above.
https://github.com/scala-js/scala-js-dom/pull/588#discussion_r717154719

armanbilge: In other words, this is source compatible. But, for anyone that specifically wanted a `Crypto` object (which I assume is rare) you're right this wouldn't work.

src/main/scala/org/scalajs/dom/MediaSource.scala
armanbilge: I guess I should fix that in this PR :)

armanbilge: Ah, I remember my concern. Since these are not officially declared types (are they?) I was concerned about namespace clashes. We need to be careful how we name them.

armanbilge: Compare with `BufferSource` which is officially declared: https://developer.mozilla.org/en-US/docs/Web/API/BufferSource

I guess what we should do is keep these for source compatibility, but deprecate them?

src/main/scala/org/scalajs/dom/TextTrack.scala
armanbilge: Fixed! :)

src/main/scala/org/scalajs/dom/package.scala
armanbilge: I can't find this page in English, but it confirms this.
https://developer.mozilla.org/de/docs/Web/API/Transferable

src/main/scala/org/scalajs/dom/MediaSource.scala
japgolly: Nice re concept. That's a very good point you raise about namespace pollution. Following Typescript's lead is a great idea.

src/main/scala/org/scalajs/dom/crypto/GlobalCrypto.scala
japgolly: > But, for anyone that specifically wanted a Crypto object (which I assume is rare) you're right this wouldn't work.

I rely on a `Crypto` object in my code :) Let me take a stab at a solution...

japgolly (changes requested):
LGTM! Just two minor changes, and the `crypto` problem. Maybe we could merge with `crypto` as is and then I take a shot at it in a follow up PR?

armanbilge:
Thanks again for your careful eye! Fixed those `TrackEvent`s that slipped through and should be good to go when green.

armanbilge:
Actually this was made SO much easier after you split all the files! So much credit to you as well that was a life saver!!

Issue #481, opened 2021-08-13
Null handling


japgolly:
What do we do in cases where a type in facade is nullable? We have `js.UndefOr` but sadly there's no `js.NullOr`. In scalajs-react I add `| Null` to facade types but the problem is it's non-trivial to get rid of the null (outside of Scala 3's experimental explicit-nulls), and I've had to add [helpers with casts](https://github.com/japgolly/scalajs-react/blob/master/util/src/main/scala/japgolly/scalajs/react/util/JsUtil.scala#L93-L97).

I absolutely hate the idea of just saying `AudioTrack` and the user having to somehow know that it's nullable. I want more clarity for users: types would be the best, information is probably the fallback (eg. annotations, worse: doc).

What do to? It might be an idea to add something like `NullOr` to scala-js-dom itself.

armanbilge:
Yes, I ran into this too when reviewing a PR. The idiom here so far seems to be _not_ to indicate the possibly `null` type. OTOH, so far this library has essentially been a pure facade and adding opinionated nice-ities is arguably out of scope?

armanbilge:
Btw, https://github.com/scala-js/scala-js-dom/issues/414#issuecomment-646650961 made me think critically about what the purpose of this library is. I don't mean to open a can-of-worms over `null` but it is helpful if we clarify the goals/scope of scala-js-dom going forward.

japgolly:
> this library has essentially been a pure facade and adding opinionated nice-ities is arguably out of scope

I agree with that sentence but I don't agree that clarifying nulls in types goes beyond that scope. In the same way the `js.UndefOr` is a facade over `A | undefined` (with a few helper methods for nice Scala usage) I think a similar concept over null is just as in scope. It adds precision to the facade types.

Plus I'd also argue that in addition to simply being pure facades, we also provide utilities for convenient Scala usage, and there's already precedent for that. (Check out the `[S_]` tagged items in the api report.) As an example, one of the most annoying things ever, is getting a `NodeList` back from a facade, and not being able to iterate over it via typical Scala collections. I think I saw somewhere recently that this is fixed in our pending PR queue? But regardless, I'd argue we've got to go beyond providing the most minimal use. I think we should hard-limit our scope to JS DOM and browser API, but within that scope I think we should make it as nicely usable as possible, and as safe/precise as possible.

japgolly:
[...]

PR #593, opened 2021-10-11
Improve `HTMLCollection`, add `HTMLOptionsCollection` and `HTMLFormControlsCollection`

ghostbuster91:
Hi,

Following suggestion from @raquo here is a PR which fixes incorrect type of `options` property in `HTMLSelectElement`.

Cheers!


src/main/scala/org/scalajs/dom/HTMLCollection.scala
```diff
@@ -14,7 +14,7 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-class HTMLCollection private[this] () extends DOMList[Element] {
+abstract class HTMLCollection extends DOMList[Element] {
```
sjrd, line 17: Consider keeping the constructor package-private, as I don't think it's possible to extend this class in user code:
```suggestion
abstract class HTMLCollection private[dom] extends DOMList[Element] {
```

armanbilge: Also, can we use a type parameter? e.g. 
```suggestion
abstract class HTMLCollection[E <: Element] extends DOMList[E] {
```

This way, you wouldn't have to override `item` and `namedItem` to get more specific return types.

src/main/scala/org/scalajs/dom/HTMLFormElement.scala
```diff
@@ -33,7 +33,7 @@ abstract class HTMLFormElement extends HTMLElement {
   /** elements returns an HTMLFormControlsCollection (HTML 4 HTMLCollection) of all the form controls contained in the
     * FORM element, with the exception of input elements which have a type attribute of image.
     */
-  def elements: HTMLCollection = js.native
+  def elements: HTMLCollection[Element] = js.native
```
ghostbuster91, line 36: I noticed that this is a quite similar situation to the one with `HtmlOptionsCollection` since according to the docs(https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormControlsCollection) this one should also extend `HtmlCollection`. On the other hand there is already a type `HTMLFormControlsCollection` which comes from typescript bindings(?). So I am not sure how this one should be approached. I guess that there is nothing wrong with leaving it as it is right now as it can be addressed in a separate PR if needed.

ghostbuster91: Sure, no problem :+1:

ghostbuster91: Hmm, I am not sure how to do it. 

```
Returns the RadioNodeList or the Element in the collection whose name or id matches the specified name, 
or null if no nodes match
```
Do we have a way to model it? Should I parameterize the collection using `Either[RadioNodeList, Element]`? I doubt that it will work that way without any additional code. 

The corresponding type in typescript looks as follow:

```
/** A collection of HTML form control elements.  */
interface HTMLFormControlsCollection extends HTMLCollectionBase {
    /**
     * Returns the item with ID or name name from the collection.
     *
     * If there are multiple matching items, then a RadioNodeList object containing 
     * all those elements is returned.
     */
    namedItem(name: string): RadioNodeList | Element | null;
}
```

armanbilge: Good idea to check the typescript! All you need is `import scala.scalajs.js.|` and you should be able to use a return signature essentially the same as the typescript (you can omit the null case).

yurique: Maybe use the `|`, too?

```scala
type RadioNodeListOrElement = RadioNodeList | Element | Null
```

and provide some extension methods (`asElement: Option[Element]` or `asElement: UndefOr[Element]`, etc) and/or extractor (`unapply(...): Option[Either[RadioNodeList, Element]]`) for it?

armanbilge: Thanks for chiming in! 👍 for `|`, as for the extension methods we leave those to other libraries, the goal here is to write a raw facade :)

armanbilge: @yurique not just you! I've wondered the same thing myself, see e.g. https://github.com/scala-js/scala-js-dom/issues/481#issuecomment-898344699 regarding null handling. And it's a fuzzy line for sure, for example [`DOMList`](https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/DOMList.scala) has implicit conversions to `Seq`, but I think this is one of the few exceptions /shrug.

src/main/scala/org/scalajs/dom/html.scala
```diff
@@ -10,7 +10,7 @@ object html {
   type Button = HTMLButtonElement
   type BR = HTMLBRElement
   type Canvas = HTMLCanvasElement
-  type Collection = HTMLCollection
+  type Collection[E <: Element] = HTMLCollection[E]
```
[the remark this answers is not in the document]
ghostbuster91: Yes, my mistake. I forgot to remove it from here.

ghostbuster91:
> Otherwise, looks great!! Thank you so much for taking this on and sticking with it, much appreciated. This is a significant improvement to this API, and just in time for 2.0.0 grin

No problem at all. I learned some new things and I am always happy to contribute back to scala ecosystem :) 
Thanks for the guidance :+1:

src/main/scala/org/scalajs/dom/raw.scala
```diff
@@ -320,8 +320,8 @@ object raw {
   @deprecated("use dom.HTMLCanvasElement instead", "2.0.0")
   type HTMLCanvasElement = dom.HTMLCanvasElement
 
-  @deprecated("use dom.HTMLCollectionElement instead", "2.0.0")
-  type HTMLCollectionElement = dom.HTMLCollection
+  @deprecated("use dom.HTMLCollection instead", "2.0.0")
+  type HTMLCollectionElement[E] = dom.HTMLCollection[E]
```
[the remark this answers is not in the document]
ghostbuster91: Ah, yes, you are right. I guess I should drink some coffee to wake up properly :)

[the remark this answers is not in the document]
armanbilge: No worries! This is just annoying maintainer stuff ☕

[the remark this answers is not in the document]
ghostbuster91: Hmm I guess that after that change I should rerun `sbt prePR`, right?

armanbilge:
@japgolly do you want to have a look at this?

src/main/scala/org/scalajs/dom/HTMLCollection.scala
```diff
@@ -14,12 +14,12 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-class HTMLCollection private[this] () extends DOMList[Element] {
-  def item(index: Int): Element = js.native
+abstract class HTMLCollection[E] extends DOMList[E] {
```
japgolly, line 17: Should this be `HTMLCollection[+E <: Node] extends DOMList[E]`?

Note: If yes, don't forget to add a `+` to the type alias too :)

armanbilge: I'm interpreting the spec as any object? Not just a `Node`?
https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection
> The HTMLCollection interface represents a generic collection (array-like object similar to arguments) of elements (in document order) and offers methods and properties for selecting from the list.

armanbilge: The `+` makes sense though 👍

PR #607, opened 2021-10-18
Un-deprecate AJAX

### LINKED ISSUE #606: `XMLHttpRequest` and friends shouldn't be deprecated
`XMLHttpRequest` and `AjaxException` are deprecated in scalajs-dom but not irl.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest

japgolly:
Closes #606


japgolly:
Oh right, I completely missed that. Well let's just leave it in there to be kind. There's already code out in the wild using it, plus the fact that AJAX isn't deprecated by browsers means there's no reason to deprecate our little helpers around it. Also it never changes so there's no maintenance cost. If browsers deprecated AJAX then I'd be onboard with deprecating it, but there seems to be no reason to remove it either (now that it's in).

A users perspective: I'm using it in scalajs-react and now I'm getting warnings that I should use the Fetch API. Like wtf, no, I'll use what I want. I don't *have* to move to Fetch and it's overreach for scalajs-dom to tell me to do so.

armanbilge:
That's fair. IMHO it should still be deprecated, at the very least to prevent new uses (since we don't offer a similar `Future`-based wrapper API for `fetch`, this one actually looks friendlier/more accessible!). But I'd be willing to compromise at "we'll keep this to the end of time for compatibility" or whatever 😆

japgolly:
I see what you mean but I don't agree. It's not our job to teach people improvements for their usage for DOM, it's our job to provide DOM facades. Like it would be nice to have in comments, but the reason I disagree with deprecation because it penalises use, and sometimes there are good reasons to use it. There's actually no problem at all using so it feels to me like we'd be forcing our (non-nuanced) opinions on downstream users.

armanbilge:
> it's our job to provide DOM facades

That's the thing, it's not a facade. If it was, I definitely would agree that it should not be deprecated. Maybe two sides of the same coin, but the wrapper API provided here is an opinion!

armanbilge:
> my problem is that it creates problems for users who've already used it. It doesn't seem right to me to force them to make unnecessary changes.

Unless I'm missing something, this doesn't create any more problems than all the other deprecations we introduced? Maybe the problem here was the message suggesting to use `Fetch`?

armanbilge:
> The deprecation warnings are the problem.

What about the million other deprecation warnings we introduced? 😆 is it just because those are much easier to resolve?

> The motivation to deprecate this is... Fetch API is nicer?

Another motivation is that we're stepping away from `ext` and just focusing on pure facades. But maybe that's more an ideological point.

Urgh, frustrating but in the end I do agree with you. It's like how we are staying on SJS 1.5 instead of latest and greatest ... technically, there's no reason we _have_ to, but it's about being kind considering our position at the bottom of the hiearchy.

japgolly:
> I thought that's what we did. All of ext is deprecated, what's left?

Oh I assumed there'd be more stuff. Damn that changes everything, I was thinking ext/Ajax was an outlier....

> Yeah, agree :heart:

What would you think about deferring it's deprecation until 3.x then?

armanbilge:
Bugger, really? Not even in Scaladoc? But it's documented?

japgolly:
idk, this is the source of truth https://github.com/scala/scala/blob/v2.13.6/src/library/scala/annotation/migration.scala#L30

armanbilge:
What if the message says "This extension/helper/whatever is deprecated, please use `dom.XMLHttpRequest` directly instead" ?

armanbilge:
😆 here you are, blessing `ext` away!

> Yep now that #458 is merged everything in ext is deprecated. chef's kiss

https://github.com/scala-js/scala-js-dom/pull/523#discussion_r701504369

armanbilge:
It's really no big deal, don't sweat it. At the time I don't think we appreciate the consequences (we were happily ticking those maintainer boxes as you say 😉) particularly to your own libraries. So I think it's fair game to re-evaluate our position here, especially since this is a deprecation-without-replacement. I think a milder deprecation via scaladoc is reasonable.

PR #609, opened 2021-10-18
Change `raw` back into a package object

### LINKED ISSUE #608: package org.scalajs.dom contains object and package with same name: raw
Only occurring with Scala 3:

```
exception caught when loading module class dom: package org.scalajs.dom contains object and package with same name: raw
one of them needs to be removed from classpath
```

https://github.com/japgolly/scalajs-react/pull/995/checks?check_run_id=3921789951#step:4:7310

japgolly:
Closes #608

What do you think about giving this a try for RC2?


armanbilge (approved):
No idea what's going on here, but yes this is very important and good we caught it in the RC stage.

armanbilge:
Also, we should have a careful look through the experimental package, I'm afraid it might have similar landmines like this?

japgolly:
Good point!

```
> rgs '^object +[a-z]'
src/main/scala/org/scalajs/dom/svg.scala
6:object svg {

src/main/scala/org/scalajs/dom/idb.scala
4:object idb {

src/main/scala/org/scalajs/dom/html.scala
4:object html {

src/main/scala/org/scalajs/dom/webworkers.scala
8:object webworkers {
```

armanbilge:
The problem is, I don't understand what should be an object or a package object. Some of those were already objects I think? Maybe not?

armanbilge:
Yeah, I'd like to take a little more time to try and understand the underlying problem first, if you don't mind. Maybe @sjrd can help as well.

If you want, let's keep this PR scoped to just changing `raw` back to a package object, since that's how it used to be.

PR #617, opened 2021-10-21
Revise and add `ReadyState` classes

### LINKED ISSUE #615: `ReadyState` is ambiguous
We have `ReadyState` as `closed | ended | open` which is currently being used in `MediaSource`.

The problem is that there are many `ReadyState` definitions depending on where it's used. For example, [`IDBRequest.readyState`](https://developer.mozilla.org/en-US/docs/Web/API/IDBRequest/readyState) is `pending | done`.

We should add types for all (or at least some common) ready state enums, and prefix them all according to their domain.

```
> rgs ' readyState: '
dom/src/main/scala/org/scalajs/dom/MediaSource.scala
16:  def readyState: ReadyState = js.native

dom/src/main/scala/org/scalajs/dom/RTCDataChannel.scala
33:  def readyState: RTCDataChannelState = js.native

dom/src/main/scala/org/scalajs/dom/IDBRequest.scala
52:  def readyState: String = js.native

dom/src/main/scala/org/scalajs/dom/MediaStreamTrack.scala
44:  val readyState: MediaStreamTrackState = js.native
110:      readyState: js.UndefOr[String] = js.undefined, remote: js.UndefOr[Boolean] = js.undefined,

dom/src/main/scala/org/scalajs/dom/HTMLDocument.scala
50:  def readyState: String = js.native

dom/src/main/scala/org/scalajs/dom/EventSource.scala
36:  def readyState: Int = js.native

dom/src/main/scala/org/scalajs/dom/FileReader.scala
26:  def readyState: Short = js.native

dom/src/main/scala/org/scalajs/dom/HTMLElement.scala
65:  var readyState: js.Any = js.native

dom/src/main/scala/org/scalajs/dom/TextTrack.scala
16:  var readyState: Int = js.native

dom/src/main/scala/org/scalajs/dom/WebSocket.scala
28:  def readyState: Int = js.native

dom/src/main/scala/org/scalajs/dom/XMLHttpRequest.scala
33:  def readyState: Int = js.native
```

japgolly:
Closes #615


armanbilge (approved):
Thanks for tracking all these down, this is definitely the right approach 👍

Issue #624, opened 2021-11-09
Consistently set defaults for `new`-based constructors


yurique:
`MutationObserverInit` is defined as a trait with vars and initial values, thus doing this won't compile:

```scala
new MutationObserverInit {
  val childList: UndefOr[Boolean] = someChildList
  ...
```

As a slight inconvenience, because the fields are initialized in the trait, the `override` modifier is required. 
But what doesn't work is overriding the vars: `scalac` complains `mutable variable cannot be overridden`.

sjrd:
This is intentional. The intended way to use it is as follows:
```scala
new MutationObserverInit {
  childList = someChildList
}
```

yurique:
Ahh, indeed! My bad! Feeling kinda stupid now =) 
Thank you, Sébastien!

yurique:
I think this was what had thrown me off:
```scala
// Before (deprecated)
val response = dom.experimental.ResponseInit(myStatus, myStatusText, myHeaders)

// After
val response = new dom.ResponseInit {
  var status = myStatus
  var statusText = myStatusText
  var headers = myHeaders
}
```

armanbilge:
Hmm, is the use of var not needed there? I grabbed that example from one of my own projects:
https://github.com/http4s/http4s-dom/blob/b900066aec70b39286925684b9501e66b36c11fe/dom/src/main/scala/org/http4s/dom/ServiceWorker.scala#L92-L95

yurique:
@armanbilge Here's how I understand it :)

`ResponseInit` is defined like this (vars are abstract):
```scala

trait ResponseInit extends js.Object {
  var status: Int
  var statusText: ByteString
  var headers: HeadersInit
}
```

and this compiles (overriding-implementing the vars from the trait):
```scala
  val response = new org.scalajs.dom.ResponseInit {
    var status     = 200
    var statusText = "OK"
    var headers    = scala.scalajs.js.Dictionary.empty[String]
  }
``` 
without putting `var`s there it won't compile because those need to be overriden (implemented).

By contrast, `MutationObserverInit` is defined like this (there *are* default implementations for the vars):
```scala
trait MutationObserverInit extends js.Object {
  var childList: js.UndefOr[Boolean] = js.undefined
...
```

[...]

Issue #622, opened 2021-11-01
Fetch request body as ReadableStream


ptrdom:
According to the specs, Fetch should be able to handle `ReadableStream` as request body. Implementing this would likely result in some changes to `ReadableStream` and `ReadableStreamController` facades - `ReadableStream` needs a constructor that implements the `underlyingSource`, which allows interaction with `ReadableStreamController`, and `ReadableStreamController` should not have its own constructor, as it is constructed by `ReadableStream` during its construction.

If implementing this would be approved, I could give it a shot.

References:
- https://developer.mozilla.org/en-US/docs/Web/API/fetch#parameters
- https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream
- https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController

armanbilge:
Thanks for this! 👍 to these changes, the goal is to match the spec so a contribution for this would be _very_ welcome 😁 

_However_, just so you are aware ... I recently implemented a `Fetch`-based client as well in [`http4s-dom`](https://github.com/http4s/http4s-dom) and tried using a `ReadableStream` in the request body, but was dissapointed to learn that many browsers have not implemented this feature for requests. See:

https://caniuse.com/mdn-api_request_request_readablestream_request_body

ptrdom:
I have just realized that also, while playing around with possible new version of the facade 😞. I guess it still makes sense to have the upgraded facade in case someone might find some use for natively available streaming implementation and to prepare for eventual browser support for `ReadableStream` in `Fetch` request body.

I will start working on the PR then.

armanbilge:
Thanks! The enhancements for `ReadableStream` in general are a good idea, and even though browsers have not implemented it, there are other environments that have, such as Cloudflare Workers (serverless):
https://developers.cloudflare.com/workers/runtime-apis/request#requestinit

armanbilge:
Good news, it seems that Chrome will finally be supporting this:
https://developer.chrome.com/articles/fetch-streaming-requests/

Hopefully other browsers will follow suit.

Issue #503, opened 2021-08-25
Auto-generate MiMa exclusions via scalafix


armanbilge:
As proposed in https://github.com/scala-js/scala-js-dom/pull/380#issuecomment-905130204.

The idea is to use @japgolly's API report machinery to automatically generate exclusions for known binary-compatible changes for facades as delineated in https://github.com/scala-js/scala-js-dom/pull/461#issuecomment-894829615. This would enable us to use MiMa without the headache of an overwhelming number of false-positives. Furthermore, I think this would be a significant step towards an auto-generation scheme as described in https://github.com/scala-js/scala-js-dom/discussions/487#discussioncomment-1185706.

I know MiMa exclusions can be specified via files (this is how Akka projects manage theirs) and also in the SBT build. The former strategy is very compatible with the API-report generation scheme, but requires an extra `prePR` step and CI overhead to keep this in sync with the source. If it's not too ambitious, I wonder if during the MiMa check the exclusions can be generated programmatically instead of read from a (programatically-generated) file. If it works, that would be a significant simplification and remove a burden from both maintenance and contribution.

armanbilge:
I was daydreaming about this and realized there was some subtlety I missed when I first wrote this up. Unlike API reports, which we want re-generated for every PR so we can compare the diff, we actually do _not_ want to re-generate MiMa exclusions because these should be based on the last stable version. Generating exclusions based on the current code is self-serving and would let you sneak all kinds of bad stuff in ;)

With this in mind: MiMa exclusions should definitely be generated into files (like the API reports) and my idea above about dynamically loading the exclusions should be tossed. Furthermore, re-generating these exclusions should not happen in a `prePR` step, but only after releasing a new version of scala-js-dom.

PR #628, opened 2021-11-10
Fetch request body as ReadableStream

ptrdom:
References https://github.com/scala-js/scala-js-dom/issues/622.


ptrdom:
I was unable to run `sbt prePR`, received this error:
```
[error] /scala-js-dom/readme/Index.scalatex:43:6: type mismatch;
[error]  found   : scalatex.site.Section.Proxy
[error]  required: scalatags.Text.all.Frag
[error]     (which expands to)  scalatags.generic.Frag[scalatags.text.Builder,String]
[error] @sect{scala-js-dom}
[error]      ^
[error] one error found
[error] (readme / Compile / compileIncremental) Compilation failed
```
It could be an issue with my environment - I am running Windows. WSL returned same errors.

armanbilge:
Hmm, looks like a problem with the README, I wonder if any of the code you changed touched that somehow. Let me try it as well.

armanbilge:
Ah, of course. The docs use `Fetch` which you just changed here right?
https://scala-js.github.io/scala-js-dom/#dom.Fetch

armanbilge:
You may want to check the code in here, see if there's something you need to update:
https://github.com/scala-js/scala-js-dom/blob/master/example/src/main/scala/example/Example.scala

ptrdom:
Yes, there is a change to `BodyInit` type - new member added to the union. Code of the example does not interact with the request body directly, not sure what should be changed there.

ptrdom:
CI does show some actual Firefox and Scala 3 errors, those I obviously can fix.

armanbilge:
Does `prePR` fail for you _before_ you make any changes? E.g. if you just it on `master`.

armanbilge:
Huh, with that same error message? Now that's definitely strange :)

dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
```diff
@@ -1,14 +1,19 @@
...
 
 /** defined at [[https://streams.spec.whatwg.org/#readable-stream ¶2.1. Readable Streams]] of whatwg Streams spec.
   *
   * @tparam T
   *   Type of the Chunks returned by the Stream. Can't make it coveriant, due to T
   */
 @js.native
-trait ReadableStream[+T] extends js.Object {
+@JSGlobal
+class ReadableStream[+T](
```
armanbilge, line 13: Unfortunately we cannot convert a `trait` to a class, this is binary incompatible :( I hope we can find a way to make this work even as a `trait`.

armanbilge (changes requested):

dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
ptrdom: Converted it back to a `trait` and added construction in companion object's `apply` through `js.Dynamic`, is that fine?

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
```diff
@@ -36,4 +41,42 @@ trait SharedTests {
...
+        Tuna("red")
+    )
+
+    val stream = ReadableStream[Tuna](
+        new ReadableStreamUnderlyingSource[Tuna] {
+          start = { (controller: ReadableStreamController[Tuna]) =>
+            controller.enqueue(Tuna("blue"))
+            controller.enqueue(Tuna("red"))
+            controller.close()
+          }: js.Function1[ReadableStreamController[Tuna], Unit]
```
ptrdom, line 60: Scala 2.11 complains about this cast, I just could not figure out how to solve it, assistance needed 😬

armanbilge: Hmm, try changing the cast to 
```suggestion
          }: js.UndefOr[js.Function1[ReadableStreamController[Tuna], js.UndefOr[Unit]]]
```

dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
armanbilge: At a glance looks good 👍 thanks for making this work!

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
```diff
@@ -36,4 +41,42 @@ trait SharedTests {
 
   @Test final def WindowIdbTest(): AsyncResult =
     IdbTest(window.indexedDB)
+
+  @Test
+  final def ReadableStreamTest: AsyncResult = async {
```
armanbilge, line 46: JSDom doesn't support `ReadableStream` yet so I don't think this can be a shared test.
- https://github.com/jsdom/jsdom/pull/3200

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
ptrdom: Getting two errors with that cast:
```
\scala-js-dom\tests-shared\src\main\scala\org\scalajs\dom\tests\shared\SharedTests.scala:56:66: type mismatch;
[error]  found   : org.scalajs.dom.ReadableStreamController[Tuna] => Unit
[error]  required: scala.scalajs.js.UndefOr[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.UndefOr[Unit]]]
[error]     (which expands to)  scala.scalajs.js.|[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.|[Unit,Unit]],Unit]
[error]           start = { (controller: ReadableStreamController[Tuna]) =>
[error]                                                                  ^
```
```
\scala-js-dom\tests-shared\src\main\scala\org\scalajs\dom\tests\shared\SharedTests.scala:60:12: type mismatch;
[error]  found   : scala.scalajs.js.UndefOr[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.UndefOr[Unit]]]
[error]     (which expands to)  scala.scalajs.js.|[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.|[Unit,Unit]],Unit]
[error]  required: scala.scalajs.js.UndefOr[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.|[Unit,scala.scalajs.js.Promise[Unit]]]]
[error]     (which expands to)  scala.scalajs.js.|[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.|[Unit,scala.scalajs.js.Promise[Unit]]],Unit]
[error]           }: js.UndefOr[js.Function1[ReadableStreamController[Tuna], js.UndefOr[Unit]]]
[error]            ^
```

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
ptrdom: Where would you prefer this test to be moved? Could just create a new trait in shared module, but this test is not really shared. Could create a new module only with this test that chrome and firefox test modules will have as a dependency, but it could be considered an overkill.

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
armanbilge: Hmm, I would just keep following the hints for what is "required" 😅

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
armanbilge: Yes, I vote for creating a new trait `BrowserTests` in shared. Hopefully JSDOM will catch up with `ReadableStream` support and it can eventually become a proper shared test.

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
```diff
@@ -14,8 +13,7 @@ import scala.scalajs.js.annotation._
   *   Type of the Chunks to be enqueued to the Stream
   */
 @js.native
-@JSGlobal
-class ReadableStreamController[-T](stream: ReadableStream[T] = null) extends js.Object {
+trait ReadableStreamController[-T] extends js.Object {
```
armanbilge, line 16: Oops, I think this is another one that can't change `class -> trait`. In general this change is not binary compatible.

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
```diff
@@ -30,11 +30,8 @@ class ReadableStreamReader[+T](stream: ReadableStream[T]) extends js.Object {
     *
     * If the reader is active, the cancel method behaves the same as that for the associated stream. When done, it
     * automatically releases the lock.
-    *
-    * //todo determine type of reason
     */
-  // not actually sure what the return type is here
-  def cancel(reason: Any): js.Promise[Any] = js.native
+  def cancel[U](reason: js.UndefOr[U]): js.Promise[U] = js.native
```
armanbilge, line 34:
```suggestion
  def cancel[U](reason: js.UndefOr[U]): js.Promise[js.UndefOr[U]] = js.native
```

If the input is undefined, I assume the output can be as well?

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
ptrdom: This change can be left out, I just wanted to remove it because this constructor is basically not provided and should not be called, as per spec.

dom/src/main/scala/org/scalajs/dom/ReadableStreamType.scala
```diff
@@ -0,0 +1,11 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** [[https://streams.spec.whatwg.org/#enumdef-readablestreamtype ReadableStreamType enum]] */
+@js.native
+sealed trait ReadableStreamType extends js.Any
```
armanbilge, line 7: For an example of how we are implementing `enum`s check the `scala-2` and `scala-3` specific source folders. This looks good for Scala 2 I think.

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
ptrdom: I wonder how the compiler manages to get to `scala.scalajs.js.UndefOr[scala.scalajs.js.Function1[org.scalajs.dom.ReadableStreamController[Tuna],scala.scalajs.js.UndefOr[Unit]]]`, `scala.scalajs.js.UndefOr[Unit]` is basically `Unit | Unit`? This is an odd and very specific Scala 2.11 issue.

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
armanbilge: Yeah, I regret not spending more time with this `ReadableStream` stuff before locking down the binary for 2.0.0. We'll have to fix these in 3.0.0 :(

I think you can just remove the constructor (maybe even make it `private`?) as a workaround.

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
sjrd: In that case:
```suggestion
  def cancel[U](reason: U): js.Promise[U] = js.native
```

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
armanbilge: It is strange. I'll looking it in more detail later, if you can get everything else working until then that would be great 😁

sjrd: Try wrapping the lambda with `js.defined(...)`. That should help the type inference.

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
ptrdom: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/cancel does show a call with no arguments passed - and my guess that is the most common usage - so should it be `def cancel[U](reason: js.UndefOr[U] = js.undefined): js.Promise[js.UndefOr[U]] = js.native` then?

ptrdom: Although `reason: U` will not return `js.undefined`, so my suggestion is also not ideal.

armanbilge: What if we just provide two overloads of this method?

```scala
def cancel[U](reason: U): js.Promise[U] = js.native
def cancel(): js.Promise[Unit] = js.native
```

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
sjrd: Make it `private[dom]`. That should do the trick.

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
sjrd: Consider using 2 overloads, then:
```scala
  def cancel[U](reason: U): js.Promise[U] = js.native
  def cancel(): js.Promise[Unit] = js.native
```

ptrdom: I was a bit late to also reply with the same message 😄

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
ptrdom: Tried these, still got errors:
```
start = js.defined({ (controller: ReadableStreamController[Tuna]) =>
  controller.enqueue(Tuna("blue"))
  controller.enqueue(Tuna("red"))
  controller.close()
}): js.UndefOr[js.Function1[ReadableStreamController[Tuna], Unit | js.Promise[Unit]]]
```
```
start = js.defined({ (controller: ReadableStreamController[Tuna]) =>
  controller.enqueue(Tuna("blue"))
  controller.enqueue(Tuna("red"))
  controller.close()
}: js.Function1[ReadableStreamController[Tuna], Unit | js.Promise[Unit]])
```
```
start = js.defined({ (controller: ReadableStreamController[Tuna]) =>
  controller.enqueue(Tuna("blue"))
  controller.enqueue(Tuna("red"))
  controller.close()
}: js.Function1[ReadableStreamController[Tuna], Unit])
```
```
start = js.defined({ (controller: ReadableStreamController[Tuna]) =>
  controller.enqueue(Tuna("blue"))
  controller.enqueue(Tuna("red"))
  controller.close()
}): js.UndefOr[js.Function1[ReadableStreamController[Tuna], Unit]]
```

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
ptrdom:
```
\scala-js-dom\dom\src\main\scala\org\scalajs\dom\ReadableStreamController.scala:18:36: Native JS classes may not have private constructors. Use `private[this]` to declare an internal constructor.
[error] class ReadableStreamController[-T] private[dom] (stream: ReadableStream[T] = null) extends js.Object {
```
What would be an ideal solution then?

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
ptrdom: 2.12 is okay with:
```
start = js.defined({ (controller: ReadableStreamController[Tuna]) =>
  controller.enqueue(Tuna("blue"))
  controller.enqueue(Tuna("red"))
  controller.close()
}): js.UndefOr[js.Function1[ReadableStreamController[Tuna], Unit | js.Promise[Unit]]]
```
But 2.11 does not accept anything 😞

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
sjrd: Try `private[this]`, as suggested.

ptrdom:
I think the only issue left now is the very first one - https://github.com/scala-js/scala-js-dom/pull/628#issuecomment-965546913.

ptrdom:
And it was a Windows related issue. Big thanks to Microsoft for WSL 👏

armanbilge:
Nice work getting it green! 🎉

dom/src/main/scala/org/scalajs/dom/ReadableStreamQueuingStrategy.scala
```diff
@@ -0,0 +1,23 @@
...
+  var highWaterMark: Double
+
+  /** (non-byte streams only)
+    *
+    * The result is used to determine backpressure, manifesting via the appropriate desiredSize property. For readable
+    * streams, it also governs when the underlying source's [[ReadableStreamUnderlyingSource.pull]] method is called.
+    *
+    * A function that computes and returns the finite non-negative size of the given chunk value.
+    */
+  var size: js.Function1[Chunk[T], Unit]
```
armanbilge, line 22:
```suggestion
  var size: js.Function1[T, Int]
```

Right?

dom/src/main/scala/org/scalajs/dom/ReadableStreamUnderlyingSource.scala
```diff
@@ -0,0 +1,58 @@
...
+  *   Type of the Chunks returned by the Stream
+  */
+trait ReadableStreamUnderlyingSource[T] extends js.Object {
+
+  /** A function that is called immediately during creation of the ReadableStream.
+    *
+    * If this setup process is asynchronous, it can return a promise to signal success or failure; a rejected promise
+    * will error the stream. Any thrown exceptions will be re-thrown by the [[ReadableStream]] constructor.
+    */
+  var start: js.UndefOr[js.Function1[ReadableStreamController[T], Unit | js.Promise[Unit]]] = js.undefined
```
armanbilge, line 19:
```suggestion
  var start: js.UndefOr[js.Function1[ReadableStreamController[T], js.UndefOr[js.Promise[Unit]]]] = js.undefined
```

They are aliases.

dom/src/main/scala/org/scalajs/dom/ReadableStreamUnderlyingSource.scala
```diff
@@ -0,0 +1,58 @@
...
+    * This function will not be called until [[start]] successfully completes. Additionally, it will only be called
+    * repeatedly if it enqueues at least one chunk or fulfills a BYOB request; a no-op [[pull]] implementation will not
+    * be continually called.
+    *
+    * If the function returns a promise, then it will not be called again until that promise fulfills. (If the promise
+    * rejects, the stream will become errored.) This is mainly used in the case of pull sources, where the promise
+    * returned represents the process of acquiring a new chunk. Throwing an exception is treated the same as returning a
+    * rejected promise.
+    */
+  var pull: js.UndefOr[js.Function1[ReadableStreamController[T], Unit | js.Promise[Unit]]] = js.undefined
```
armanbilge, line 34: `js.UndefOr` here, like above.

dom/src/main/scala/org/scalajs/dom/ReadableStreamUnderlyingSource.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  /** A function that is called whenever the consumer cancels the stream, via [[ReadableStream.cancel]] or
+    * [[ReadableStreamReader.cancel():scala\.scalajs\.js\.Promise[Unit]*]]. It takes as its argument the same value as
+    * was passed to those methods by the consumer. If the shutdown process is asynchronous, it can return a promise to
+    * signal success or failure; the result will be communicated via the return value of the [[cancel]] method that was
+    * called. Additionally, a rejected promise will error the stream, instead of letting it close. Throwing an exception
+    * is treated the same as returning a rejected promise.
+    */
+  var cancel: js.UndefOr[js.Function1[js.Any, Unit | js.Promise[Unit]]] = js.undefined
```
armanbilge, line 43: `js.UndefOr` here, like above.

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/BrowserTests.scala
```diff
@@ -0,0 +1,56 @@
...
+trait BrowserTests {
+
+  @Test
+  final def ReadableStreamTest: AsyncResult = async {
+    case class Tuna(color: String)
+
+    val expectedTunas = Seq(
+        Tuna("blue"),
+        Tuna("red")
+    )
```
armanbilge, line 26: One_Fish_Two_Fish_Red_Fish_Blue_Fish_(cover_art)

armanbilge (commented):
This is a _very_ complex API, so thank you for all your efforts and apologies if I'm confused about something in the comments!

ptrdom:
@armanbilge Thank you for the review! I think I have addressed all of your comments.

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
```diff
@@ -30,11 +30,9 @@ class ReadableStreamReader[+T](stream: ReadableStream[T]) extends js.Object {
     *
     * If the reader is active, the cancel method behaves the same as that for the associated stream. When done, it
     * automatically releases the lock.
-    *
-    * //todo determine type of reason
     */
-  // not actually sure what the return type is here
-  def cancel(reason: Any): js.Promise[Any] = js.native
+  def cancel[U](reason: U): js.Promise[U] = js.native
+  def cancel(): js.Promise[Unit] = js.native
```
armanbilge, line 35: Sorry to fiddle with this again :( I think the return type should just be `js.Promise[Unit]`, at least according to this:
https://streams.spec.whatwg.org/#generic-reader-mixin-definition

The problem is that MDN is inconsistent and suggests an API like what we have currently. I think it's probably better to stick to the spec, also a `Unit` return type is less likely to get users into trouble 😅 
https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/cancel

Also I think we need to fix `closed` above.

dom/src/main/scala/org/scalajs/dom/ReadableStreamUnderlyingSource.scala
```diff
@@ -0,0 +1,57 @@
...
+    * exception.
+    */
+  var `type`: js.UndefOr[ReadableStreamType] = js.undefined
+
+  /** (byte streams only)
+    *
+    * Can be set to a positive integer to cause the implementation to automatically allocate buffers for the underlying
+    * source code to write into.
+    */
+  var autoAllocateChunkSize: js.UndefOr[Double] = js.undefined
```
armanbilge, line 56: Maybe make this an `Int` too?

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
armanbilge: Aha, maybe it's not inconsistent. The MDN docs are for `ReadableStreamDefaultReader`, but the spec is for the `ReadableStreamGenericReader` like what we have here. So we should definitely follow the spec here, sorry!

armanbilge (commented):
Sorry, just a few more things I noticed on this pass!

Let's get another pair of eyes, whenever you have a chance @sjrd or @japgolly, much appreciated! :)

dom/src/main/scala/org/scalajs/dom/ReadableStreamController.scala
```diff
@@ -39,7 +39,7 @@ class ReadableStreamController[-T](stream: ReadableStream[T] = null) extends js.
     * @return
     *   seems like its an undefOr[Int] of the size
     */
-  def enqueue(chunk: Chunk[T]): js.UndefOr[Int] = js.native
+  def enqueue(chunk: T): js.UndefOr[Int] = js.native
```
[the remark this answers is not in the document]
ptrdom: What is the purpose of having `chunk: js.UndefOr[T]`? If there would be a `ReadableStream[Unit]`, then `T` could just be `Unit`. If it is anything but a `Unit`, I could not find anywhere that it should be possible to pass a `Unit`.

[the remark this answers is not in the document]
armanbilge: That's a good question. I'm just going by the definition in the spec, which suggests it can be optional:
https://streams.spec.whatwg.org/#rs-default-controller-class-definition
```idl
interface ReadableStreamDefaultController {
  readonly attribute unrestricted double? desiredSize;

  undefined close();
  undefined enqueue(optional any chunk);
  undefined error(optional any e);
};
```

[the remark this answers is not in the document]
ptrdom: This change does produce an error Scala 2.11 - `contravariant type T occurs in invariant position in type scala.scalajs.js.UndefOr[T] of value chunk`. Is contravariance needed in `ReadableStreamController`?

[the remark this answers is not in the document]
armanbilge: Not again 😅  we can do the overload trick again here instead.

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
ptrdom: I wrote a test for this one, neither of browsers return a `js.Promise` with given reason.

armanbilge: Thanks for looking into this! Much appreciated. I guess MDN is just wrong ...

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
```diff
@@ -31,7 +31,7 @@ class ReadableStreamReader[+T](stream: ReadableStream[T]) extends js.Object {
     * If the reader is active, the cancel method behaves the same as that for the associated stream. When done, it
     * automatically releases the lock.
     */
-  def cancel[U](reason: U): js.Promise[U] = js.native
+  def cancel[U](reason: U): js.Promise[Unit] = js.native
```
armanbilge, line 34: We no longer need the type parameter :)

```suggestion
  def cancel(reason: Any): js.Promise[Unit] = js.native
```

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
ptrdom: It did sound weird to me when I read it, kind of a pointless loop through of a value.

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/BrowserTests.scala
```diff
@@ -77,22 +79,53 @@ trait BrowserTests {
...
+      _ <- stream
+        .getReader()
+        .cancel()
+        .map(assertEquals((), _))
+      _ <- reasonPromise.future.map(assertEquals((), _))
+    } yield ()
+  }
+
+  @Test
+  final def ReadableStreamReaderCancelPromiseTest: AsyncResult = async {
```
armanbilge, line 110: 👍

armanbilge (commented):
This looks great, thanks for all your efforts! Just one last tiny thing 😅

dom/src/main/scala/org/scalajs/dom/ReadableStreamReader.scala
ptrdom: Ah, committing the suggestion will not be enough, need to regenerate `api-reports` 😞

armanbilge: Sorry about that :( we'd like to phase out the API reports b/c it's frustrating for contributors, we have a plan in https://github.com/scala-js/scala-js-dom/issues/503.

armanbilge (approved):

dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
```diff
@@ -92,3 +92,18 @@ trait ReadableStream[+T] extends js.Object {
...
+  def apply[T](
+      underlyingSource: js.UndefOr[ReadableStreamUnderlyingSource[T]] = js.undefined,
+      queuingStrategy: js.UndefOr[QueuingStrategy[T]] = js.undefined
+  ): ReadableStream[T] = {
+    js.Dynamic
+      .newInstance(js.Dynamic.global.selectDynamic("ReadableStream"))(
+          underlyingSource.asInstanceOf[js.Any],
+          queuingStrategy.asInstanceOf[js.Any]
+      )
+      .asInstanceOf[ReadableStream[T]]
```
sjrd, line 107: I'm afraid this is plain incorrect. It won't create a valid `ReadableStream`, since the result won't be an `instanceof ReadableStream`. It also won't have all the methods that `ReadableStream` is supposed to have.

Since it is a class, it really need to be instantiated with `new`. You could achieve that with `js.Dynamic.newInstance`, but I don't think that is the best choice here, as it will (re-)create a precedent for `apply` methods that actually create stuff.

Instead, I think the best path forward is:
1. Define a `class ReadableStreamClass` that extends `ReadableStream` and provides the correct constructor.
2. Add a self-type to `ReadableStream` that it must be a `ReadableStreamClass`.
3. Document that `ReadableStream` should be created through `ReadableStreamClass`, but other remain typed everywhere as `ReadableStream` (for historical/compatibility reasons).

Concretely, it would look like this:
```scala
/** defined at [[https://streams.spec.whatwg.org/#readable-stream ¶2.1. Readable Streams]] of whatwg Streams spec.
  *
  * For historical reasons, ReadableStream is a trait rather than a class.
  * In order to instantiate it, use `new ReadableStreamClass(...)`.
  * Prefer `ReadableStream` in typings everywhere else, notably in
  * method signatures.
  *
  * @tparam T
  *   Type of the Chunks returned by the Stream. Can't make it coveriant, due to T
  */
@js.native
trait ReadableStream[+T] extends js.Object { this: ReadableStreamClass =>
  ...
}

@js.native
@JSGlobal("ReadableStream")
class ReadableStreamClass(underlyingSource: ..., queueingStrategy: ...)
    extends ReadableStream
```
When you get to scalajs-dom v3, you can turn `ReadableStream` into a class, and make `ReadableStreamClass` an `@deprecated type` alias of `ReadableStream`.

sjrd (changes requested):

dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
armanbilge: Hum, I'm not sure I agree. Would you mind expanding more on this point?

> I don't think that is the best choice here, as it will (re-)create a precedent for apply methods that actually create stuff.

We all agree we want to avoid this precedent ... but is this a precedent for users, or a precedent for maintainers, or both/something else?

My concern is about discovery: it's much more intuitive to look to the `apply` method in the companion object then for a differently-named class.

Furthermore, _if_ scala-js-dom v3 ever happens (😅) by that time we will hopefully be in a Scala 3+ world where _all_ classes can be created without the `new` keyword anyway! So we could make this change source-compatibly I think?

sjrd: Hum, perhaps you're right. I guess we can leave it as is, then.

dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
```diff
@@ -92,3 +92,18 @@ trait ReadableStream[+T] extends js.Object {
...
 }
+
+object ReadableStream {
+
+  def apply[T](
+      underlyingSource: js.UndefOr[ReadableStreamUnderlyingSource[T]] = js.undefined,
+      queuingStrategy: js.UndefOr[QueuingStrategy[T]] = js.undefined
+  ): ReadableStream[T] = {
+    js.Dynamic
+      .newInstance(js.Dynamic.global.selectDynamic("ReadableStream"))(
```
sjrd, line 103:
```suggestion
      .newInstance(js.Dynamic.global.ReadableStream)(
```

sjrd (approved):

armanbilge:
@ptrdom thanks again for this fantastic contribution! It is now published as `2.0.0+25-ed28f246-SNAPSHOT`.

julienrf:
Hello! Do you plan to cut a release with this changes soon?

armanbilge:
@julienrf hello :) Out of curiosity, is this blocking something? Although the API facaded here is described in the spec, it's basically unsupported in browsers at this time.
https://caniuse.com/mdn-api_request_request_readablestream_request_body

Otherwise, personally I'm 👍 to a release :) Since we've added new APIs, the next release will be 2.1.0 in which we also need to fix https://github.com/scala-js/scala-js-dom/issues/624. So it will take a little time to get it ready.

julienrf:
It is not just about the change to the Fetch facade, the PR contains other changes related to the `ReadableStream` API. As far as I understand, those changes are currently supported by the browsers: https://caniuse.com/?search=readablestream

This is not really blocking us since we can always define the facade we want outside of scala-js-dom, but it would be simpler to just use the one of scala-js-dom. Our related PR is https://github.com/endpoints4s/endpoints4s/pull/972

armanbilge:
Ah, that's right, you want to use this as your stream implementation :) thanks for the reply. I'll start getting a release together.

armanbilge:
@julienrf I just released [v2.1.0](https://github.com/scala-js/scala-js-dom/releases/tag/v2.1.0). @ptrdom thank you again for this fantastic contribution, also looks like you've been doing great work in endpoints4s :)

julienrf:
Thank you!

PR #639, opened 2021-11-29
Change ReadableStream#pipeTo to return js.Promise[Unit]

matil019:
Nice to meet you!

According to [the whatwg spec](https://streams.spec.whatwg.org/#rs-class-definition), `ReadableStream#pipeTo` should return a `Promise<undefined>`. This PR modifies the binding to return `js.Promise[Unit]` instead of a plain `Unit`.

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
+++ b/dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
@@ -75,7 +75,7 @@ trait ReadableStream[+T] extends js.Object {
     *
     * //todo: determine the type of options
     */
-  def pipeTo(dest: WriteableStream[T], options: Any = js.native): Unit = js.native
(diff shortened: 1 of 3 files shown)


dom/src/main/scala/org/scalajs/dom/ReadableStream.scala
```diff
@@ -75,7 +75,7 @@ trait ReadableStream[+T] extends js.Object {
     *
     * //todo: determine the type of options
     */
-  def pipeTo(dest: WriteableStream[T], options: Any = js.native): Unit = js.native
+  def pipeTo(dest: WriteableStream[T], options: Any = js.native): js.Promise[Unit] = js.native
```
armanbilge, line 78: If you're interested, while we are here it would be nice to fix the todo create a facade for the options :) no pressure!

armanbilge (approved):
Nice catch and thanks for the PR!

PR #657, opened 2021-12-26
Always use `var` and `undefined` for `new`-based constructors

### LINKED ISSUE #624: Consistently set defaults for `new`-based constructors

armanbilge:
Fixes #624. I went through all of the -`Init` classes and checked that they are using `var` and `js.undefined` as their default value. Are there non-`Init` classes that we should check too?

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/ResponseInit.scala
+++ b/dom/src/main/scala/org/scalajs/dom/ResponseInit.scala
@@ -4,7 +4,7 @@ import scala.scalajs.js
 
 /** See [[https://fetch.spec.whatwg.org/#response-class ¶6.4 Response class]] definition in whatwg Fetch spec. */
 trait ResponseInit extends js.Object {
-  var status: Int
-  var statusText: ByteString
-  var headers: HeadersInit
+  var status: js.UndefOr[Int] = js.undefined
+  var statusText: js.UndefOr[ByteString] = js.undefined
+  var headers: js.UndefOr[HeadersInit] = js.undefined
 }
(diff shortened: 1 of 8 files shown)


dom/src/main/scala/org/scalajs/dom/ResponseInit.scala
```diff
@@ -4,7 +4,7 @@ import scala.scalajs.js
 
 /** See [[https://fetch.spec.whatwg.org/#response-class ¶6.4 Response class]] definition in whatwg Fetch spec. */
 trait ResponseInit extends js.Object {
-  var status: Int
-  var statusText: ByteString
-  var headers: HeadersInit
+  var status: js.UndefOr[Int] = js.undefined
+  var statusText: js.UndefOr[ByteString] = js.undefined
+  var headers: js.UndefOr[HeadersInit] = js.undefined
 }
```
armanbilge, line 10: @sjrd is changing these return signatures ok for bincompat? I realized I don't know what the semantics are for `js.Object`.

armanbilge:
Oh btw, would everyone be okay with releasing v2.1.0 shortly after we merge this PR? I may fix another open issue or two but not really a priority. Feel free to point out any that are important.

dom/src/main/scala/org/scalajs/dom/ResponseInit.scala
sjrd: Yes, this is fine.

sjrd (approved):

PR #667, opened 2022-01-12
Add WebCryptoApi tests

AlexITC:
Initially, this showcased a broken method in the web crypto api facade.

Let's add some tests to prevent this from occurring again, also, these tests should be useful for people willing to use the web crypto api in scalajs.


tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
```diff
@@ -3,7 +3,7 @@ package org.scalajs.dom.tests.shared
 import org.scalajs.dom.tests.shared.AsyncTesting._
 import org.junit.Test
 
-trait SharedTests {
+trait SharedTests extends WebCryptoApiTests {
```
armanbilge, line 6: I think `SharedTests` are also run on JSDOM which IIRC doesn't support Web Crypto. If it doesn't work here, we can mix them into the `BrowserTests` trait instead. Otherwise, very nicely organized 👍

AlexITC: Thanks for the hint, I'm still discovering details and I can't run all tests locally.

I'm looking into some weird errors in scala 2.11:

```
[error] /scala-js-dom/tests-shared/src/main/scala/org/scalajs/dom/tests/shared/WebCryptoApiTests.scala:35:11: overriding value salt in trait Pbkdf2Params of type org.scalajs.dom.BufferSource;
[error]  value salt has incompatible type
[error]       val salt = "salt".getBytes.toTypedArray.buffer
[error]           ^
[error] /scala-js-dom/tests-shared/src/main/scala/org/scalajs/dom/tests/shared/WebCryptoApiTests.scala:37:11: overriding value hash in trait Pbkdf2Params of type org.scalajs.dom.HashAlgorithmIdentifier;
[error]  value hash has incompatible type
[error]       val hash = "SHA-512"
[error]           ^
[error] /scala-js-dom/tests-shared/src/main/scala/org/scalajs/dom/tests/shared/WebCryptoApiTests.scala:93:17: overriding value additionalData in trait AesGcmParams of type org.scalajs.dom.BufferSource;
[error]  value additionalData has incompatible type
[error]             val additionalData = "".getBytes.toTypedArray.buffer
[error]                 ^
[error] /scala-js-dom/tests-shared/src/main/scala/org/scalajs/dom/tests/shared/WebCryptoApiTests.scala:109:17: overriding value additionalData in trait AesGcmParams of type org.scalajs.dom.BufferSource;
[error]  value additionalData has incompatible type
[error]             val additionalData = "".getBytes.toTypedArray.buffer
```

AlexITC: I should have fixed everything, the CI needs you approval to run the workflow.

Thanks.

armanbilge (approved):
Thanks, LGTM! Much appreciated :)

Now let's see if we can come up with a fix in https://github.com/scala-js/scala-js-dom/pull/669...

PR #669, opened 2022-01-12
Fix `dom.crypto.getRandomValues`

### LINKED ISSUE #668: Some methods from dom.crypto are broken
scala-js-dom 2.0.0 deprecated the `dom.crypto.crypto` API, suggesting to use [dom.crypto](https://github.com/scala-js/scala-js-dom/blob/main/dom/src/main/scala/org/scalajs/dom/crypto/package.scala) package instead. Unfortunately, many methods (if not all) are broken.

On the other hand, `dom.crypto.crypto` and `dom.webcrypto` seem to work.

I'm yet to try all of the methods from these packages, I have tried some methods and all of them behave this way, it would be ideal to add test coverage for the package.

To showcase the problem, I have created a draft PR (#667), I'm happy to submit the test coverage and fix the integration but I'm not sure what's the acceptable way, I could just go linking the new packages to `GlobalCrypto` as a simple fix.

Thanks.

armanbilge:
Will fix #668.


armanbilge:
For df03142 I get the following in CI:
```
[error] Test org.scalajs.dom.tests.chrome.ChromeTests.cryptoGetRandomValues failed: scala.scalajs.js.JavaScriptException: TypeError: Illegal invocation, took 0.000300001 sec
[error]     at org.scalajs.dom.tests.shared.BrowserTests$class.cryptoGetRandomValues(file:///home/runner/work/scala-js-dom/scala-js-dom/tests-chrome/target/scala-2.11/testschrome-test-fastopt/main.js:3106)

[error] Test org.scalajs.dom.tests.firefox.FirefoxTests.cryptoGetRandomValues failed: scala.scalajs.js.JavaScriptException: TypeError: 'getRandomValues' called on an object that does not implement interface Crypto., took 0 sec
[error]     at org.scalajs.dom.tests.shared.BrowserTests$class.cryptoGetRandomValues(file:///home/runner/work/scala-js-dom/scala-js-dom/tests-firefox/target/scala-2.11/testsfirefox-test-fastopt/main.js:3106)
```

dom/src/main/scala/org/scalajs/dom/crypto/package.scala
```diff
@@ -15,9 +15,8 @@ package object crypto {
   val subtle: dom.SubtleCrypto = js.native
 
   /** Fills the passed TypedArray with cryptographically sound random values. */
-  @JSGlobal("crypto.getRandomValues")
-  @js.native
-  def getRandomValues(array: ArrayBufferView): ArrayBufferView = js.native
+  @inline def getRandomValues(array: ArrayBufferView): ArrayBufferView =
+    webcrypto.getRandomValues(array)
```
sjrd, line 19: That's not a binary compatible change. An `@js.native` thing must remain `@js.native`, and non-native thing must remain non-native.

sjrd (changes requested):

dom/src/main/scala/org/scalajs/dom/crypto/package.scala
armanbilge: Ah, thanks, I wanted to ask you about that.

Then, I'm not sure if this is possible to fix at all. Do you have any suggestions?

sjrd: Not right now. I'm about to go to sleep.

But what is *actually* the issue in the first place? Why doesn't the `@JSGlobal` work?

armanbilge: No worries, look when you have a chance and get well soon :)

I posted the error in https://github.com/scala-js/scala-js-dom/pull/669#issuecomment-1011519586.

armanbilge: Anyone have any ideas for this one?

japgolly: A binary-incompatible workaround would require a release of 3.0.0 right? I think the only option is to fix what looks to be a bug in Scala.js.

sjrd: Super ugly, but when the situation is desperate, one can always rely on the good old `protected[package]` trick:
```scala
  // for binary compat; unusable
  @JSGlobal("crypto.getRandomValues")
  @js.native
  protected[crypto] def getRandomValues(array: ArrayBufferView): ArrayBufferView = js.native

  @inline
  def getRandomValues(array: ArrayBufferView)(implicit dummy: DummyImplicit): ArrayBufferView =
    webcrypto.getRandomValues(array)
```

armanbilge:
I assume that releasing this fix will require a bump to 2.2.0?

sjrd (approved):

armanbilge:
After reading https://github.com/scalacenter/sbt-version-policy/issues/71 I understand forward binary compatibility is really just an approximation used for checking backward source compatibility. So even though we are breaking forward binary compatibility here, we are not breaking source compatibility, so we can release this in 2.1.1.

sjrd:
You're changing the public API (even if it's just an "added" implicit parameter) so you are potentially breaking source compatibility.

In fact, given the nature of scalajs-dom, it is extremely unlikely that there will *ever* be a patch release. It's virtually 100% API and no implementation, so any change triggers a minor version update.

armanbilge:
Thanks, that's very helpful.

Issue #683, opened 2022-03-07
encodeURIComponent/decodeURIComponent missing


AlexITC:
Ref:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

The obvious workaround being `org.scalajs.dom.window.asInstanceOf[js.Dynamic].encodeURIComponent(string).toString` available but it isn't ideal.

armanbilge:
Thanks for the report! Happy to take a PR for this :)

AlexITC:
Actually, I started the PR just after opening the ticket, which is submitted by now.

armanbilge:
Thanks, fast work! :)

sjrd:
These are not DOM-specific. There are part of the ECMAScript specification, and therefore they are in Scala.js core:
https://github.com/scala-js/scala-js/blob/main/library/src/main/scala/scala/scalajs/js/URIUtils.scala

armanbilge:
Woops! I should have paid more attention when I couldn't find them in the Typescript `lib.dom.d.ts`. Sorry.

@AlexITC you'll be happy to see it's all typed with `String` there so I guess I was wrong about that too :)

AlexITC:
Thanks for sharing, I made the mistake to rely on github search when looking for these methods.

PR #684, opened 2022-03-07
Add encodeURIComponent/decodeURIComponent methods

AlexITC:
Solves #683

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Window.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Window.scala
@@ -457,7 +457,7 @@ class Window
     * @return
     *   A new string representing the provided uriComponent encoded as a URI component.
     */
-  def encodeURIComponent(uriComponent: String): String = js.native
+  def encodeURIComponent(uriComponent: js.Any): String = js.native
 
   /** Decodes a Uniform Resource Identifier (URI) component previously created by [[encodeURIComponent]] or by a similar
     * routine.
--- a/dom/src/main/scala/org/scalajs/dom/Window.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Window.scala
@@ -457,7 +457,7 @@ class Window
     * @return
     *   A new string representing the provided uriComponent encoded as a URI component.
     */
-  def encodeURIComponent(uriComponent: js.Any): String = js.native
+  def encodeURIComponent(uriComponent: Any): String = js.native
 
   /** Decodes a Uniform Resource Identifier (URI) component previously created by [[encodeURIComponent]] or by a similar
     * routine.
--- a/dom/src/main/scala/org/scalajs/dom/Window.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Window.scala
@@ -445,4 +445,31 @@ class Window
     * number. You can get the number of pixels the document is scrolled horizontally from the scrollX property.
     */
   def scrollY: Double = js.native
+
+  /** Encodes a URI by replacing each instance of certain characters by one, two, three, or four escape sequences
+    * representing the UTF-8 encoding of the character (will only be four escape sequences for characters composed of
+    * two "surrogate" characters).
+    *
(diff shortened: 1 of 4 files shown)


dom/src/main/scala/org/scalajs/dom/Window.scala
```diff
@@ -445,4 +445,31 @@ class Window
...
     */
   def scrollY: Double = js.native
+
+  /** Encodes a URI by replacing each instance of certain characters by one, two, three, or four escape sequences
+    * representing the UTF-8 encoding of the character (will only be four escape sequences for characters composed of
+    * two "surrogate" characters).
+    *
+    * @param uriComponent
+    *   A string, number, boolean, null, undefined, or any object. Before encoding, the uriComponent gets converted to
+    *   string.
```
armanbilge, line 455: Hmm, should we model this with `Any` instead of `String`?

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
```diff
@@ -36,4 +36,11 @@ trait SharedTests {
 
   @Test final def WindowIdbTest(): AsyncResult =
     IdbTest(window.indexedDB)
+
+  @Test final def encodeDecodeUriTest(): Unit = {
+    val input = "?=/&:шеллы"
+    val encoded = window.encodeURIComponent(input)
+    val decoded = window.decodeURIComponent(encoded)
+    org.junit.Assert.assertEquals(input, decoded)
+  }
```
armanbilge, line 45: 👍 nice!

tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala
```diff
@@ -36,4 +36,11 @@ trait SharedTests {
 
   @Test final def WindowIdbTest(): AsyncResult =
     IdbTest(window.indexedDB)
+
+  @Test final def encodeDecodeUriTest(): Unit = {
```
armanbilge, line 40:
```suggestion
  @Test final def encodeDecodeUriComponentTest(): Unit = {
```

dom/src/main/scala/org/scalajs/dom/Window.scala
AlexITC: Good question, I'd personally leave it as a string because it would avoid questions on how the method behaves when it gets something different. If you still want me to change it to `Any`, let me know. I don't see many Scala projects providing `Any` here and I'd vote for type-safety instead.

armanbilge: I hope the other maintainers can weigh in on this as well, but I'd be inclined to stick to the spec and use `Any`. 

For example, someone can open an issue saying they want this method for boolean or number, and we should absolutely accept a PR for that because the spec says it is supported. So then when someone wants this method for any object, then we up right here :)

> I'd personally leave it as a string because it would avoid questions on how the method behaves when it gets something different.

Since this is only a facade, these are questions for the underlying implementation :)

AlexITC:
Comments addressed, thanks!

dom/src/main/scala/org/scalajs/dom/Window.scala
```diff
@@ -445,4 +445,31 @@ class Window
...
+    *
+    * @return
+    *   A new string representing the provided uriComponent encoded as a URI component.
+    */
+  def encodeURIComponent(uriComponent: js.Any): String = js.native
```
armanbilge, line 460: There's very few circumstances to use `js.Any`.

```suggestion
  def encodeURIComponent(uriComponent: Any): String = js.native
```

armanbilge (approved):
Thank you very much! :)

sjrd (changes requested):
See my comment here:
https://github.com/scala-js/scala-js-dom/issues/683#issuecomment-1060251593

PR #685, opened 2022-03-07
Allow to easily override the chrome driver's path

AlexITC:
The chrome driver's path can now be set through the `CHROME_DRIVER`
env variable, which is necessary for running tests.

Example:
- `CHROME_DRIVER=/bin/chromedriver sbt prePR`

This is an undocumented setting that is required while contributing changes.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -93,6 +93,10 @@ object Build {
     .settings(
       Test / jsEnv := {
         System.setProperty("webdriver.chrome.silentOutput", "true")
+        // Allow setting up the chrome driver's absolute path from an environment variable
+        sys.env.get("CHROME_DRIVER").foreach { chromeDriver =>
+          System.setProperty("webdriver.chrome.driver", chromeDriver)
+        }
         val o = new ChromeOptions()
         o.setHeadless(true)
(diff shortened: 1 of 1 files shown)


armanbilge:
Seems fine, but how is this different from `sbt -Dwebdriver.chrome.driver=/bin/chromedriver ...`?

AlexITC:
The main difference is that you can add `export CHROME_DRIVER=/bin/chromedriver` to your `~/.bashrc` file and forget about setting it up again, running simpler commands.

armanbilge:
Is there precedent for a `CHROME_DRIVER` env used by any other tools? Sbt supports an `~/.sbtrc` file.

AlexITC:
Thanks for the hint, this removes the need for my PR.

AlexITC:
I talked to soon, could you give an example `~/.sbtrc` that should handle this? apparently `System.setProperty("webdriver.chrome.driver", "/bin/chromedriver")` isn't a valid syntax.

armanbilge:
try `eval System.setProperty("webdriver.chrome.driver", "/bin/chromedriver")`

Once you figure it out, would appreciate a PR the contributing docs mentioning this :)

PR #686, opened 2022-03-17
Add `EventListenerOptions#signal`

armanbilge:
Somehow, there's always one more facade to add 😆 

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters


japgolly (approved):
LGTM

PR #691, opened 2022-04-06
Add `WorkerOptions` to [Shared]Worker constructors

armanbilge:
https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker
https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker/SharedWorker

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
+++ b/dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
@@ -2,6 +2,7 @@ package org.scalajs.dom
 
 import scala.scalajs.js
 import scala.scalajs.js.annotation._
+import scala.scalajs.js.|
 
 /** The SharedWorker interface represents a specific kind of worker that can be accessed from several browsing contexts,
   * such as several windows, iframes or even workers. They implement an interface different than dedicated workers and
@@ -17,13 +18,14 @@ import scala.scalajs.js.annotation._
   *   {{{var myWorker = new SharedWorker("aURL", name);}}}
   * @param stringUrl
   *   A DOMString representing the URL of the script the worker will execute. It must obey the same-origin policy.
-  * @param name
-  *   An optional argument that specifies an existing SharedWorkerGlobalScope.name — if this is specified then that
-  *   SharedWorkerGlobalScope will be used as the scope for this shared worker.
+  * @param options
+  *   A DOMString specifying an identifying name for the SharedWorkerGlobalScope representing the scope of the worker,
+  *   which is mainly useful for debugging purposes. Or, an object containing option properties that can set when
+  *   creating the object instance.
   */
 @js.native
 @JSGlobal
-class SharedWorker(stringUrl: String, name: js.UndefOr[String] = js.native) extends AbstractWorker {
+class SharedWorker(scriptURL: String, options: js.UndefOr[String | WorkerOptions] = js.native) extends AbstractWorker {
 
   /** The port property of the SharedWorker interface returns a [[MessagePort]] object used to communicate and control
     * the shared worker.
--- a/dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
+++ b/dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
@@ -16,7 +16,7 @@ import scala.scalajs.js.|
   * thrown.
   * @example
   *   {{{var myWorker = new SharedWorker("aURL", name);}}}
-  * @param stringUrl
+  * @param scriptURL
   *   A DOMString representing the URL of the script the worker will execute. It must obey the same-origin policy.
   * @param options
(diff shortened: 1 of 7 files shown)


dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
```diff
@@ -17,13 +18,14 @@ import scala.scalajs.js.annotation._
...
-  *   SharedWorkerGlobalScope will be used as the scope for this shared worker.
+  * @param options
+  *   A DOMString specifying an identifying name for the SharedWorkerGlobalScope representing the scope of the worker,
+  *   which is mainly useful for debugging purposes. Or, an object containing option properties that can set when
+  *   creating the object instance.
   */
 @js.native
 @JSGlobal
-class SharedWorker(stringUrl: String, name: js.UndefOr[String] = js.native) extends AbstractWorker {
+class SharedWorker(scriptURL: String, options: js.UndefOr[String | WorkerOptions] = js.native) extends AbstractWorker {
```
armanbilge, line 28: I wasn't quite sure the best thing to do here. The other option I guess is to have two separate constructors.

dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
```diff
@@ -17,13 +18,14 @@ import scala.scalajs.js.annotation._
   *   {{{var myWorker = new SharedWorker("aURL", name);}}}
   * @param stringUrl
```
armanbilge, line 19:
```suggestion
  * @param scriptURL
```

dom/src/main/scala/org/scalajs/dom/WorkerOptions.scala
```diff
@@ -0,0 +1,10 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+@js.native
```
sjrd, line 5: This should be non-native, and have `= js.undefined` fields.

dom/src/main/scala/org/scalajs/dom/Worker.scala
```diff
@@ -10,10 +10,15 @@ import scala.scalajs.js.annotation._
...
+  *
+  * @param scriptURL
+  *   A USVString representing the URL of the script the worker will execute. It must obey the same-origin policy.
+  * @param options
+  *   An object containing option properties that can be set when creating the object instance.
   */
 @js.native
 @JSGlobal
-class Worker(stringUrl: String) extends AbstractWorker {
+class Worker(scriptURL: String, options: js.UndefOr[WorkerOptions] = js.native) extends AbstractWorker {
```
sjrd, line 21: An overload would provide better inference here, I think. It may allow to use the `new { ... }` idiom in Scala 3.

sjrd (changes requested):

dom/src/main/scala/org/scalajs/dom/WorkerOptions.scala
armanbilge: Woops, I botched that :)

dom/src/main/scala/org/scalajs/dom/Worker.scala
armanbilge: Ah, good point about the `new { ... }`. Is it worth to do something similar for `SharedWorker` as well?

dom/src/main/scala/org/scalajs/dom/SharedWorker.scala
```diff
@@ -15,15 +16,23 @@ import scala.scalajs.js.annotation._
...
 @JSGlobal
-class SharedWorker(stringUrl: String, name: js.UndefOr[String] = js.native) extends AbstractWorker {
+class SharedWorker private[this] (scriptURL: String, options: js.UndefOr[String | WorkerOptions] = js.native)
+    extends AbstractWorker {
+
+  def this(scriptURL: String) = this(scriptURL, js.undefined)
+
+  def this(scriptURL: String, name: String) = this(scriptURL, js.defined(name))
+
+  def this(scriptURL: String, options: WorkerOptions) = this(scriptURL, js.defined(options))
```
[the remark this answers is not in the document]
sjrd: That works, but you're making it hard on yourself. The following is simpler:
```suggestion
class SharedWorker(scriptURL: String) extends AbstractWorker {
  def this(scriptURL: String, name: String) = this(scriptURL)

  def this(scriptURL: String, options: WorkerOptions) = this(scriptURL)
```
Note that actual arguments in the `= this(...)` call are always ignored. We just need to make the typechecker happy, but they will never be used.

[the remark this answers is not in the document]
armanbilge: Ah, I didn't know that :) this is much nicer. That makes sense.

sjrd (approved):

PR #694, opened 2022-04-13
Add Façade Types to Support ResizeObserver

chrisshafer:
Mozilla Docs : https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
Similar to [MutationObserver](https://github.com/scala-js/scala-js-dom/blob/main/dom/src/main/scala/org/scalajs/dom/MutationObserver.scala)

I wasn't sure about adding [ResizeObserverEntry.devicePixelContentBoxSize](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/devicePixelContentBoxSize), since as of now it isn't supported by Safari. I decided to leave out, but I would be happy to add it into this PR.

I also added [DOMRectReadOnly](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly), which seems a bit out of place in this PR but a few of the interfaces require that type. I'm also a bit concerned about adding it in this state due to the already implemented [DOMRect](https://github.com/scala-js/scala-js-dom/blob/main/dom/src/main/scala/org/scalajs/dom/DOMRect.scala). In the Mozilla docs it indicates that DOMRect extends the DOMRectReadOnly interface, but that doesn't really transfer well to Scala (vars vs defs). Looking for some guidance on approach here.

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
+++ b/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
@@ -1,0 +1,33 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The ResizeObserver constructor creates a new ResizeObserver object, which can be used to report changes to the
+  * content or border box of an Element or the bounding box of an SVGElement - MDN
+  *
+  * @param callback
+  *   The function called whenever an observed resize occurs. - MDN
+  */
+@js.native
+@JSGlobal
+class ResizeObserver(callback: js.Function2[js.Array[ResizeObserverEntry], ResizeObserver, _]) extends js.Object {
+
+  /** Starts observing the specified Element or SVGElement.
+    *
+    * MDN
+    */
+  def observe(target: Node, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
+
+  /** Unobserves all observed Element or SVGElement targets.
+    *
+    * MDN
+    */
+  def disconnect(): Unit = js.native
+
+  /** Ends the observing of a specified Element or SVGElement.
+    *
+    * MDN
+    */
+  def unobserve(target: Node): Unit = js.native
+}
--- a/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
+++ b/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
@@ -14,11 +14,11 @@ import scala.scalajs.js.annotation.JSGlobal
 class ResizeObserver(callback: js.Function2[js.Array[ResizeObserverEntry], ResizeObserver, _]) extends js.Object {
 
   /** Starts observing the specified Element or SVGElement. */
-  def observe(target: Node, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
+  def observe(target: Element, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
 
   /** Unobserves all observed Element or SVGElement targets. */
   def disconnect(): Unit = js.native
 
   /** Ends the observing of a specified Element or SVGElement. */
-  def unobserve(target: Node): Unit = js.native
+  def unobserve(target: Element): Unit = js.native
 }
--- a/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
+++ b/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
@@ -14,7 +14,8 @@ import scala.scalajs.js.annotation.JSGlobal
 class ResizeObserver(callback: js.Function2[js.Array[ResizeObserverEntry], ResizeObserver, _]) extends js.Object {
 
   /** Starts observing the specified Element or SVGElement. */
-  def observe(target: Element, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
+  def observe(target: Element): Unit = js.native
+  def observe(target: Element, options: ResizeObserverOptions): Unit = js.native
 
   /** Unobserves all observed Element or SVGElement targets. */
   def disconnect(): Unit = js.native
--- a/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
+++ b/dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
@@ -4,30 +4,21 @@ import scala.scalajs.js
 import scala.scalajs.js.annotation.JSGlobal
 
 /** The ResizeObserver constructor creates a new ResizeObserver object, which can be used to report changes to the
-  * content or border box of an Element or the bounding box of an SVGElement - MDN
+  * content or border box of an Element or the bounding box of an SVGElement
   *
   * @param callback
-  *   The function called whenever an observed resize occurs. - MDN
+  *   The function called whenever an observed resize occurs.
   */
 @js.native
 @JSGlobal
 class ResizeObserver(callback: js.Function2[js.Array[ResizeObserverEntry], ResizeObserver, _]) extends js.Object {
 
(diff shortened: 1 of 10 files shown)


armanbilge:
Thank you very much for the PR! To answer your questions:

> I wasn't sure about adding [ResizeObserverEntry.devicePixelContentBoxSize](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/devicePixelContentBoxSize), since as of now it isn't supported by Safari. I decided to leave out, but I would be happy to add it into this PR.

Our current policy is to accept standardized/standards-track APIs, regardless of their browser support. So feel free to add it :)

https://github.com/scala-js/scala-js-dom/blob/main/CONTRIBUTING.md#facades

> but that doesn't really transfer well to Scala (vars vs defs)

I think this is a question for @sjrd: could we do something like this and expect it to work, or something else?
```scala
trait FooReadOnly {
  def baz: Int = js.native
}

trait Foo extends FooReadOnly {
  def baz: Int = js.native
  def baz_=(x: Int): Unit = js.native
}
```

sjrd:
> could we do something like this and expect it to work, or something else?

Yes, that works.

chrisshafer:
@armanbilge 
```
    val rect = new DOMRect
    println(rect.left)
    rect.left = 10.0
    println(rect.left)
```
```
Cannot set property left of #<DOMRectReadOnly> which has only a getter
```
I'm assuming the `_=` + js.native doesn't work because there isn't a setter for the `left` field. Looks like there are only setters for x, y, width and height. This also breaks the current facade, as you get the same error. The setter works fine with height.

My guess is DomRect needs to loose the setters on `left`, `right`, `bottom`, and `top`, and we need to add `x` and `y`? Though thats a breaking API change.

In Chrome:

I tried mutating `left` with assignment in vanilla js as well which didn't work: 
```
var rect = new DOMRect();

console.log("left");
console.log(rect.left);
rect.left = 10;
console.log(rect.left);

console.log("height");
console.log(rect.height);
rect.height = 10;
console.log(rect.height);
```
```
left
0
0
height
0
10
```
```
rect.x = 10
10
rect.left
10
```

Screen Shot 2022-04-14 at 10 30 47 AM

dom/src/main/scala/org/scalajs/dom/ResizeObserverOptions.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import scala.scalajs.js
+
+@js.native
+trait ResizeObserverOptions extends js.Object {
+  var box: js.UndefOr[String] = js.native
+}
+
+/** Factory for [[ResizeObserverOptions]] objects. */
+object ResizeObserverOptions {
```
armanbilge, line 11: We've deprecated `apply` constructors. Instead we expect users to do:
```scala
new ResizeObserverOptions {
  box = "whatever"
}
```

dom/src/main/scala/org/scalajs/dom/ResizeObserverOptions.scala
```diff
@@ -0,0 +1,21 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+@js.native
+trait ResizeObserverOptions extends js.Object {
+  var box: js.UndefOr[String] = js.native
+}
```
armanbilge, line 8:
```suggestion
trait ResizeObserverOptions extends js.Object {
  var box: js.UndefOr[String] = js.undefined
}
```

dom/src/main/scala/org/scalajs/dom/ResizeObserverOptions.scala
armanbilge: Actually, seems instead of a `String` this should be a sort of enum. We encode these separately for Scala 2/3, see the `scala-2` and `scala-3` sources for examples.
```typescript
type ResizeObserverBoxOptions = "border-box" | "content-box" | "device-pixel-content-box";
```

armanbilge (commented):
Nice work! Took a first pass through with pointers about our current idioms/code style, thanks!

chrisshafer:
@armanbilge I think I addressed all your suggestions? Sorry I made a bit of a mess of the commits. I can squash when this is ready to go.

dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
```diff
@@ -0,0 +1,24 @@
...
+  *
+  * @param callback
+  *   The function called whenever an observed resize occurs.
+  */
+@js.native
+@JSGlobal
+class ResizeObserver(callback: js.Function2[js.Array[ResizeObserverEntry], ResizeObserver, _]) extends js.Object {
+
+  /** Starts observing the specified Element or SVGElement. */
+  def observe(target: Node, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
```
armanbilge, line 17:
```suggestion
  def observe(target: Element, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
```

dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  /** Unobserves all observed Element or SVGElement targets. */
+  def disconnect(): Unit = js.native
+
+  /** Ends the observing of a specified Element or SVGElement. */
+  def unobserve(target: Node): Unit = js.native
```
armanbilge, line 23:
```suggestion
  def unobserve(target: Element): Unit = js.native
```

dom/src/main/scala/org/scalajs/dom/ResizeObserverOptions.scala
```diff
@@ -0,0 +1,7 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+trait ResizeObserverOptions extends js.Object {
+  var box: js.UndefOr[String] = js.undefined
```
armanbilge, line 6:
```suggestion
  var box: js.UndefOr[ResizeObserverBoxOption] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
```diff
@@ -0,0 +1,24 @@
...
+  *
+  * @param callback
+  *   The function called whenever an observed resize occurs.
+  */
+@js.native
+@JSGlobal
+class ResizeObserver(callback: js.Function2[js.Array[ResizeObserverEntry], ResizeObserver, _]) extends js.Object {
+
+  /** Starts observing the specified Element or SVGElement. */
+  def observe(target: Element, options: js.UndefOr[ResizeObserverOptions] = js.native): Unit = js.native
```
armanbilge, line 17: Ah, sorry, one more thing. Let's write this as overloads, since this works better for type inference esp. on Scala 3. I recently learned this technique in https://github.com/scala-js/scala-js-dom/pull/691#discussion_r844629202.

```suggestion
  def observe(target: Element): Unit = js.native
  def observe(target: Element, options: ResizeObserverOptions): Unit = js.native
```

armanbilge (commented):
Looks good to me! Apologies, I did notice one last small thing :)

No need for a squash commit, personally I find keeping "messy" history to be valuable :)

Thanks for all of your work on this!

armanbilge (approved):

dom/src/main/scala/org/scalajs/dom/ResizeObserver.scala
chrisshafer: Oh interesting, thats good to know!

chrisshafer:
No problem at all. Happy to give back where I can 😄 . We use this library (and Scala.js as a whole) extensively since our platform/product is pure Scala. Appreciate what you all have done here!

Issue #191, opened 2015-12-22
Support for Shadow DOM


schmitch:
Hello, it would be great if Scala JS would also have a Shadow DOM implementation, i.e. provide a way to access Shadow DOM's.

Currently the Spec is in Draft:
https://w3c.github.io/webcomponents/spec/shadow/#widl-Element-attachShadow-ShadowRoot-ShadowRootInit-shadowRootInitDict

However there are already really great polyfills here:
http://webcomponents.org/polyfills/shadow-dom/

mseddon:
Definitely a sound addition!

If anyone wants to contribute this a few thoughts:
* It should probably live in `org.scalajs.dom.experimental.shadowroot`
* There would be a package object for `shadowroot` defining implicit coercions from `Element` onto a trait exposing `shadowRoot`, `attachShadow` etc. members.
* Since the API is experimental and likely to change in future, a comment at the top of the file linking the specification and a note with it's release date (currently 15th December) will aid tracking support longer term.

mseddon:
ShadowDOM has a dependency on https://www.w3.org/TR/cssom-view/, which we currently do not implement.

unarist:
Now, core parts of the ShadowDOM are in the DOM Standard.
https://github.com/WICG/webcomponents/issues/661

Also now can we have this as a non-experimental API?
https://caniuse.com/shadowdomv1

armanbilge:
@unarist if you're up for it will consider a PR! :)

PR #697, opened 2022-04-24
Partial façade for custom element support

peter-jerry-ye:
Create a partial façade for custom elements, i.e. only a subset of api, but should be enough to declare and register custom elements. 

Initial support for #191


dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -0,0 +1,21 @@
...
+
+/** The CustomElementRegistry interface provides methods for registering custom elements and querying registered
+  * elements. To get an instance of it, use the window.customElements property.
+  */
+@js.native
+@JSGlobal
+abstract class CustomElementRegistry extends js.Object {
+
+  /** Defines a new custom element. */
+  def define(name: String, constructor: js.Dynamic, options: CustomElementDefineOptions = js.native): Unit
```
armanbilge, line 20: Hmm, it would be nice to do something better than `js.Dynamic` here. I'm a little confused how this interface works. I see that Typescript uses:
```typescript
interface CustomElementConstructor {
    new (...params: any[]): HTMLElement;
}
```

The IDL definition is:
```idl
callback CustomElementConstructor = HTMLElement ();
```

@sjrd is there a good way to express constructors in facades? Thanks.

dom/src/main/scala/org/scalajs/dom/CustomElementDefineOptions.scala
```diff
@@ -0,0 +1,8 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** A CustomElementDefineOptions object represents additional options associated with CustomElementRegsitry.define. */
+trait CustomElementDefineOptions extends js.Object {
+  var `extends`: js.UndefOr[Boolean] = js.undefined
```
armanbilge, line 7:
```suggestion
  /** String specifying the name of a built-in element to extend. Used to create a customized built-in element. */
  var `extends`: js.UndefOr[String] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/AttachShadowRootOptions.scala
```diff
@@ -0,0 +1,9 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** A AttachShadowRootOptions object represents additional options associated with Element.attachShadow. */
+trait AttachShadowRootOptions extends js.Object {
```
armanbilge, line 6:
```suggestion
trait ShadowRootInit extends js.Object {
```

armanbilge: Also please copy docs for `mode` and `delegatesFocus` from:
https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow

dom/src/main/scala/org/scalajs/dom/CustomElementDefineOptions.scala
```diff
@@ -0,0 +1,8 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** A CustomElementDefineOptions object represents additional options associated with CustomElementRegsitry.define. */
+trait CustomElementDefineOptions extends js.Object {
```
armanbilge, line 6:
```suggestion
trait ElementDefinitionOptions extends js.Object {
```

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** The CustomElementRegistry interface provides methods for registering custom elements and querying registered
+  * elements. To get an instance of it, use the window.customElements property.
+  */
+@js.native
+@JSGlobal
+abstract class CustomElementRegistry extends js.Object {
```
armanbilge, line 17: Btw @sjrd what's a good rule of thumb for deciding to use `trait` or `abstract class`?

dom/src/main/scala/org/scalajs/dom/AttachShadowRootOptions.scala
```diff
@@ -0,0 +1,9 @@
...
+  var mode: js.UndefOr[String] = js.undefined
```
armanbilge, line 7: Would you mind creating an "enum" for this with `open` and `closed`? See the examples in the `scala-2/` and `scala-3/` sources.

armanbilge (commented):
Thank you very much for the PR, glad to see these APIs getting some attention! 😁

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
sjrd: Unfortunately `js.Dynamic` is the best we can do. That's what `js.constructorOf[C]` returns.

armanbilge: What if we did something like this?

```scala
@js.native
sealed trait CustomElementConstructor[T <: HTMLElement] extends js.Any {
  def apply(params: js.Array[Any]*): T = js.native
}

object CustomElementConstructor {
  @inline def of[T <: HTMLElement](implicit tag: ConstructorTag[T]): CustomElementConstructor[T] =
    tag.constructor.asInstanceOf[CustomElementConstructor[T]]
}
```

sjrd: The `apply` method will not be correct. So that's not a good idea.

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
sjrd: Use a class of the identifier is defined. Because then you can do `instanceof`, even though you can't instantiate it. It's nota rule of thumb. It's very clear cut.

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
armanbilge: I'm confused how it's less correct than `Dynamic` (unless the signature is just plain wrong). But I do believe this is not worth the extra type-safety (effectively, constraining the constructor to be that of an `HTMLElement`).

sjrd: The signature is wrong. It says that you can call the thing, but you can't. You would have to `new` it.

armanbilge: Hum, thanks. So (if you don't mind one more question :) how _do_ you use a signature like `CustomElementRegistry.get()` from Scala.js?
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/get

I assume the facade would be like:
```scala
/** returns the constructor for a previously-defined custom element  */
def get(name: String): js.UndefOr[js.Dynamic] = js.native
```

But then if you can't call it, how do you `new` it?

armanbilge: Nevermind, sorry, I see there is a `newInstance` method on `js.Dynamic`.
https://www.scala-js.org/api/scalajs-library/latest/scala/scalajs/js/Dynamic$.html

armanbilge: Ok surely this should work 😂 but I can believe it's not worth the type safety.

```scala
@js.native
sealed trait CustomElementConstructor[+T <: HTMLElement] extends js.Any

object CustomElementConstructor {
  @inline def of[T <: HTMLElement](implicit tag: ConstructorTag[T]): CustomElementConstructor[T] =
    tag.constructor.asInstanceOf[CustomElementConstructor[T]]

  @inline def newInstance[T](ctor: CustomElementConstructor[T])(params: js.Array[Any]*): T =
    js.Dynamic.newInstance(ctor.asInstanceOf[js.Dynamic])(params: _*).asInstanceOf[T]
}
```

dom/src/main/scala-2/org/scalajs/dom/EncapsulationMode.scala
```diff
@@ -0,0 +1,11 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+@js.native
+sealed trait EncapsulationMode extends js.Any
```
armanbilge, line 6: Sorry to nitpick these names, in most cases it makes sense to stay consistent with names used in the whatwg spec and Typescript definitions. Thanks!

```suggestion
sealed trait ShadowRootMode extends js.Any
```

PR #700, opened 2022-04-26
Add report validity facade method

987Nabil:
`setCustomValidity`s error message is shown only when `reportValidity`
is called afterwards.
See https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity
for example.


armanbilge:
Thanks for your contribution! :) Would you mind adding a scaladoc from the descriptions on the MDN page?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/reportValidity
https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reportValidity

987Nabil:
Done @armanbilge 😃

PR #704, opened 2022-05-22
Add `ServiceWorkerRegistrationOptions`

armanbilge:
https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register


dom/src/main/scala/org/scalajs/dom/ServiceWorkerRegistrationOptions.scala
```diff
@@ -0,0 +1,28 @@
...
+  * http://creativecommons.org/licenses/by-sa/2.5/
+  *
+  * Everything else is under the MIT License http://opensource.org/licenses/MIT
+  */
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** An object containing registration options. */
+trait ServiceWorkerRegistrationOptions extends js.Object {
```
japgolly, line 12: This doesn't need to be `@js.native`?

japgolly: Turns out not according to existing source 🤷 k

japgolly (approved):

PR #708, opened 2022-06-30
Create HTMLDialogElement

### LINKED ISSUE #223: Support HTMLDialogElement
Support for the Dialog element is [slowly gaining support](http://caniuse.com/#feat=dialog) among browsers and is already implemented in Chrome/Opera. Add a scalajs interface per the HTMLDialogElement spec: https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement

Lasering:
Fixes #223.

Coudn't find an example of how overloaded methods are implemented in scala-js-dom. Looked at HTMLElement.focus() but it seems the case where focus receives options was not implemented.

Ran `prePR` without problems, was expecting `prePR` to edit/generate files in `api-reports`. What is the process to commit changes there?

### DIFF
--- a/api-reports/2_12.txt
+++ b/api-reports/2_12.txt
@@ -4149,6 +4149,210 @@ HTMLDataListElement[JC] var tabIndex: Int
 HTMLDataListElement[JC] def tagName: String
 HTMLDataListElement[JC] var textContent: String
 HTMLDataListElement[JC] var title: String
+HTMLDialogElement[JC] var accessKey: String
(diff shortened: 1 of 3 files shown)


armanbilge:
Thanks for this!!

I think the file you added is missing the `.scala` extension, that's why the tooling isn't working :)

Lasering:
Now it should be proper.

Lasering:
@armanbilge shouldn't a type alias `type Dialog = HTMLDialogElement` be added here https://github.com/scala-js/scala-js-dom/blob/main/dom/src/main/scala/org/scalajs/dom/html.scala? I can make a new PR to fix it.

PR #709, opened 2022-07-04
Performance api improvements

cquiroz:
These are some improvements to the Performance API after testing it for a while

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Performance.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Performance.scala
@@ -31,7 +31,7 @@ class Performance private[this] () extends js.Object {
   /** Returns an array of [[PerformanceEntry]] objects currently present in the performance timeline for a given type */
   def getEntriesByType(entryType: String): js.Array[PerformanceEntry] = js.native
 
-  /** Is a jsonizer returning a json object representing the Performance object. */
+  /** Returns a JSON representation of the [[Performance]] object. */
   def toJSON(): js.Object = js.native
 
   @deprecated("No such API in the spec", "2.4.0")
@@ -70,7 +70,7 @@ class Performance private[this] () extends js.Object {
   /** Removes all or specific [[PerformanceMeasure]] objects from the browser's performance timeline. */
   def clearMeasures(measureName: String = js.native): Unit = js.native
 
-  /** sets the desired size of the browser's resource timing buffer which stores the "resource" performance entries. */
+  /** Sets the desired size of the browser's resource timing buffer which stores the "resource" performance entries. */
   def setResourceTimingBufferSize(maxSize: Int): Unit = js.native
 
   /** Returns a DOMHighResTimeStamp representing the amount of milliseconds elapsed since the start of the navigation,
--- a/dom/src/main/scala/org/scalajs/dom/Performance.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Performance.scala
@@ -34,13 +34,13 @@ class Performance private[this] () extends js.Object {
   /** Returns a JSON representation of the [[Performance]] object. */
   def toJSON(): js.Object = js.native
 
-  @deprecated("No such API in the spec", "2.4.0")
+  @deprecated("No such API in the spec", "2.8.0")
   def getMeasures(measureName: String = js.native): js.Dynamic = js.native
 
   /** Removes all or specific [[PerformanceMark]] objects from the browser's performance timeline. */
   def clearMarks(markName: String = js.native): Unit = js.native
 
-  @deprecated("No such API in the spec", "2.4.0")
+  @deprecated("No such API in the spec", "2.8.0")
   def getMarks(markName: String = js.native): js.Dynamic = js.native
 
   /** Removes all performance entries with an `entryType` of "resource" from the browser's performance timeline and sets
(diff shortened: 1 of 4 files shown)


dom/src/main/scala/org/scalajs/dom/Performance.scala
```diff
@@ -15,7 +15,7 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-class Performance extends js.Object {
+class Performance private[this] () extends js.Object {
```
cquiroz, line 18: You cannot create objects of type `Performance`. Is this the correct way to forbid it?

armanbilge: Yes I think we do this in other places as well.

dom/src/main/scala/org/scalajs/dom/Performance.scala
```diff
@@ -28,7 +28,7 @@ class Performance extends js.Object {
     */
   def timing: PerformanceTiming = js.native
 
-  def getEntriesByType(entryType: String): js.Dynamic = js.native
+  def getEntriesByType(entryType: String): js.Array[PerformanceEntry] = js.native
```
armanbilge, line 31: Since there are only a limited set of values that can be passed as an `entryType` I wonder if we should create a specific type for it.
https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/entryType

E.g. like `InputType`.
https://github.com/scala-js/scala-js-dom/blob/1a944e4863aaf52694e29e38fb2a174f351f2ceb/dom/src/main/scala-3/org/scalajs/dom/InputType.scala

dom/src/main/scala/org/scalajs/dom/Performance.scala
sjrd: Yes, this is correct.

dom/src/main/scala/org/scalajs/dom/Performance.scala
armanbilge: Hmm, maybe I was wrong 😅 

> As the list of supported entries varies per browser and is evolving, this property allows web developers to check which are available.

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/supportedEntryTypes

dom/src/main/scala/org/scalajs/dom/PerformanceEntry.scala
```diff
@@ -16,7 +16,7 @@ class PerformanceEntry extends js.Object {
 
   def startTime: Double = js.native
 
-  def duration: Int = js.native
+  def duration: Double = js.native
```
armanbilge, line 19: Although technically this can be a `Double`, it's time unit is milliseconds. `Int.MaxValue` milliseconds is almost 25 days, which I feel like should be more than sufficient 😅

dom/src/main/scala/org/scalajs/dom/Performance.scala
```diff
@@ -28,29 +28,49 @@ class Performance extends js.Object {
     */
   def timing: PerformanceTiming = js.native
 
-  def getEntriesByType(entryType: String): js.Dynamic = js.native
+  /** Returns an array of [[PerformanceEntry]] objects currently present in the performance timeline for a given type */
+  def getEntriesByType(entryType: String): js.Array[PerformanceEntry] = js.native
 
   /** Is a jsonizer returning a json object representing the Performance object. */
```
zetashift, line 34: nit:
```suggestion
  /** Returns a JSON representation of the [[Performance]] object. */
```

dom/src/main/scala/org/scalajs/dom/Performance.scala
```diff
@@ -28,29 +28,49 @@ class Performance extends js.Object {
...
+  def getEntriesByName(name: String, `type`: String = js.native): js.Array[PerformanceEntry] = js.native
 
-  def getEntries(): js.Dynamic = js.native
+  /** All [[PerformanceEntry]] objects currently present in the performance timeline. */
+  def getEntries(): js.Array[PerformanceEntry] = js.native
 
+  /** Removes all or specific [[PerformanceMeasure]] objects from the browser's performance timeline. */
   def clearMeasures(measureName: String = js.native): Unit = js.native
 
+  /** sets the desired size of the browser's resource timing buffer which stores the "resource" performance entries. */
```
zetashift, line 73: nit:
```suggestion
  /** Sets the desired size of the browser's resource timing buffer which stores the "resource" performance entries. */
```

dom/src/main/scala/org/scalajs/dom/PerformanceEntry.scala
```diff
@@ -9,14 +9,29 @@ package org.scalajs.dom
...
+
+  /** The name for a performance entry. It acts as an identifier, but it does not have to be unique. The value depends
+    * on the subclass.
+    */
   def name: String = js.native
 
+  /** The first timestamp recorded for this performance entry. The meaning of this property depends on the value of this
+    * entry's [[entryType]].
+    */
   def startTime: Double = js.native
```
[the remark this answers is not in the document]
armanbilge: We probably don't want an opaque type, since then you can't do numerical ops on it.

We could make a transparent type alias e.g. `type DOMHighRestTimestamp = Double` but it's a hat flip whether its documentation or obfuscation 😂

PR #715, opened 2022-07-29
Add URL-based overloads for History

armanbilge:
- https://developer.mozilla.org/en-US/docs/Web/API/History/pushState
- https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState
- https://github.com/microsoft/TypeScript/blob/1490037f178d222c11351cb8a91d1380d224f085/lib/lib.dom.d.ts#L8896-L8897


armanbilge:
Actually, I don't think this is described in the spec 🤔 

https://html.spec.whatwg.org/multipage/history.html#the-history-interface

PR #718, opened 2022-08-20
Add `History#scrollRestoration`

zetashift:
Facade for https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration

If there is anything missing or wrong, let me know!


dom/src/main/scala-3/org/scalajs/dom/ScrollRestoration.scala
```diff
@@ -0,0 +1,14 @@
...
+
+opaque type ScrollRestoration <: String = String
+
+/**
+  * see [[https://html.spec.whatwg.org/multipage/history.html#the-history-interface]] 
+  * which contains the spec for ScrollRestoration
+  */
+object ScrollRestoration {
+  val auto: ScrollRestoration = "auto"
+  val manual: ScrollRestoration = "manual"
```
armanbilge, line 13:
```suggestion
  /** The location on the page to which the user has scrolled will be restored. */
  val auto: ScrollRestoration = "auto"
  /** The location on the page is not restored. The user will have to scroll to the location manually. */
  val manual: ScrollRestoration = "manual" 
```

dom/src/main/scala/org/scalajs/dom/History.scala
```diff
@@ -71,4 +71,15 @@ class History extends js.Object {
...
+  /** The scrollRestoration property of History interface allows web applications to explicitly set default scroll
+    * restoration behavior on history navigation.
+    *
+    * Can have onne of the followings values:
+    *
+    * auto: The location on the page to which the user has scrolled will be restored.
+    *
+    * manual: The location on the page is not restored. The user will have to scroll to the location manually.
+    */
+  var scrollRestoration: ScrollRestoration = js.native
```
armanbilge, line 84:
```suggestion

  /** The `scrollRestoration` property of [[History]] interface allows web applications to explicitly set default scroll
    * restoration behavior on history navigation.
    */
  var scrollRestoration: ScrollRestoration = js.native
```

armanbilge (commented):
Thanks for your contribution! Looks good, just a couple Scaladoc nitpicks :)

dom/src/main/scala-2/org/scalajs/dom/ScrollRestoration.scala
```diff
@@ -0,0 +1,15 @@
...
+@js.native
+sealed trait ScrollRestoration extends js.Any
+
+/**
+  * see [[https://html.spec.whatwg.org/multipage/history.html#the-history-interface]] 
+  * which contains the spec for ScrollRestoration
+  */
+object ScrollRestoration {
+  val auto: ScrollRestoration = "auto".asInstanceOf[ScrollRestoration]
+  val manual: ScrollRestoration = "manual".asInstanceOf[ScrollRestoration]
```
armanbilge, line 14: Sorry, can we add the same scaladoc comments here as well! :)

armanbilge (approved):
Thank you!!

PR #719, opened 2022-08-20
Add `RequestInit#duplex`

zetashift:
Adds `duplex` to `Request` https://fetch.spec.whatwg.org/#dom-requestinit-duplex


dom/src/main/scala/org/scalajs/dom/Request.scala
```diff
@@ -47,4 +47,9 @@ class Request(input: RequestInfo, init: RequestInit = null) extends Body {
   def keepalive: Boolean = js.native
 
   def signal: AbortSignal = js.native
+
+  /** "half" is the only valid value and it is for initiating a half-duplex fetch (i.e., the user agent sends the entire
+    * request before processing the response).
+    */
+  def duplex: RequestDuplex = js.native
```
armanbilge, line 54: Hmm, I'm not sure this is the right place to add it. I think we want to add it as a `var` on `RequestInit`, not `Request` itself.

zetashift: I agree, I found it funky to put it in Request...but I also didn't know about `RequestInit` haha.

armanbilge (approved):
Thank you!! Finally we can having streaming HTTP requests!

PR #720, opened 2022-08-20
Add `Element#replaceWith`

### LINKED ISSUE #692: Missing basic DOM operation: Element.replaceWith
Pretty much supported in all browsers: https://caniuse.com/mdn-api_element_replacewith
DOM Standard: https://dom.spec.whatwg.org/#interface-childnode
MDN: https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceWith

zetashift:
https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceWith

Let me know what you think!

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Element.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Element.scala
@@ -265,4 +265,9 @@ abstract class Element extends Node with NodeSelector with ParentNode with NonDo
 
   /** Returns the open shadow root that is hosted by the element, or null if no open shadow root is present. */
   def shadowRoot: ShadowRoot = js.native
+
+  /** The Element.replaceWith() method replaces this Element in the children list of its parent with a set of Node or
+    * string objects. String objects are inserted as equivalent Text nodes.
+    */
+  def replaceWith(nodes: (Node | String)*): Unit = js.native
 }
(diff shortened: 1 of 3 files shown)


dom/src/main/scala/org/scalajs/dom/Element.scala
```diff
@@ -265,4 +265,9 @@ abstract class Element extends Node with NodeSelector with ParentNode with NonDo
 
   /** Returns the open shadow root that is hosted by the element, or null if no open shadow root is present. */
   def shadowRoot: ShadowRoot = js.native
+
+  /** The Element.replaceWith() method replaces this Element in the children list of its parent with a set of Node or
+    * string objects. String objects are inserted as equivalent Text nodes.
+    */
+  def replaceWith(nodes: (Node | String)*): Unit = js.native
```
armanbilge, line 272:
```suggestion
  /** Replaces this Element in the children list of its parent with a set of Node or
    * string objects. String objects are inserted as equivalent Text nodes.
    */
  def replaceWith(nodes: (Node | String)*): Unit = js.native
```

armanbilge (commented):
LGTM, just another scaladoc tweak 😅

armanbilge (approved):
3/3 🚀

PR #721, opened 2022-08-21
Fixup `Blob` and `File` APIs

zetashift:
I'm unsure about a lot of things about the code I wrote, for example moving the `name` to the constructor part. And should `FileOptions` be in it's separate file? Also unsure about the type of `bits`.

Anyhow let me know what you think then I'll gladly change it.


dom/src/main/scala/org/scalajs/dom/File.scala
```diff
@@ -19,8 +19,25 @@ import scala.scalajs.js.annotation._
...
+    * relative to the directory selected by the user in an <input> element with its webkitdirectory attribute set.
+    *
+    * @return
+    *   A string containing the path of the file relative to the ancestor directory the user selected.
+    */
+  def webkitRelativePath: String = js.native
+}
+
+/** An options object containing optional attributes for the file. */
+trait FileOptions extends js.Object {
```
armanbilge, line 40: Yes, our current style is to have one-class-per-file, so this can be moved to its own file :) (I _think_ it notes this in the contributing guide somewhere, otherwise we should add it.)

dom/src/main/scala/org/scalajs/dom/File.scala
```diff
@@ -19,8 +19,25 @@ import scala.scalajs.js.annotation._
...
+    */
+  def lastModified: Int = js.native
+
+  /** The File.webkitRelativePath is a read-only property that contains a string which specifies the file's path
+    * relative to the directory selected by the user in an <input> element with its webkitdirectory attribute set.
+    *
+    * @return
+    *   A string containing the path of the file relative to the ancestor directory the user selected.
+    */
+  def webkitRelativePath: String = js.native
```
armanbilge, line 36: The naming of this is really strange (I assume "webkit" is referring to the browser engine?). I do see its on MDN, but not in the linked spec ... I kinda want to avoid this one for now maybe, I'm having a hard time believing this is the final, standardized name. I'll dig into it a bit more see if I can find anything.

armanbilge (commented):
Ok ... so looks like this one is going to be a bit of a rabbit hole 😅 

The official specification is here:
https://www.w3.org/TR/FileAPI/#dfn-file

There it defines `options` as a `FilePropertyBag` which extends a `BlobPropertyBag`.

It also defines a `BlobPart` for the `fileBits`.

dom/src/main/scala/org/scalajs/dom/File.scala
```diff
@@ -19,8 +19,25 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-abstract class File extends Blob {
+abstract class File(bits: js.Iterable[Any], name: String, options: FileOptions) extends Blob {
```
armanbilge, line 22: TypeScript just uses a `js.Array` here. Does `js.Array` extend `js.Iterable`?
https://github.com/microsoft/TypeScript/blob/38e91fb12099b526e9a2cb5ad6aacfc89d5060ca/lib/lib.dom.d.ts#L5220

armanbilge: Yes it does! So I think `js.Iterable` here makes sense :)

https://www.scala-js.org/api/scalajs-library/latest/scala/scalajs/js/Array.html

armanbilge: Oh also, I don't think this should be an `abstract class`.

dom/src/main/scala/org/scalajs/dom/File.scala
armanbilge: Ok, so the docs for this reference a `webkitdirectory` property as well.
https://developer.mozilla.org/en-US/docs/Web/API/File/webkitRelativePath

Then the docs for that say:
> Though originally implemented only for WebKit-based browsers, webkitdirectory is also usable in Microsoft Edge as well as Firefox 50 and later. However, even though it has relatively broad support, it is still not standard and should not be used unless you have no alternative.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#webkitdirectory

So I'm inclined to just avoid this whole mess 😅

armanbilge: Wait, what 🤯 https://github.com/mdn/content/pull/15815/commits/561f7161feddc7f65af8d1dd86415fcf40ddd496

> Despite the name File.webkitRelativePath is standardized 

Well okay then ... let's keep it 😂

zetashift: What a rollercoaster of emotions :P

dom/src/main/scala/org/scalajs/dom/File.scala
zetashift: Ah yea the CONTRIBUTING indeed mentions:
> don't add multiple top-level types.

oops heheheh

dom/src/main/scala/org/scalajs/dom/File.scala
zetashift: I removed the abstract class and made it a class, since the `Iterable` can stay, can this be seen as resolved?

dom/src/main/scala/org/scalajs/dom/File.scala
```diff
@@ -19,8 +19,19 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-abstract class File extends Blob {
+class File(bits: js.Iterable[Any], name: String, options: FilePropertyBag) extends Blob {
```
armanbilge, line 22: Since `options` is optional we should add a second constructor with just `bits` and `name`.

And also `bits` should use `BlobPart` instead of `Any` :)

dom/src/main/scala/org/scalajs/dom/File.scala
armanbilge: Yup, thanks! But a couple more comments about it below :)

dom/src/main/scala/org/scalajs/dom/File.scala
zetashift: BlobPart isn't defined yet, according to the spec it looks like this:

> typedef ([BufferSource](https://heycam.github.io/webidl/#BufferSource) or [Blob](https://www.w3.org/TR/FileAPI/#dfn-Blob) or [USVString](https://heycam.github.io/webidl/#idl-USVString)) BlobPart;

In `Blob.scala` blobParts is defined as:
```
blobParts: js.Array[js.Any] = js.native
```
I guess we want a dedicated BlobParts type?

zetashift: Maybe something like this:

```scala
enum BlobPart {
  case BufferSource 
  case Blob 
  case USVString
} 

```

armanbilge: We can define it in the package object:
```scala
type BlobPart = BufferSource | Blob | String
```

dom/src/main/scala/org/scalajs/dom/File.scala
```diff
@@ -19,8 +19,20 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-abstract class File extends Blob {
+class File(bits: js.Iterable[BlobPart], name: String, options: FilePropertyBag = js.native)
+    extends Blob(bits, options) {
 
-  /** Returns the name of the file. For security reasons, the path is excluded from this property. */
-  def name: String = js.native
```
armanbilge, line 25: Why did this get removed?

zetashift: My editor showed me an error that `def name` was doubly defined, because of the `name` thing in the constructor...so I removed it and that error went away hehe.

armanbilge: Hmm, well it also takes it away from the public API 😅  maybe just rename the constructor one instead? e.g. `_name`.

zetashift: The error being:
```
[error] -- [E120] Naming Error: /dev/scala/scala-js-dom/dom/src/main/scala/org/scalajs/dom/File.scala:26:6 
[error] 26 |  def name: String = js.native
[error]    |      ^
[error]    |      Double definition:
[error]    |      private[this] val name: String in class File at line 22 and
[error]    |      def name: String in class File at line 26
```

zetashift: Ah no I just posted that while `prePR` was running :P

zetashift:
I think I did an oopsie with Git

dom/src/main/scala/org/scalajs/dom/File.scala
```diff
@@ -19,8 +19,23 @@ import scala.scalajs.js.annotation._
...
+    extends Blob(bits, options) {
 
   /** Returns the name of the file. For security reasons, the path is excluded from this property. */
   def name: String = js.native
+
+  /** The File.lastModified read-only property provides the last modified date of the file as the number of milliseconds
+    * since the Unix epoch (January 1, 1970 at midnight). Files without a known last modified date return the current
+    * date.
+    */
+  def lastModified: Int = js.native
```
armanbilge, line 32: We should use a `Double` here. `Int` doesn't nearly have enough range 😛 
```
scala> import scala.concurrent.duration._
import scala.concurrent.duration._

scala> Int.MaxValue.millis.toDays
val res0: Long = 24
```

dom/src/main/scala/org/scalajs/dom/FilePropertyBag.scala
```diff
@@ -0,0 +1,8 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** An options object containing optional attributes for the file. */
+trait FilePropertyBag extends BlobPropertyBag {
+  var lastModified: js.UndefOr[Int] = js.undefined
```
armanbilge, line 7: Same here :)

dom/src/main/scala/org/scalajs/dom/File.scala
zetashift: First attempt had Double and then I changed it because I swore that I saw `Int` being used somewhere ><.

Oh well, easy fix atleast

api-reports/2_13.txt
```diff
@@ -1946,15 +1946,20 @@ FetchEventInit[JT] var request: js.UndefOr[Request]
...
 File[JC] def size: Double
 File[JC] def slice(start: Double?, end: Double?, contentType: String?): Blob
 File[JC] def stream(): ReadableStream[Uint8Array]
 File[JC] def text(): js.Promise[String]
 File[JC] def `type`: String
+File[JC] def webkitRelativePath: String
 FileList[JC] @JSBracketAccess def apply(index: Int): T
 FileList[JC] def item(index: Int): File
 FileList[JC] def length: Int
+FilePropertyBag[JT] var endings: js.UndefOr[String]
```
armanbilge, line 1960: Oops, sorry to be annoying but looks like `endings` is an "enum" `EndingType`.
https://w3c.github.io/FileAPI/#enumdef-endingtype

zetashift: Not at all annoying, I like feedback like this!

armanbilge (approved):
Ok, I think that's everything. Good work on this, nice fix up of these APIs!

zetashift:
Anytime! 
I have `Image` half finished I think, but this week I'm spending more time in the learning-backend-development realm, so I hopefully I get that PR done next week.

dom/src/main/scala/org/scalajs/dom/File.scala
zetashift: https://github.com/softwaremill/sttp/blob/master/core/src/main/scalajs/sttp/client3/dom/experimental/File.scala#L34 this was why I changed the `Double` to an `Int`, the sttp example used it as well haha

Issue #706, opened 2022-06-21
Upgrade to Scala 3.1.3?


armanbilge:
Notably, it includes https://github.com/lampepfl/dotty/pull/14318. The bug it fixes seems to cause all sorts of problems when trying to do things with sourcemaps.

I realize this lib is mostly facade where line numbers may not matter, but there are a few places where it could.

Edit: note that this would force an update to Scala.js 1.7 as well.

armanbilge:
I will move ahead with this.
1. The official advice for Scala 3 is to upgrade.
    https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html
3. Scala.js 1.7.0 was released over a year ago.
    https://www.scala-js.org/news/2021/08/04/announcing-scalajs-1.7.0/

PR #723, opened 2022-08-22
Update to Scala.js 1.7.1, Scala 2.12.15, 3.1.3

armanbilge:
Closes https://github.com/scala-js/scala-js-dom/issues/706.


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
 
   object Ver {
     val scala211 = "2.11.12"
-    val scala212 = "2.12.14"
+    val scala212 = "2.12.15"
```
armanbilge, line 9: This brings us in line with 1.7.1.
https://www.scala-js.org/news/2021/10/07/announcing-scalajs-1.7.1/

Issue #621, opened 2021-10-30
Create facade for `Image`


armanbilge:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image

ScalablyTyped _thought_ we had this, so we may as well add it :)

zetashift:
So far I've got this:

```scala
/** Image creates a new HTMLImageElement instance. It is functionally equivalent to document.createElement('img').
  *
  * @param width
  *   The width of the image (i.e., the value for the width attribute).
  * @param height
  *   The height of the image (i.e., the value for the height attribute).
  */
class Image(width: Int, height: Int) extends HTMLImageElement
```
`
However in my exploration I see that the `HTMLImageElement` is missing some properties, for example:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset

Want me to add these as well in the same PR?

armanbilge:
Sure, if you're up for it! :)

PR #725, opened 2022-09-01
Add `Image` facade and additional facades for `HTMLImageElement` #621

zetashift:
Another good first issue!

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
@@ -67,8 +67,8 @@ abstract class HTMLImageElement extends HTMLElement {
   var onload: js.Function1[Event, _] = js.native
 
   /** Indicates the x-coordinate of the <img> element's left border edge relative to the root element's origin. */
-  def x: Double = js.native
+  def x: Int = js.native
 
   /** Indicates the y-coordinate of the <img> element's top border edge relative to the root element's origin. */
-  def y: Double = js.native
+  def y: Int = js.native
 }
--- a/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
@@ -31,7 +31,7 @@ abstract class HTMLImageElement extends HTMLElement {
   /** A string whose value provides a hint to the user agent on how to handle the loading of the image which is
     * currently outside the window's visual viewport.
     */
-  var loading: ImageLoadingMode = js.native
+  var loading: LazyLoadingState = js.native
 
   /** Reflects the src HTML attribute, containing the URL of the image. */
   var src: String = js.native
@@ -67,8 +67,8 @@ abstract class HTMLImageElement extends HTMLElement {
   var onload: js.Function1[Event, _] = js.native
 
   /** Indicates the x-coordinate of the <img> element's left border edge relative to the root element's origin. */
-  def x: Long = js.native
+  def x: Double = js.native
 
   /** Indicates the y-coordinate of the <img> element's top border edge relative to the root element's origin. */
-  def y: Long = js.native
+  def y: Double = js.native
 }
--- a/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
@@ -28,11 +28,16 @@ abstract class HTMLImageElement extends HTMLElement {
   /** Allows you to specify the layout width of the image for each of a list of media conditions. */
   var sizes: String = js.native
 
+  /** A string whose value provides a hint to the user agent on how to handle the loading of the image which is
+    * currently outside the window's visual viewport.
+    */
+  var loading: String = js.native
+
   /** Reflects the src HTML attribute, containing the URL of the image. */
   var src: String = js.native
 
-  /** srcset is a string which identifies one or more image candidate strings, separated using commas (,) each
-    * specifying image resources to use under given circumstances.
+  /** Identifies one or more image candidate strings, separated using commas (,) each specifying image resources to use
+    * under given circumstances.
     *
     * Each image candidate string contains an image URL and an optional width or pixel density descriptor that indicates
     * the conditions under which that candidate should be used instead of the image specified by the src property.
@@ -60,4 +65,10 @@ abstract class HTMLImageElement extends HTMLElement {
   def complete: Boolean = js.native
 
   var onload: js.Function1[Event, _] = js.native
+
+  /** Indicates the x-coordinate of the <img> element's left border edge relative to the root element's origin. */
+  def x: Long = js.native
+
+  /** Indicates the y-coordinate of the <img> element's top border edge relative to the root element's origin. */
+  def y: Long = js.native
 }
--- a/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
@@ -25,8 +25,22 @@ abstract class HTMLImageElement extends HTMLElement {
   /** Reflects the alt HTML attribute, indicating fallback context for the image. */
   var alt: String = js.native
 
+  /** Allows you to specify the layout width of the image for each of a list of media conditions. */
+  var sizes: String = js.native
+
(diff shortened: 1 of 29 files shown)


dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
```diff
@@ -25,9 +25,28 @@ abstract class HTMLImageElement extends HTMLElement {
   /** Reflects the alt HTML attribute, indicating fallback context for the image. */
   var alt: String = js.native
 
+  /** Allows you to specify the layout width of the image for each of a list of media conditions. */
+  var sizes: String = js.native
+
+  /** A string whose value provides a hint to the user agent on how to handle the loading of the image which is
+    * currently outside the window's visual viewport.
+    */
+  var loading: String = js.native
```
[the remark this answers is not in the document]
armanbilge: Good question. Seems like it should be?

> The loading IDL attribute must [reflect](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect) the [loading](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-loading) content attribute, [limited to only known values](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#limited-to-only-known-values).

https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-loading

dom/src/main/scala/org/scalajs/dom/Image.scala
```diff
@@ -0,0 +1,10 @@
+package org.scalajs.dom
+
+/** Image creates a new HTMLImageElement instance. It is functionally equivalent to document.createElement('img').
+  *
+  * @param width
+  *   The width of the image (i.e., the value for the width attribute).
+  * @param height
+  *   The height of the image (i.e., the value for the height attribute).
+  */
+class Image(width: Int, height: Int) extends HTMLImageElement
```
armanbilge, line 10: It looks like `width` and `height` are optional, so we should add some additional constructors.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image

dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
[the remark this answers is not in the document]
armanbilge: Thanks! This should be `ImageLoadingMode` now right?

dom/src/main/scala/org/scalajs/dom/Image.scala
zetashift: Ah, I see, I think if they aren't specified they default to 0: https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-width but I'm not entirely sure

dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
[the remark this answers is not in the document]
zetashift: oopsy-daisy forgot about that :P

dom/src/main/scala/org/scalajs/dom/Image.scala
armanbilge: It actually doesn't matter what you put there; that code will never be executed since it's a facade. We just need something to make the compiler happy (confusing right!).

zetashift: oh yea, that's hard to wrap my head around, I thought the compiler never lies to me :P

armanbilge (commented):
Overall LGTM! Thanks :)

The only thing I'm going think on is `ImageLoadingMode`. Unless we can find a more "official" name for it somewhere, "mode" seems fine to me.
- MDN calls it a "hint"
    https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading
- Typescript calls it a "policy"
    https://github.com/microsoft/TypeScript/blob/4110b80fbb24f2ac6dd284ef3511ca69ac6db517/lib/lib.dom.d.ts#L6869-L6870
- WHATWG doesn't seem to have a name for it at all.
  https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-loading

zetashift:
Actually finding a name for `ImageLoadingMode` was the hardest part hahaha. I looked at the specs and it didn't have a name for it.
MDN also feels very unofficial with the "hint" so I used `mode` just to get started.
Policy sounds okay, but they also just drop that in a comment, the type itself doesn't really have a name...

I was also wondering if it should be `Image` and not `HTMLImageLoadingMode` :P. I don't have a real preference, `hint` could probably be the most familiar to people because MDN describes it like that.

armanbilge:
Ha yeah I thought as much 😂 

Looking at the existing enums we have, both "mode" and "policy" are common. So I think either is a fine choice. No "hint".
https://github.com/scala-js/scala-js-dom/tree/main/dom/src/main/scala-3/org/scalajs/dom

Now it's just a philosophical question of what the difference between those is 🤔

zetashift:
I guess "mode" sounds like it can only be certain cases, whereas policy sounds more like "something that just should be followed but might not be"?

Naming is hard :P

armanbilge:
lol, well what about this? They seem to call it a "state"
https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes

The other thing that's notable is that it's not specific to `Image`, seems that `iframe` uses it too. So I wonder if it should actually be shared by both 🤔

zetashift:
Yuck, I don't find state fitting, but spec is spec...

Can rename to something like `LoadingState`

armanbilge:
Ok ... coming back to this ... 😅  

I think I'm in favor of `LazyLoadingState`. Since it's the "lazy loading attribute", and "Some attributes, called enumerated attributes, take on a finite set of states". So that seems like the specciest name.

dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
```diff
@@ -46,4 +65,10 @@ abstract class HTMLImageElement extends HTMLElement {
   def complete: Boolean = js.native
 
   var onload: js.Function1[Event, _] = js.native
+
+  /** Indicates the x-coordinate of the <img> element's left border edge relative to the root element's origin. */
+  def x: Long = js.native
+
+  /** Indicates the y-coordinate of the <img> element's top border edge relative to the root element's origin. */
+  def y: Long = js.native
```
armanbilge, line 73: Woops, we can't use `Long` in facades (since there are no 64 big integers in JS). These should either be `Int` or `Double` if they can be larger than a 32-bit `Int`.

zetashift: I wonder why I wrote Long there?
I'll update my PR soon-ish! :D

armanbilge: And I wonder why I basically said "LGTM except bikeshedding" 😂 

Thanks!

zetashift: Because naming things is hard!

zetashift:
@armanbilge I renamed it to LazyLoadingState and changed the Longs to Doubles!

dom/src/main/scala/org/scalajs/dom/HTMLImageElement.scala
```diff
@@ -46,4 +65,10 @@ abstract class HTMLImageElement extends HTMLElement {
   def complete: Boolean = js.native
 
   var onload: js.Function1[Event, _] = js.native
+
+  /** Indicates the x-coordinate of the <img> element's left border edge relative to the root element's origin. */
+  def x: Double = js.native
+
+  /** Indicates the y-coordinate of the <img> element's top border edge relative to the root element's origin. */
+  def y: Double = js.native
```
armanbilge, line 73: Sorry, more bikeshed: can these realistically exceed `Int.MaxValue`? We are currently using `Int` for `width`, `height` etc. I genuinely don't know how huge monitors get these days 😂  I'm curious if there's a specific reason why you picked `Double`. Numeric types are really annoying 😕

zetashift: I have no idea anymore why I put a double there...I'll put Int there🤣

armanbilge (approved):
Great work on this, thanks for reminding me about it! 🚲 🏚️

PR #736, opened 2022-11-02
Fixup `Clipboard` to use added `ClipboardItem`

zetashift:
I probably missed something, let me know what can be improved!

### DIFF
--- a/dom/src/main/scala-3/org/scalajs/dom/PresentationStyle.scala
+++ b/dom/src/main/scala-3/org/scalajs/dom/PresentationStyle.scala
@@ -1,0 +1,15 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/**
+  * Helps distinguish whether apps "pasting" a clipboard item should insert the contents of an appropriate representation inline at the point of paste or if it should be treated as an attachment.
+  * See [[https://w3c.github.io/clipboard-apis/#enumdef-presentationstyle PresentationStyle enum]]
+  */
+opaque type PresentationStyle <: String = String
+
+object PresentationStyle {
+  val unspecified: PresentationStyle = "unspecified"
+  val inline: PresentationStyle = "inline"
+  val attachment: PresentationStyle = "attachment"
+}
--- a/dom/src/main/scala/org/scalajs/dom/Clipboard.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Clipboard.scala
@@ -25,7 +25,7 @@ trait Clipboard extends EventTarget {
     *
     * To read from the clipboard, you must first have the "clipboard-read" permission.
     */
-  def read(): js.Promise[DataTransfer] = js.native
+  def read(): js.Promise[ClipboardItem] = js.native
 
   /** The readText() method returns a Promise which resolves with a copy of the textual contents of the system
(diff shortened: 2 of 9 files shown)


dom/src/main/scala/org/scalajs/dom/Clipboard.scala
```diff
@@ -25,7 +25,7 @@ trait Clipboard extends EventTarget {
     *
     * To read from the clipboard, you must first have the "clipboard-read" permission.
     */
-  def read(): js.Promise[DataTransfer] = js.native
+  def read(): js.Promise[ClipboardItem] = js.native
```
armanbilge, line 28: Should this be an array?
```suggestion
  def read(): js.Promise[js.Array[ClipboardItem]] = js.native
```

dom/src/main/scala/org/scalajs/dom/Clipboard.scala
```diff
@@ -38,7 +38,7 @@ trait Clipboard extends EventTarget {
     * Before you can write to the clipboard, you need to use the Permissions API to get the "clipboard-write"
     * permission.
     */
-  def write(data: DataTransfer): js.Promise[Unit] = js.native
+  def write(data: ClipboardItem): js.Promise[Unit] = js.native
```
armanbilge, line 41: Ditto.
```suggestion
  def write(data: js.Array[ClipboardItem]): js.Promise[Unit] = js.native
```

dom/src/main/scala/org/scalajs/dom/Clipboard.scala
zetashift: Ah you're right! Didn't read MDN correctly

dom/src/main/scala-3/org/scalajs/dom/PresentationStyle.scala
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+  * Helps distinguish whether apps "pasting" a clipboard item should insert the contents of an appropriate representation inline at the point of paste or if it should be treated as an attachment.
+  * See [[https://w3c.github.io/clipboard-apis/#enumdef-presentationstyle PresentationStyle enum]]
+  */
+opaque type PresentationStyle <: String = String
+
+object PresentationStyle {
+  val unspecified: PresentationStyle = "unspecified"
+  val inline: PresentationStyle = "inline"
```
armanbilge, line 13: I guess this is a so-called "soft" keyword in Scala 3 so it doesn't need to be `` `inline` ``.

armanbilge (approved):
I think it's good! Hooray now people can actually use this API 😂

dom/src/main/scala-3/org/scalajs/dom/PresentationStyle.scala
zetashift: huh I forgot that `inline` is a thing in Scala...want me to change it to `inline` to be sure?

armanbilge: The compiler is happy, so I'm happy :)

Further reading:
- https://docs.scala-lang.org/scala3/reference/soft-modifier.html

Issue #729, opened 2022-10-19
`Storage#length` should be a `def`


armanbilge:
https://github.com/scala-js/scala-js-dom/blob/dde2aba0a9aa7ee85a7242e875f12bc54509ffc9/dom/src/main/scala/org/scalajs/dom/Storage.scala#L19

PR #737, opened 2022-11-02
Change Storage#length to a def instead of var #729

zetashift:
My feelings about this PR
image


armanbilge (approved):
Thank you so much, this is a big help 😃

PR #744, opened 2022-11-05
Remove stale todo

armanbilge:
The spec uses `js.Any` there anyway and I see no real compelling reason to change it. The "`JSON`" type is just a POJSO plain-old-javascript-object.

https://fetch.spec.whatwg.org/#body

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Body.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Body.scala
@@ -22,9 +22,7 @@ trait Body extends js.Object {
   /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a FormData object. */
   def formData(): js.Promise[FormData] = js.native
 
-  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a JSON object. //todo:
-    * define the JSON type, and return a Promise[JSON] as per spec
-    */
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a JSON object. */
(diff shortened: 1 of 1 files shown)


dom/src/main/scala/org/scalajs/dom/Body.scala
```diff
@@ -22,9 +22,7 @@ trait Body extends js.Object {
   /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a FormData object. */
   def formData(): js.Promise[FormData] = js.native
 
-  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a JSON object. //todo:
-    * define the JSON type, and return a Promise[JSON] as per spec
-    */
+  /** Takes a Response stream and reads it to completion. It returns a promise that resolves with a JSON object. */
   def json(): js.Promise[js.Any] = js.native
```
zetashift, line 26: Why not `js.Object` then atleast? Tho the spec says indeed says `any` so either way `Promise[JSON]` is unnecessary.

zetashift (approved):

dom/src/main/scala/org/scalajs/dom/Body.scala
armanbilge: Good question, it's because valid JSON may parse to something that is not a `js.Obect`.
```
> JSON.parse('123')
123
```

zetashift: Ah right! Thank you.
Then indeed `js.Any` is the way to go.

PR #752, opened 2022-12-03
Add `Element#closest`

DanielMoss:
Hi! This just adds a facade. ~You might spot that when I ran `prePR` some of the API report formatting changed. I'm not sure why this is, but I'm happy to manually pluck those changes out if you like.~ (done)

[MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest).


zetashift (approved):
Oh wow, that formatting thing is something I haven't seen before! 

The actual code in `Element.scala` looks good to go, it's a cool utility helper as well.

PR #753, opened 2022-12-16
Add alias to HTMLDialogElement in html object.

Lasering:
Add alias to HTMLDialogElement.


zetashift:
The CI seems to fail because, `sbt prePR` command wasn't run, and no new api-reports.txt were commited.
Can you try running that command? More information here: https://github.com/scala-js/scala-js-dom/blob/main/CONTRIBUTING.md#submitting-a-pr

armanbilge (approved):
Thanks!

Issue #746, opened 2022-11-12
`Location#origin` should return a `String`


armanbilge:
There's a reason it doesn't, but it's not a particularly good one.

https://github.com/scala-js/scala-js-dom/blob/410f4392df7ea815b640cef10db3e072a6c388c8/dom/src/main/scala/org/scalajs/dom/Location.scala#L47-L49

https://developer.mozilla.org/en-US/docs/Web/API/Location/origin#browser_compatibility

armanbilge:
Oh and also this `flag` parameter is not part of the spec, just a firefox thing ...
https://github.com/scala-js/scala-js-dom/blob/410f4392df7ea815b640cef10db3e072a6c388c8/dom/src/main/scala/org/scalajs/dom/Location.scala#L55

PR #760, opened 2023-01-27
Fixup `Location#origin` and `Location#reload`

zetashift:
Fixes https://github.com/scala-js/scala-js-dom/issues/746


zetashift:
Oops, I failed at git-fu, fixing it now :P

PR #761, opened 2023-01-27
Fixup Location#origin and Location#reload 

zetashift:
Fixes https://github.com/scala-js/scala-js-dom/issues/746

(this time not building off from the wrong branch ;P)

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Location.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Location.scala
@@ -46,13 +46,13 @@ trait Location extends js.Object {
     *
     * This property also does not exist consistently on IE, even as new as IE11, hence it must be UndefOr.
     */
-  def origin: js.UndefOr[String] = js.native
+  def origin: String = js.native
 
   /** The Location.reload()method Reloads the resource from the current URL. Its optional unique parameter is a Boolean,
     * which, when it is true, causes the page to always be reloaded from the server. If it is false or not specified,
     * the browser may reload the page from its cache.
     */
-  def reload(flag: Boolean = js.native): Unit = js.native
+  def reload: Unit = js.native
 
   /** The Location.replace()method replaces the current resource with the one at the provided URL. The difference from
     * the assign() method is that after using replace() the current page will not be saved in session History, meaning
--- a/api-reports/2_12.txt
+++ b/api-reports/2_12.txt
@@ -15482,11 +15482,11 @@ Location[JT] var hash: String
 Location[JT] var host: String
 Location[JT] var hostname: String
 Location[JT] var href: String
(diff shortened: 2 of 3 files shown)


dom/src/main/scala/org/scalajs/dom/Location.scala
```diff
@@ -46,13 +46,13 @@ trait Location extends js.Object {
     *
     * This property also does not exist consistently on IE, even as new as IE11, hence it must be UndefOr.
```
armanbilge, line 47: We should update the doc 😛 

I'd be fine to just drop this sentence, at this point in time we are not consistently providing information about which browsers implement what.

https://caniuse.com/mdn-api_origin

dom/src/main/scala/org/scalajs/dom/Location.scala
```diff
@@ -46,13 +46,13 @@ trait Location extends js.Object {
...
     */
-  def origin: js.UndefOr[String] = js.native
+  def origin: String = js.native
 
   /** The Location.reload()method Reloads the resource from the current URL. Its optional unique parameter is a Boolean,
     * which, when it is true, causes the page to always be reloaded from the server. If it is false or not specified,
     * the browser may reload the page from its cache.
     */
-  def reload(flag: Boolean = js.native): Unit = js.native
+  def reload: Unit = js.native
```
armanbilge, line 55: Same here :)

https://developer.mozilla.org/en-US/docs/Web/API/Location/reload

dom/src/main/scala/org/scalajs/dom/Location.scala
```diff
@@ -43,16 +43,14 @@ trait Location extends js.Object {
...
 
-  /** The Location.reload()method Reloads the resource from the current URL. Its optional unique parameter is a Boolean,
-    * which, when it is true, causes the page to always be reloaded from the server. If it is false or not specified,
-    * the browser may reload the page from its cache.
+  /** Reloads the resource from the current URL, like the Refresh button. The reload may be blocked and a SECURITY_ERROR
+    * DOMException thrown. This happens if the origin of the script calling location.reload() differs from the origin of
+    * the page that owns the Location object.
     */
-  def reload(flag: Boolean = js.native): Unit = js.native
+  def reload: Unit = js.native
```
armanbilge, line 53: We still need the `()` since its a method
```suggestion
  def reload(): Unit = js.native
```

https://www.scala-js.org/doc/interoperability/facade-types.html#remarks

zetashift: What a rookie mistake by me :P, I really thought it didn't need it

armanbilge (approved):
Thanks!

Issue #735, opened 2022-10-27
Support Web Locks API


armanbilge:
https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API

PR #762, opened 2023-01-28
Add Web Lock APIs

armanbilge:
Closes https://github.com/scala-js/scala-js-dom/issues/735.

- https://developer.mozilla.org/en-US/docs/Web/API/LockManager
- https://developer.mozilla.org/en-US/docs/Web/API/Lock
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator/locks

Names for the "anonymous" types were taken from the TypeScript definitions.


dom/src/main/scala/org/scalajs/dom/Lock.scala
```diff
@@ -0,0 +1,23 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[Lock]] interface of the Web Locks API provides the name and mode of a lock. This may be a newly requested lock
+  * that is received in the callback to [[LockManager.request(name:String,callback:* LockManager.request]], or a record
```
armanbilge, line 7: I'm ... afraid this bullshit will set a bad precedent 😂

zetashift: Could be worse no, I don't think there are a lot of these kind of elaborate scaladoc comments.

All to please the tooling...I hope!

armanbilge: I think it's really nice to have scaladoc that you can click on stuff, and it will link you properly. But yeah, linking for over-loaded methods is a PITA.

dom/.scala-build/.scalafmt.conf
```diff
@@ -0,0 +1,28 @@
+version = "3.7.1"
```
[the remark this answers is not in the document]
armanbilge: Oh lol where did this come from 🤔

zetashift (approved):
LGTM!

Issue #738, opened 2022-11-02
Refresh `DataTransfer` API


armanbilge:
I think there are some changes. Also it can use the new `FrozenArray` type.

https://github.com/scala-js/scala-js-dom/blob/714813e31cf5838d9844141f7bc1b90275f79fa3/dom/src/main/scala/org/scalajs/dom/DataTransfer.scala

https://html.spec.whatwg.org/multipage/dnd.html#datatransfer

zetashift:
I have a question concerning `effectAllowed` and `dropEffect`. They return enums form the spec

> The dropEffect attribute controls the drag-and-drop feedback that the user is given during a drag-and-drop operation. When the [DataTransfer](https://html.spec.whatwg.org/multipage/dnd.html#datatransfer) object is created, the [dropEffect](https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-dropeffect) attribute is set to a string value. On getting, it must return its current value. On setting, if the new value is one of "none", "copy", "link", or "move", then the attribute's current value must be set to the new value. Other values must be ignored.

> The effectAllowed attribute is used in the drag-and-drop processing model to initialize the [dropEffect](https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-dropeffect) attribute during the [dragenter](https://html.spec.whatwg.org/multipage/dnd.html#event-dnd-dragenter) and [dragover](https://html.spec.whatwg.org/multipage/dnd.html#event-dnd-dragover) events. When the [DataTransfer](https://html.spec.whatwg.org/multipage/dnd.html#datatransfer) object is created, the [effectAllowed](https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-effectallowed) attribute is set to a string value. On getting, it must return its current value. On setting, if [drag data store](https://html.spec.whatwg.org/multipage/dnd.html#drag-data-store)'s [mode](https://html.spec.whatwg.org/multipage/dnd.html#drag-data-store-mode) is the [read/write mode](https://html.spec.whatwg.org/multipage/dnd.html#concept-dnd-rw) and the new value is one of "none", "copy", "copyLink", "copyMove", "link", "linkMove", "move", "all", or "uninitialized", then the attribute's current value must be set to the new value. Otherwise it must be left unchanged.

But how do I name these things lol, `EffectAllowedTypes` and `DropEffectTypes`? or `Kinds` instead of types?

armanbilge:
[...]

PR #763, opened 2023-01-30
Refresh DataTransfer API

zetashift:
Should resolve https://github.com/scala-js/scala-js-dom/issues/738


dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
```diff
@@ -0,0 +1,39 @@
...
+  /** Removes all the entries in the drag data store. */
+  def clear(): Unit = js.native
+
+}
+
+@js.native
+object DataTransferItemList extends DataTransferItemList {
+  /** Returns the DataTransferItem object representing the indexth entry in the drag data store. */
+  def apply(index: Int): DataTransferItem = js.native
+}
```
zetashift, line 39: I'm having a hard time getting this getting to compile/typecheck, any tips?

armanbilge: Hmm, maybe it should be similar to `FrozenArray`.
https://github.com/scala-js/scala-js-dom/blob/21d6badf500435b66f645e2114d9f623afc9415f/dom/src/main/scala/org/scalajs/dom/FrozenArray.scala#L17-L23

armanbilge: but probably shouldn't extend `js.Iterable`. and `length` should be a def since it can be changing

dom/src/main/scala/org/scalajs/dom/DataTransfer.scala
```diff
@@ -19,16 +19,11 @@ trait DataTransfer extends js.Object {
...
     */
-  var effectAllowed: String = js.native
+  var effectAllowed: EffectAllowedValue = js.native
 
-  /** The actual effect that will be used, and should always be one of the possible values of effectAllowed.
-    *
-    * See [[DragEffect]] for possible values.
-    */
-  var dropEffect: String = js.native
+  /** The actual effect that will be used, and should always be one of the possible values of effectAllowed. */
```
zetashift, line 25: I guess this scaladoc should also be improved ><

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
armanbilge: Oh, and also it should be a `class` with `@JSGlobal` annotation 😂

dom/src/main/scala/org/scalajs/dom/DataTransferItem.scala
```diff
@@ -0,0 +1,26 @@
...
+trait DataTransferItem extends js.Object {
+
+  /** Returns the drag data item kind, one of: "string", "file". */
+  def kind: DragDataItemKind = js.native
+
+  /** Returns the drag data item type string. */
+  def `type`: String = js.native
+
+  /** Invokes the callback with the string data as the argument, if the drag data item kind is text. */
+  def getAsString(callback: () => String) = js.native
```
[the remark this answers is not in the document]
armanbilge: I think that should be a `js.Function1[String, Unit]` —its a JS callback function, that takes a `String`.

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
zetashift: I'm extending `js.Object` and `length` is a `def`!

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** A list of [[DataTransferItem]] objects representing items being dragged. During a drag operation, each DragEvent has
+  * a dataTransfer property and that property is a DataTransferItemList.
+  */
+@js.native
+@JSGlobal
+class DataTransferItemList extends js.Object {
```
armanbilge, line 17: Since the constructor cannot be used publicly.

```suggestion
class DataTransferItemList private[this] extends js.Object {
```

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Returns the number of items in the drag data store. */
+  def length: Int = js.native
+
+  /** Adds a new entry for the given data to the drag data store. If the data is plain text then a type string has to be
+    * provided also.
+    */
+  def add(data: String, `type`: String): Unit = js.native
```
armanbilge, line 25: I wonder if `type` should be a `MIMEType`. But our list is woefully incomplete 😅 
https://github.com/scala-js/scala-js-dom/blob/21d6badf500435b66f645e2114d9f623afc9415f/dom/src/main/scala-3/org/scalajs/dom/MIMEType.scala#L5-L15

dom/src/main/scala/org/scalajs/dom/DataTransferItem.scala
```diff
@@ -0,0 +1,26 @@
...
+  *
+  * Everything else is under the MIT License http://opensource.org/licenses/MIT
+  */
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** Each DataTransferItem object is associated with a [[DataTransfer]] object. */
+@js.native
+trait DataTransferItem extends js.Object {
```
armanbilge, line 13: It appears to be a `class` with a non-public constructor.

```suggestion
class DataTransferItem private[this] extends js.Object {
```

dom/src/main/scala/org/scalajs/dom/DataTransferItem.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  /** Returns the drag data item kind, one of: "string", "file". */
+  def kind: DragDataItemKind = js.native
+
+  /** Returns the drag data item type string. */
+  def `type`: String = js.native
```
armanbilge, line 19: Same question, if this should return a `MIMEType`.

dom/src/main/scala/org/scalajs/dom/DataTransferItem.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  /** Invokes the callback with the string data as the argument, if the drag data item kind is text. */
+  def getAsString(callback: Function1[String, Unit]): Unit = js.native
+
+  /** Returns a File object, if the drag data item kind is File. */
+  def getAsFile(): js.UndefOr[File] = js.native
```
armanbilge, line 25: Hm, I see this can return `null`, but can it return `undefined` ?

zetashift: I have no idea really, I just used `UndefOr` because it's what I know ;p, is there a `NullOr[???]`?

armanbilge: No, in the case of `null` we just write `: File` :)

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
zetashift: oh wow that needs to be updated!

dom/src/main/scala/org/scalajs/dom/DataTransferItem.scala
zetashift: I think we can just make this return a string (as well as the other one). No need to complicate things and the spec says string, and TS dom bindings use string as well

armanbilge: Agree!

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
```diff
@@ -0,0 +1,41 @@
...
+  def add(data: File): Unit = js.native
+
+  /** Removes the indexth entry in the drag data store. */
+  def remove(index: Int): Unit = js.native
+
+  /** Removes all the entries in the drag data store. */
+  def clear(): Unit = js.native
+
+  @js.annotation.JSBracketAccess
+  /** Returns the DataTransferItem object representing the indexth entry in the drag data store. */
```
armanbilge, line 39:
```suggestion
  /** Returns the DataTransferItem object representing the indexth entry in the drag data store. */
  @js.annotation.JSBracketAccess
```

dom/src/main/scala/org/scalajs/dom/DataTransferItemList.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Adds a new entry for the given data to the drag data store. If the data is plain text then a type string has to be
+    * provided also.
+    */
+  def add(data: String, `type`: String): Unit = js.native
+
+  /** Adds a new entry for the given data to the drag data store. If the data is plain text then a type string has to be
+    * provided also.
+    */
+  def add(data: File): Unit = js.native
```
armanbilge, line 30: MDN says this has a defined return value :)
> A [DataTransferItem](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem) containing the specified data. If the drag item couldn't be created (for example, if the associated [DataTransfer](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer) object has no data store), null is returned.

https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItemList/add#return_value

dom/src/main/scala/org/scalajs/dom/DataTransferItem.scala
```diff
@@ -0,0 +1,28 @@
...
+class DataTransferItem private[this] extends js.Object {
+
+  /** Returns the drag data item kind, one of: "string", "file". */
+  def kind: DragDataItemKind = js.native
+
+  /** Returns the drag data item type string. */
+  def `type`: String = js.native
+
+  /** Invokes the callback with the string data as the argument, if the drag data item kind is text. */
+  def getAsString(callback: Function1[String, Unit]): Unit = js.native
```
armanbilge, line 24: Unqualified `Function1` is the scala type.

```suggestion
  def getAsString(callback: js.Function1[String, Unit]): Unit = js.native
```

api-reports/2_12.txt
```diff
@@ -1297,13 +1297,24 @@ DOMTokenList[JC] def remove(token: String): Unit
 DOMTokenList[JC] def toggle(token: String): Boolean
 DOMTokenList[JC] def toggle(token: String, force: Boolean): Boolean
 DataTransfer[JT] def clearData(format: String?): Unit
-DataTransfer[JT] var dropEffect: String
-DataTransfer[JT] var effectAllowed: String
+DataTransfer[JT] var dropEffect: DropEffectValue
+DataTransfer[JT] var effectAllowed: EffectAllowedValue
```
armanbilge, line 1301: Ok, after many sleepless nights I propose:
- `DataTransferDropEffectKind`
- `DataTransferEffectAllowedKind`

Rationale:
- Prefix with `DataTransfer` to "namespace" them. Since we are just making these names up, there's no guarantee that there won't be some other unrelated `DropEffect` or `EffectAllowed` API added in the future, that would clash with the name.
- Postfix with `Kind`, because the spec describes it as "the kind of operation". Furthermore, looking through other enums we've facaded, we have various "kinds" but not any "values".

WDYT?

zetashift: The prefix to namespace sounds extremely meh, but I can't argue with that reasoning :P.
And yes postfixing with `Kind` sounds great!

armanbilge: > The prefix to namespace sounds extremely me

Heh yeah, welcome to browsers where everything is in the global namespace 😛  anyway we have various precedents for doing that as well, some of which I think has been copied from spec/TS/something 🤔

armanbilge (approved):
Nice work! This was a big one.

Issue #772, opened 2023-04-12
Remove attributes class members from Node


andriygm:
Currently, the Node class contains members not present in JavaScript's Node class. Namely,

`def attributes: NamedNodeMap = js.native`
`def hasAttributes(): Boolean = js.native`

These should be removed.

PR #775, opened 2023-05-06
Remove attributes members from Node class

zetashift:
Resolves https://github.com/scala-js/scala-js-dom/issues/772

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Node.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Node.scala
@@ -82,15 +82,6 @@ abstract class Node extends EventTarget {
   /** Returns the Document that this node belongs to. If no document is associated with it, returns null. */
   def ownerDocument: Document = js.native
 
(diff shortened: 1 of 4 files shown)


dom/src/main/scala/org/scalajs/dom/Node.scala
```diff
@@ -138,11 +129,6 @@ abstract class Node extends EventTarget {
     */
   def contains(otherNode: Node): Boolean = js.native
 
-  /** hasAttributes returns a boolean value of true or false, indicating if the current element has any attributes or
-    * not.
-    */
-  def hasAttributes(): Boolean = js.native
```
armanbilge, line 144: Instead of removing these completely, I think they just need to be moved to the `Element` class.

- https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes
- https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttributes

zetashift: Aaahh good catch!

armanbilge (approved):
Thank you!! :)

PR #787, opened 2023-06-30
add `DateTimeFormatOptions#fractionalSecondDigits`

mraulim:
Hi there!

* run `sbt prePR`

Thanks for contributing!


dom/src/main/scala/org/scalajs/dom/intl/DateTimeFormatOptions.scala
```diff
@@ -85,6 +90,7 @@ object DateTimeFormatOptions {
     val minute0 = minute
     val second0 = second
     val timeZoneName0 = timeZoneName
+    val fractionalSecondDigits0 = fractionalSecondDigits
```
zetashift, line 93: If this is deprecated, then it doesn't need to be included, right?

zetashift (approved):
LGTM! Thank you very much for this :D

armanbilge (approved):

PR #788, opened 2023-07-09
Add `options` parameter to `Document#createElement`

alfonso-landin:
Also to `Document#createElementNS`

https://dom.spec.whatwg.org/#ref-for-dom-document-createelement
https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Document.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Document.scala
@@ -8,7 +8,6 @@ package org.scalajs.dom
 
 import scala.scalajs.js
 import scala.scalajs.js.annotation._
-import scala.scalajs.js.|
 
 /** Each web page loaded in the browser has its own document object. The Document interface serves as an entry point to
   * the web page's content (the DOM tree, including elements such as &lt;body&gt; and &lt;table&gt;) and provides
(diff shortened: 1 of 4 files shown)


dom/src/main/scala/org/scalajs/dom/Document.scala
```diff
@@ -124,10 +125,11 @@ abstract class Document extends Node with NodeSelector with DocumentEvent with P
   /** In an HTML document creates the specified HTML element or HTMLUnknownElement if the element is not known. In a XUL
     * document creates the specified XUL element. In other documents creates an element with a null namespaceURI.
     */
-  def createElement(tagName: String): Element = js.native
+  def createElement(tagName: String, options: String | ElementCreationOptions = js.native): Element = js.native
```
armanbilge, line 128: I've been thinking about it for a bit, and I think instead of union types we should prefer overloads where possible. It's a bit more verbose but they tend to work better with inference. I'm going to add this to the contributing guide.

```suggestion
  def createElement(tagName: String): Element = js.native
  def createElement(tagName: String, options: String): Element = js.native
  def createElement(tagName: String, options: ElementCreationOptions): Element = js.native
```

zetashift: The verbosity isn't bad at all!

armanbilge (approved):
thank you!

PR #794, opened 2023-07-17
Allow `MediaSource` to be passed in to `URL.createObjectURL`

yilinwei:
See [here](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static#parameters).

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/URL.scala
+++ b/dom/src/main/scala/org/scalajs/dom/URL.scala
@@ -2,7 +2,6 @@ package org.scalajs.dom
 
 import scala.scalajs.js
 import scala.scalajs.js.annotation._
-import scala.scalajs.js.|
 
 /** The URL object provides static methods used for creating object URLs. */
 @js.native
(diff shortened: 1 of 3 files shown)


dom/src/main/scala/org/scalajs/dom/URL.scala
```diff
@@ -16,9 +17,9 @@ object URL extends js.Object {
 
   /** The URL.createObjectURL() static method creates a DOMString containing an URL representing the object given in
     * parameter. The URL lifetime is tied to the document in the window on which it was created. The new object URL
-    * represents the specified File object or Blob object.
+    * represents the specified File, Blob or MediaSource object.
     */
-  def createObjectURL(blob: Blob): String = js.native
+  def createObjectURL(obj: Blob | File | MediaSource): String = js.native
```
armanbilge, line 22: `File` extends `Blob` so I don't think we need to add it as another case.

yilinwei: Sure I don't mind either way; what do you want to do with the docstring?

armanbilge: You can either copy it, or just leave it on the first overload. I think that's what we've done in other places.

dom/src/main/scala/org/scalajs/dom/URL.scala
```diff
@@ -16,9 +16,12 @@ object URL extends js.Object {
 
   /** The URL.createObjectURL() static method creates a DOMString containing an URL representing the object given in
     * parameter. The URL lifetime is tied to the document in the window on which it was created. The new object URL
-    * represents the specified File object or Blob object.
+    * represents the specified File, Blob or MediaSource object.
     */
-  def createObjectURL(blob: Blob): String = js.native
+  def createObjectURL(obj: Blob): String = js.native
```
armanbilge, line 21: Can we restore this to `blob: Blob`? thanks!

PR #797, opened 2023-07-22
WIP: Audio API changes

yilinwei:
This is in progress at the moment, since I have an active project which is binding to most of these methods.

I'm opening it, mainly so that there's some visibility on the changes which I'm in the progress of making.

I haven't copy-pasted the docs nor run some of the release reports yet and won't until I finish the changes.


dom/src/main/scala/org/scalajs/dom/OscillatorNode.scala
```diff
@@ -19,8 +20,10 @@ import scala.scalajs.js
   *   - Channel count: 2 (not used in the default count mode)
   *   - Channel interpretation: speakers
   */
+@JSGlobal
 @js.native
-trait OscillatorNode extends AudioNode {
+class OscillatorNode(context: BaseAudioContext, options: OscillatorNodeOptions = js.native)
```
armanbilge, line 25: Unfortunately we cannot change a trait to a class backwards-compatibly :( we'll just have to live with this mistake. We can add `apply` methods for construction and open an issue similar to https://github.com/scala-js/scala-js-dom/issues/745 so we remember for our next found of breaking changes.

yilinwei: There's also a few other changes which I don't know are going to be backwards compatible.

1. Changing some signatures from a `String` to a `sealed trait ... extends js.Any` such as in `AudioNode#channelInterpretation`
2. Changing some signatures which were wrongly typed as an `Int` but returns a `String`, `AudioNode#channelCountMode`
3. Insertion of a type into the hierarchy `BaseAudioContext` and `AudioScheduledNode` and moving some methods as a result
4. Changing a method which had a default parameter to separate methods to allow proper inheritance `AudioScheduledNode#start`.

We could split it into a change where we add the constructors and an object containing 1., but since I'm (unfortunately) using the method which needs a signature change, I'm happy maintaining a fork for a while.

Am I right in thinking this would only require a minor bump though? I note that the issue that you've linked has a tag of a major version bump?

armanbilge: At a glance (1) through (4) sound fine: we can make those changes in a minor bump. But specifically changing a `trait` to a `class` can only be done in a major bump because it is breaking.

> I'm happy maintaining a fork for a while.

At the moment there are no plans for a breaking release 😅 so "for a while" would be "for the foreseeable future".

yilinwei: Hmm I see; OK. I'll switch over the constructors then once I'm closer to being done.

dom/src/main/scala/org/scalajs/dom/OscillatorNode.scala
```diff
@@ -33,21 +34,17 @@ trait OscillatorNode extends AudioNode {
...
     * This replaces the now-obsolete OscillatorNode.setWaveTable.
     */
   def setPeriodicWave(wave: PeriodicWave): Unit = js.native
+}
 
-  /** Used to set the event handler for the ended event, which fires when the tone has stopped playing. */
-  var onended: js.Function1[Event, _] = js.native
+object OscillatorNode {
+  def apply(context: BaseAudioContext, options: js.UndefOr[OscillatorNodeOptions] = js.undefined): GainNode =
+    js.Dynamic.newInstance(js.Dynamic.global.GainNode)(context, options).asInstanceOf[GainNode]
```
yilinwei, line 49: This is frustrating, it doesn't compile in `3.x`. Not sure whether it's a regression or expected behaviour.


```
|Found:    (options : scala.scalajs.js.UndefOr[org.scalajs².dom.GainNodeOptions])
[error]    |Required: scalajs.js.Any
```

armanbilge:
@yilinwei is this PR ready-for-review? otherwise do you mind if we mark as draft so you can un-draft it when it's ready :)

yilinwei:
It's not ready, I'm still slowly playing around with the APIs to make sure
they're correct.

Please mark it as a draft; I didn't realise that was a feature at all!

dom/src/main/scala/org/scalajs/dom/AudioWorkletGlobalScope.scala
```diff
@@ -0,0 +1,49 @@
...
+  * classes.
+  *
+  * Each BaseAudioContext has a single AudioWorklet available under the audioWorklet property, which runs its code in a
+  * single AudioWorkletGlobalScope.
+  *
+  * As the global execution context is shared across the current BaseAudioContext, it's possible to define any other
+  * variables and perform any actions allowed in worklets — apart from defining AudioWorkletProcessor derived classes.
+  */
+@js.native
+trait AudioWorkletGlobalScope extends WorkletGlobalScope {
```
yilinwei, line 22: Need to double-check whether this exists as a global on the processor environment.

zetashift:
Thank you for this! I'll find some time to review this ASAP, and we're hoping to cut a release (with this PR) sometime in January.

dom/src/main/scala/org/scalajs/dom/OscillatorNode.scala
```diff
@@ -33,21 +33,20 @@ trait OscillatorNode extends AudioNode {
...
 
   /** Represents the shape of the oscillator wave generated. Different waves will produce different tones. */
-  var `type`: String = js.native // Not sure if this is correct ...
-
-  /** This method specifies the exact time to start playing the tone. */
-  def start(when: Double = js.native): Unit = js.native
-
-  /** This method specifies the exact time to stop playing the tone. */
-  def stop(when: Double = js.native): Unit = js.native
+  var `type`: OscillatorNodeType = js.native // Not sure if this is correct ...
```
zetashift, line 36: This looks good to me!

dom/src/main/scala/org/scalajs/dom/AudioBufferSourceNode.scala
```diff
@@ -63,16 +63,20 @@ trait AudioBufferSourceNode extends AudioNode {
...
-    *   end immediately. If this method is called twice or more, an exception is raised.
-    */
-  def stop(when: Double = js.native): Unit = js.native
+  def start(when: Double, offset: Double): Unit = js.native
+
+}
+
+object AudioBufferSourceNode {
+
+  import js.`|`.undefOr2jsAny
```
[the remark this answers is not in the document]
zainab-ali: From what I understand, this is an implicit conversion between `UndefOr[...]` and `js.Any`, required for `options`. Is there an alternative way of writing it?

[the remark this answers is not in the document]
armanbilge: Yes, that's right. It was an annoying bug that it needs to be explicitly imported on Scala 3. It's fixed in the latest Scala.js, but we try to keep this project on a conservative version.

- https://github.com/scala-js/scala-js/pull/4904

dom/src/main/scala/org/scalajs/dom/AudioContext.scala
```diff
@@ -131,47 +48,18 @@ class AudioContext extends EventTarget {
     */
   def createMediaStreamDestination(): MediaStreamAudioDestinationNode = js.native
```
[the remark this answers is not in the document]
zainab-ali: Neither I or @yilinwei will have time to test it properly for a while. It might be best to get this PR in and add it later.

dom/src/main/scala/org/scalajs/dom/AudioBufferSourceNodeOptions.scala
```diff
@@ -0,0 +1,65 @@
...
+  /** Represents an enumerated value describing the way channels must be matched between the node's inputs and outputs.
+    * (See AudioNode.channelCountMode for more information including default values.)
+    */
+  var channelCountMode: js.UndefOr[AudioNodeChannelCountMode] = js.undefined
+
+  /** Represents an enumerated value describing the meaning of the channels. This interpretation will define how audio
+    * up-mixing and down-mixing will happen. The possible values are "speakers" or "discrete". (See
+    * AudioNode.channelCountMode for more information including default values.)
+    */
+  var channelInterpretation: js.UndefOr[AudioNodeChannelInterpretation] = js.undefined
```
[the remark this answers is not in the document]
zainab-ali: Sorry - that was a copy / paste mistake. Thanks for picking up on it!

zainab-ali:
Thanks! I've addressed all the feedback - let me know if we've missed anything.

dom/src/main/scala/org/scalajs/dom/AudioContext.scala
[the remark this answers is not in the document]
zetashift: Yes, understandable!

zetashift (approved):

zetashift:
Thank you both a lot for this! I'll wait till @armanbilge reviewed this as well before merging, because reviewing is quite a manual task and I might have missed something.

zainab-ali:
@armanbilge would you be able to take a look when you have the chance?

Issue #798, opened 2023-07-23
It looks we are missing some methods from the FormData class


G-yhlee:
In js world ,

FormData Object have `append` and `get` method,
```js
// example ) https://developer.mozilla.org/en-US/docs/Web/API/FormData/get
const formData = new FormData(event.target);
formData.append("username", "Chris");
formData.append("username", "Bob");
formData.get("username"); // Returns "Chris"
```

but scalajs FormData class has only `append` method, 
we need to add `get` method like this...

```scala
// current  
formData.asInstanceOf[js.Dynamic].get("username").asInstanceOf[String]

// todo :  add get method
formData.get("username")
```
ref.
https://github.com/scala-js/scala-js-dom/blob/d157768ac7def180e0ba00fd5e9bdb20920e159a/dom/src/main/scala/org/scalajs/dom/FormData.scala#L18

faveoled:
Here's the variant generated by AI (untested):
```scala
@js.native
@JSGlobal
class FormData(form: HTMLFormElement = js.native) extends js.Object {

  /** Appends a key/value pair to the FormData object. */
  def append(name: String, value: String | Blob, blobName: String = js.native): Unit = js.native

  /** Deletes a key/value pair from the FormData object. */
  def delete(name: String): Unit = js.native

  /** Returns an iterator that iterates through all key/value pairs contained in the FormData. */
  def entries(): Iterator[js.Tuple2[String, String | Blob]] = js.native

  /** Returns the first value associated with a given key from within a FormData object. */
  def get(name: String): String | Blob = js.native

  /** Returns an array of all the values associated with a given key from within a FormData. */
  def getAll(name: String): js.Array[String | Blob] = js.native

  /** Returns whether a FormData object contains a certain key. */
  def has(name: String): Boolean = js.native

  /** Returns an iterator iterates through all keys of the key/value pairs contained in the FormData. */
  def keys(): Iterator[String] = js.native

  /** Sets a new value for an existing key inside a FormData object, or adds the key/value if it does not already exist. */
  def set(name: String, value: String | Blob, blobName: String = js.native): Unit = js.native

  /** Returns an iterator that iterates through all values contained in the FormData. */
  def values(): Iterator[String | Blob] = js.native
}

```

G-yhlee:
This is updated and merged at https://github.com/scala-js/scala-js-dom/pull/800

Thanks all !

PR #800, opened 2023-07-26
Update dom/form data

G-yhlee:
related issue https://github.com/scala-js/scala-js-dom/issues/798

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/FormData.scala
+++ b/dom/src/main/scala/org/scalajs/dom/FormData.scala
@@ -8,6 +8,7 @@ package org.scalajs.dom
 
 import scala.scalajs.js
 import scala.scalajs.js.annotation._
+import scala.scalajs.js.|
 
 /** XMLHttpRequest Level 2 adds support for the new FormData interface. FormData objects provide a way to easily
   * construct a set of key/value pairs representing form fields and their values, which can then be easily sent using
@@ -15,10 +16,40 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-class FormData(form: HTMLFormElement = js.native) extends js.Object {
+class FormData(form: HTMLFormElement = js.native) extends js.Iterable[js.Tuple2[String, String]] {
 
   /** Appends a key/value pair to the FormData object. */
   def append(name: js.Any, value: js.Any, blobName: String = js.native): Unit = js.native
+
+  /** Deletes a key/value pair from the FormData object. */
+  def delete(name: String): Unit = js.native
+
+  /** Returns the first value associated with a given key from within a FormData object. */
+  def get(name: String): String | Blob = js.native
+
+  /** Returns whether a FormData object contains a certain key. */
+  def has(name: String): Boolean = js.native
+
+  /** Sets a new value for an existing key inside a FormData object, or adds the key/value if it does not already exist.
+    */
+  def set(
+      name: String, value: String | Blob, blobName: String = js.native
+  ): Unit = js.native
+
+  @JSName(js.Symbol.iterator)
+  override def jsIterator(): js.Iterator[js.Tuple2[String, String]] = js.native
+
+  /** Returns an iterator that iterates through all key/value pairs contained in the FormData. */
+  def entries(): js.Iterator[js.Tuple2[String, String | Blob]] = js.native
+
+  /** Returns an array of all the values associated with a given key from within a FormData. */
+  def getAll(name: String): js.Array[String | Blob] = js.native
+
+  /** Returns an iterator iterates through all keys of the key/value pairs contained in the FormData. */
+  def keys(): js.Iterator[String] = js.native
+
+  /** Returns an iterator that iterates through all values contained in the FormData. */
+  def values(): js.Iterator[String | Blob] = js.native
 }
 
 @js.native
--- a/dom/src/main/scala/org/scalajs/dom/FormData.scala
+++ b/dom/src/main/scala/org/scalajs/dom/FormData.scala
@@ -16,7 +16,7 @@ import scala.scalajs.js.|
   */
 @js.native
 @JSGlobal
-class FormData(form: HTMLFormElement = js.native) extends js.Iterable[js.Tuple2[String, String]] {
+class FormData(form: HTMLFormElement = js.native) extends js.Iterable[js.Tuple2[String, String | Blob]] {
 
   /** The `append()` method of the `FormData` interface appends a new value onto an existing key inside a `FormData`
     * object, or adds the key if it does not already exist.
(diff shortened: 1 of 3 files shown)


dom/src/main/scala/org/scalajs/dom/FormData.scala
```diff
@@ -8,17 +8,48 @@ package org.scalajs.dom
...
+class FormData(form: HTMLFormElement = js.native) extends js.Iterable[js.Tuple2[String, String]] {
 
   /** Appends a key/value pair to the FormData object. */
   def append(name: js.Any, value: js.Any, blobName: String = js.native): Unit = js.native
+
+  /** Deletes a key/value pair from the FormData object. */
+  def delete(name: String): Unit = js.native
+
+  /** Returns the first value associated with a given key from within a FormData object. */
+  def get(name: String): String | Blob = js.native
```
zetashift, line 28: Thank you for the PR! Would you mind adding overloads instead of using `|`? See related comment https://github.com/scala-js/scala-js-dom/pull/788#discussion_r1269805846 as it helps with inference more

G-yhlee: Thanks for review :) 
I changed the `append` and `get` methods to functions with overrides.

( And the rest of the function seems to be a case that cannot be replaced with an override )

zetashift:
@G-yhlee I messed up my git-fu, but if you can accept this PR for MDN docstrings on your fork https://github.com/G-yhlee/scala-js-dom-clone/pull/1 then I believe we're good to go!

dom/src/main/scala/org/scalajs/dom/FormData.scala
```diff
@@ -8,17 +8,112 @@ package org.scalajs.dom
...
+    */
+  def delete(name: String): Unit = js.native
+
+  /** The `get()` method of the `FormData` interface returns the first value associated with a given key from within a
+    * `FormData` object. If you expect multiple values and want all of them, use the `getAll()` method instead.
+    *
+    * @param name
+    *   A string representing the name of the key you want to retrieve.
+    * @return
+    *   A value whose key matches the specified name. Otherwise, `null`.
```
zetashift, line 58: I can't remember well, but @armanbilge do you know if we need to do anything special if this can return `null`?

zetashift (approved):
Thank you!!

dom/src/main/scala/org/scalajs/dom/FormData.scala
```diff
@@ -8,17 +8,112 @@ package org.scalajs.dom
...
+  /** The `append()` method of the `FormData` interface appends a new value onto an existing key inside a `FormData`
+    * object, or adds the key if it does not already exist.
+    *
+    * @param name
+    *   The name of the field whose data is contained in value.
+    * @param value
+    *   The field's value. This can be a string or `Blob` (including subclasses such as File). If none of these are
+    *   specified the value is converted to a string.
+    */
+  def append(name: js.Any, value: String): Unit = js.native
```
armanbilge, line 30: Why is `name: js.Any`, isn't only `String` allowed here?

dom/src/main/scala/org/scalajs/dom/FormData.scala
```diff
@@ -8,17 +8,112 @@ package org.scalajs.dom
...
+import scala.scalajs.js.|
 
 /** XMLHttpRequest Level 2 adds support for the new FormData interface. FormData objects provide a way to easily
   * construct a set of key/value pairs representing form fields and their values, which can then be easily sent using
   * the XMLHttpRequest send() method.
   */
 @js.native
 @JSGlobal
-class FormData(form: HTMLFormElement = js.native) extends js.Object {
+class FormData(form: HTMLFormElement = js.native) extends js.Iterable[js.Tuple2[String, String]] {
```
armanbilge, line 19:
```suggestion
class FormData(form: HTMLFormElement = js.native) extends js.Iterable[js.Tuple2[String, String | Blob]] {
```

dom/src/main/scala/org/scalajs/dom/FormData.scala
```diff
@@ -8,17 +8,112 @@ package org.scalajs.dom
...
+    * @param name
+    *   The name of the field whose data is contained in value.
+    * @param value
+    *   The field's value. This can be a string or `Blob` (including subclasses such as File). If none of these are
+    *   specified the value is converted to a string.
+    * @param blobName
+    *   The filename reported to the server (a string), when a `Blob` or `File` is passed as the second parameter. The
+    *   default filename for `Blob` objects is "blob". The default filename for `File` objects is the file's filename.
+    */
+  def append(name: String, value: Blob, blobName: String): Unit = js.native
```
armanbilge, line 44: The last argument is optional.

```suggestion
  def append(name: String, value: Blob, blobName: String = js.native): Unit = js.native
```

dom/src/main/scala/org/scalajs/dom/FormData.scala
```diff
@@ -8,17 +8,112 @@ package org.scalajs.dom
...
+  /** The `set()` method of the `FormData` interface sets a new value for an existing key inside a `FormData` object, or
+    * adds the key/value if it does not already exist.
+    *
+    * @param name
+    *   The name of the field whose data is contained in value.
+    * @param value
+    *   The field's value.
+    */
+  def set(
+      name: String, value: Blob, blobName: String
```
armanbilge, line 92:
```suggestion
      name: String, value: Blob, blobName: String = js.native
```

dom/src/main/scala/org/scalajs/dom/FormData.scala
armanbilge, line 19: Also, it seems there are three possible constructors.

```js
new FormData()
new FormData(form)
new FormData(form, submitter)
```

```scala
class FormData extends ... {
  def this(form: HTMLFormElement) = this()
  def this(form: HTMLFormElement, submitter: HTMLElement) = this
}
```

armanbilge (commented):
Sorry, I noticed just a few more things. Thanks for your work on this!

dom/src/main/scala/org/scalajs/dom/FormData.scala
G-yhlee:
```scala
class FormData(form: HTMLFormElement = js.native) extends ... {
  def this(form: HTMLFormElement) = this(form)
  def this(form: HTMLFormElement, submitter: HTMLElement) = this(form)
}
```
I think this is good point ! 
I've looked at other api codes using `this()` and i make this code, 
but I'm not sure if this change is correct.

I'd be grateful if someone could fix this part for sure

G-yhlee: clear

dom/src/main/scala/org/scalajs/dom/FormData.scala
G-yhlee: updated :)

dom/src/main/scala/org/scalajs/dom/FormData.scala
armanbilge: > I've looked at other api codes using `this()` and i make this code,

That won't work, because there are two constructors that take a `form`. How wil it know which one to use?

I believe the correct change is the one I suggested in https://github.com/scala-js/scala-js-dom/pull/800#discussion_r1280018822.

G-yhlee: got it ! 
I will add this part :)

armanbilge (approved):
Thanks for the revisions. Nice work on this!!

Issue #771, opened 2023-04-11
Use XMLDocument type when parsing XML string in DOMParser


andriygm:
Currently, DOMParser's parseFromString method returns a Document type when using MIME type "text/xml". We should instead return an XMLDocument. It doesn't look like the XMLDocument type exists, so that should be created as well.

armanbilge:
Hmm, I just took a proper look at this and I'm not sure if we can make an improvement here.

We should definitely add `XMLDocument`, I'm about to open a PR that does that.

However, it is difficult to refine the return type of `parseFromString` based on its arguments. This is a very common problem: for example `document.createElement("div")` is returning an `HTMLDivElement` but we are not reflecting that in the type. For the user, the solution is simply to do an `asInstanceOf` cast.

In the future if we dropped support for Scala 2 (or split sources across Scala 2/3) we could possibly do something interesting with literal types. That is tracked in this issue:
- https://github.com/scala-js/scala-js-dom/issues/296

andriygm:
Is `parseXMLFromString` too unwieldy?

PR #804, opened 2023-08-09
Add `XMLDocument`

armanbilge:
`XMLDocument` is essentially just a marker class so this was extremely straightforward. Linking to https://github.com/scala-js/scala-js-dom/issues/771.


zetashift (approved):
LGTM!

PR #806, opened 2023-08-12
Add support for Web NFC API

FabioPinheiro:
Web NFC API - https://developer.mozilla.org/en-US/docs/Web/API/Web_NFC_API

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
@@ -1,0 +1,19 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[NDEFMessage]] interface of the Web NFC API represents the content of an NDEF message that has been read from
+  * or could be written to an NFC tag. An instance is acquired by calling the NDEFMessage() constructor or from the
+  * NDEFReadingEvent.message property, which is passed to the reading event.
+  *
+  * @param records
+  *   The records property of NDEFMessage interface represents a list of NDEFRecords present in the NDEF message.
+  */
+@js.native
+@JSGlobal
+class NDEFMessage(recordsAgr: js.Array[NDEFRecordInit]) extends js.Object {
+
+  /** Returns the list of NDEF records contained in the message. */
+  var records: FrozenArray[NDEFRecord] = js.native
+}
--- a/dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
@@ -7,12 +7,15 @@ import scala.scalajs.js.annotation.JSGlobal
   * or could be written to an NFC tag. An instance is acquired by calling the NDEFMessage() constructor or from the
   * NDEFReadingEvent.message property, which is passed to the reading event.
   *
-  * @param records
-  *   The records property of NDEFMessage interface represents a list of NDEFRecords present in the NDEF message.
+  * @see
+  *   https://w3c.github.io/web-nfc/#the-ndefmessage-interface
+  *
+  * @param messageInit
+  *   property of NDEFMessage interface represents a list of NDEFRecords present in the NDEF message.
   */
 @js.native
 @JSGlobal
-class NDEFMessage(recordsAgr: js.Array[NDEFRecordInit]) extends js.Object {
+class NDEFMessage(messageInit: js.Array[NDEFRecordInit]) extends js.Object {
 
   /** Returns the list of NDEF records contained in the message. */
   def records: FrozenArray[NDEFRecord] = js.native
--- a/dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
@@ -15,5 +15,5 @@ import scala.scalajs.js.annotation.JSGlobal
 class NDEFMessage(recordsAgr: js.Array[NDEFRecordInit]) extends js.Object {
 
   /** Returns the list of NDEF records contained in the message. */
-  var records: FrozenArray[NDEFRecord] = js.native
+  def records: FrozenArray[NDEFRecord] = js.native
 }
--- a/dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
@@ -1,0 +1,60 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[NDEFReader]] interface of the Web NFC API (https://developer.mozilla.org/en-US/docs/Web/API/Web_NFC_API) is
+  * used to read from and write data to compatible NFC devices, e.g. NFC tags supporting NDEF, when these devices are
+  * within the reader's magnetic induction field.
+  */
+@JSGlobal("NDEFReader")
+@js.native
+class NDEFReader() extends EventTarget {
+
+  /** Activates a reading device and returns a Promise that either resolves when an NFC tag read operation is scheduled
+    * or rejects if a hardware or permission error is encountered. This method triggers a permission prompt if the "nfc"
+    * permission has not been previously granted.
+    *
+    * @return
+    *   a Promise that resolves immediately after scheduling read operations for the NFC adapter.
+    */
+  def scan(options: NDEFScanOptions = js.native): js.Promise[Unit] = js.native
+
+  /** Attempts to write an NDEF message to a tag and returns a Promise that either resolves when a message has been
+    * written to the tag or rejects if a hardware or permission error is encountered. This method triggers a permission
+    * prompt if the "nfc" permission has not been previously granted.
+    *
+    * @param message
+    *   The message to be written, either a string object or literal, an ArrayBuffer, a TypedArray, a DataView, or an
+    *   array of records. A record has the following members:
+    * @param options
+    *   An object with the following properties:
+    *
+    * @return
+    *   a Promise that either resolves when a message has been written to the tag or rejects if a hardware or permission
+    *   error is encountered.
+    */
+  def write(message: String, options: NDEFWriteOptions): js.Promise[Unit] = js.native
+  def write(message: js.typedarray.ArrayBuffer, options: NDEFWriteOptions): js.Promise[Unit] = js.native
+
+  def write(message: js.typedarray.TypedArray[_, _],
+      options: NDEFWriteOptions = js.native): js.Promise[Unit] = js.native
+  def write(message: js.typedarray.DataView, options: NDEFWriteOptions): js.Promise[Unit] = js.native
+  def write(message: js.Array[NDEFRecord], options: NDEFWriteOptions): js.Promise[Unit] = js.native
+
+  def write(message: String): js.Promise[Unit] = js.native
+  def write(message: js.typedarray.ArrayBuffer): js.Promise[Unit] = js.native
+  // def write(message:js.typedarray.TypedArray[NDEFRecord, ???]): js.Promise[Unit] = js.native
+  def write(message: js.typedarray.DataView): js.Promise[Unit] = js.native
+  def write(message: js.Array[NDEFRecord]): js.Promise[Unit] = js.native
+
+  /** The reading event of the NDEFReader interface is fired whenever a new reading is available from compatible NFC
+    * devices (e.g. NFC tags supporting NDEF) when these devices are within the reader's magnetic induction field.
+    */
+  var onreading: js.Function1[NDEFReadingEvent, Any] = js.native
+
+  /** The readingerror event of the NDEFReader interface is fired whenever an error occurs during reading of NFC tags,
+    * e.g. when tags leave the reader's magnetic induction field.
+    */
+  var onreadingerror: js.Function1[Event, Any] = js.native
+}
--- a/dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
@@ -6,6 +6,9 @@ import scala.scalajs.js.annotation.JSGlobal
 /** The [[NDEFReader]] interface of the Web NFC API (https://developer.mozilla.org/en-US/docs/Web/API/Web_NFC_API) is
   * used to read from and write data to compatible NFC devices, e.g. NFC tags supporting NDEF, when these devices are
   * within the reader's magnetic induction field.
+  *
+  * @see
+  *   https://w3c.github.io/web-nfc/#the-ndefreader-object
   */
 @JSGlobal("NDEFReader")
 @js.native
--- a/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
@@ -1,0 +1,22 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** The NDEFReadingEvent interface of the Web NFC API represents events dispatched on new NFC readings obtained by
+  * NDEFReader.
+  *
+  * @see
+  *   https://developer.mozilla.org/en-US/docs/Web/API/NDEFReadingEvent
+  */
+@js.native
+trait NDEFReadingEvent extends Event {
+
+  /** Returns an NDEFMessage object containing the received message. */
+  var message: NDEFMessage = js.native
+
+  /** Returns the serial number of the device, which is used for anti-collision and identification, or an empty string
+    * if no serial number is available.
+    */
+  var serialNumber: String = js.native
+
+}
--- a/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
@@ -1,15 +1,19 @@
 package org.scalajs.dom
 
 import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
 
 /** The NDEFReadingEvent interface of the Web NFC API represents events dispatched on new NFC readings obtained by
   * NDEFReader.
   *
   * @see
   *   https://developer.mozilla.org/en-US/docs/Web/API/NDEFReadingEvent
+  * @see
+  *   https://w3c.github.io/web-nfc/#the-ndefreader-object
   */
 @js.native
-trait NDEFReadingEvent extends Event {
+@JSGlobal
+class NDEFReadingEvent(typeArg: String, init: NDEFReadingEventInit) extends Event(typeArg, init) {
 
   /** Returns an NDEFMessage object containing the received message. */
   def message: NDEFMessage = js.native
@@ -18,5 +22,4 @@ trait NDEFReadingEvent extends Event {
     * if no serial number is available.
     */
   def serialNumber: String = js.native
-
 }
--- a/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
@@ -12,11 +12,11 @@ import scala.scalajs.js
 trait NDEFReadingEvent extends Event {
 
   /** Returns an NDEFMessage object containing the received message. */
-  var message: NDEFMessage = js.native
+  def message: NDEFMessage = js.native
 
   /** Returns the serial number of the device, which is used for anti-collision and identification, or an empty string
     * if no serial number is available.
     */
-  var serialNumber: String = js.native
+  def serialNumber: String = js.native
 
 }
--- a/dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
@@ -14,25 +14,25 @@ class NDEFRecord(options: NDEFRecordInit) extends js.Object {
     * "empty", "text", "url", "smart-poster", "absolute-url", "mime", or "unknown" or else an external type name, which
     * consists of a domain name and custom type name separated by a colon (":").
     */
-  var recordType: String = js.native
+  def recordType: String = js.native
 
   /** Returns the MIME type of the record. This value will be null if recordType is not equal to "mime". */
-  var mediaType: js.UndefOr[String] = js.native
+  def mediaType: js.UndefOr[String] = js.native
 
   /** Returns the record identifier, which is an absolute or relative URL used to identify the record.
     *
     * Note: The uniqueness of the identifier is enforced only by the generator of the record.
     */
-  var id: js.UndefOr[String] = js.native
+  def id: js.UndefOr[String] = js.native
 
   /** Returns a DataView containing the raw bytes of the record's payload. */
-  var data: js.typedarray.DataView = js.native
+  def data: js.typedarray.DataView = js.native
 
   /** Returns the encoding of a textual payload, or null otherwise. */
-  var encoding: js.UndefOr[String] = js.native
+  def encoding: js.UndefOr[String] = js.native
 
   /** Returns the language of a textual payload, or null if one was not supplied. */
-  var lang: js.UndefOr[String] = js.native
+  def lang: js.UndefOr[String] = js.native
 
   /** Converts [[NDEFRecord.data]] to a sequence of records. This allows parsing the payloads of record types which may
     * contain nested records, such as smart poster and external type records.
--- a/dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
+++ b/dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
@@ -1,0 +1,41 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[NDEFRecord]] interface of the Web NFC API provides data that can be read from, or written to, compatible NFC
+  * devices, e.g. NFC tags supporting NDEF.
+  */
+@js.native
+@JSGlobal
+class NDEFRecord(options: NDEFRecordInit) extends js.Object {
+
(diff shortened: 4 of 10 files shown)


FabioPinheiro:
I'm getting an error when running `sbt prePR`:

```sbt
[error] scalafix.internal.v1.FileException: unexpected error processing file /home/fabio/workspace/scala-js-dom/dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
[error] Caused by: java.util.NoSuchElementException: None.get
[error] 	at scala.None$.get(Option.scala:529)
[error] 	at scala.None$.get(Option.scala:527)
[error] 	at org.scalajs.dom.scalafix.GenerateApiReport.org$scalajs$dom$scalafix$GenerateApiReport$$process(GenerateApiReport.scala:37)
...
```

dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[NDEFReader]] interface of the Web NFC API (https://developer.mozilla.org/en-US/docs/Web/API/Web_NFC_API) is
+  * used to read from and write data to compatible NFC devices, e.g. NFC tags supporting NDEF, when these devices are
+  * within the reader's magnetic induction field.
+  */
+@JSGlobal("NDEFReader")
+@js.native
+class NDEFReader() extends EventTarget {
```
FabioPinheiro, line 12: The Web_NFC_API is only available on mobile for now.

If we call `NDEFReader()` it may throw a `scala.scalajs.js.JavaScriptException` with a `ReferenceError` (.message -> "NDEFReader is not defined")

What is the recommendation for APIs that are not supported on all platforms

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,52 @@
...
+  def toRecords(): js.Array[NDEFRecord] = js.native
+}
+
+object NDEFRecord {
+
+  type Data = String |
+    js.typedarray.ArrayBuffer |
+    // js.typedarray.TypedArray[NDEFRecord, ???] |
+    js.typedarray.DataView |
+    js.Array[NDEFRecord]
```
FabioPinheiro, line 51: Also, it's unclear from the documentation what the return type is.

and some places say it can be any of those types in other place say its a `DataView`

https://developer.mozilla.org/en-US/docs/Web/API/NDEFRecord/NDEFRecord#:~:text=Contains%20the%20data%20to%20be%20transmitted.%20It%20can%20be%20a%20string%20object%20or%20literal%2C%20an%20ArrayBuffer%2C%20a%20TypedArray%2C%20a%20DataView%2C%20or%20an%20array%20of%20nested%20records.

https://developer.mozilla.org/en-US/docs/Web/API/NDEFRecord/data#:~:text=A%20DataView%20that%20contains%20encoded%20payload%20data%20of%20the%20record.

dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
FabioPinheiro: https://discord.com/channels/632150470000902164/635668814956068864/1139909798469570581

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
FabioPinheiro: I'm going to take the Liberty off saying the return type is `DataView`

This `DataView` is what happened on Android's Chrome is returning.
We can change in the future

dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
```diff
@@ -0,0 +1,65 @@
...
+  var onreading: js.Function1[NDEFReadingEvent, Any] = js.native
+
+  /** The readingerror event of the NDEFReader interface is fired whenever an error occurs during reading of NFC tags,
+    * e.g. when tags leave the reader's magnetic induction field.
+    */
+  var onreadingerror: js.Function1[Event, Any] = js.native
+}
+
+@js.native
+trait NDEFScanOptions extends js.Object {
```
FabioPinheiro, line 51: Should we move this to new files?

dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
```diff
@@ -0,0 +1,15 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[NDEFMessage]] interface of the Web NFC API represents the content of an NDEF message that has been read from
+  * or could be written to an NFC tag. An instance is acquired by calling the NDEFMessage() constructor or from the
+  * NDEFReadingEvent.message property, which is passed to the reading event.
+  *
+  * @param records\
```
zetashift, line 10:
```suggestion
  * @param records
```

dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
```diff
@@ -0,0 +1,52 @@
...
+    *   The message to be written, either a string object or literal, an ArrayBuffer, a TypedArray, a DataView, or an
+    *   array of records. A record has the following members:
+    * @param options
+    *   An object with the following properties:
+    *
+    * @return
+    *   a Promise that either resolves when a message has been written to the tag or rejects if a hardware or permission
+    *   error is encountered.
+    */
+  def write(
```
zetashift, line 38: Could you make overloads for this method instead of using `|`? See https://github.com/scala-js/scala-js-dom/blob/main/CONTRIBUTING.md#facades

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
zetashift: The spec mentions it's a `DataView`: 
https://w3c.github.io/web-nfc/#the-ndefrecord-interface
```
readonly attribute USVString recordType;
  readonly attribute [USVString](https://webidl.spec.whatwg.org/#idl-USVString)? mediaType;
  readonly attribute [USVString](https://webidl.spec.whatwg.org/#idl-USVString)? id;
  readonly attribute [DataView](https://webidl.spec.whatwg.org/#idl-DataView)? [data](https://w3c.github.io/web-nfc/#dom-ndefrecord-data);
  ```

File: dom/src/main/scala/org/scalajs/dom/NDEFReader.scala (hunk 5 above)
Comment (FabioPinheiro) [reply dom/src/main/scala/org/scalajs/dom/NDEFReader.scala:38]: yes but the field 'options' is optional

**Review by armanbilge — COMMENTED** [review]
Sorry, I've had some review comments "pending" for a week. I haven't finished looking at all the changes carefully, but here are my comments so far. Thanks for your work and your patience!

File: dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala [hunk 6]
```diff
@@ -0,0 +1,15 @@
...
+  *   The records property of NDEFMessage interface represents a list of NDEFRecords present in the NDEF message.
+  */
+@js.native
+@JSGlobal
+class NDEFMessage(var records: js.Array[NDEFRecord] = js.native) extends js.Object
```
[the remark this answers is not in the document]
FabioPinheiro: Good point.

Yes the MDN says the constructor method `Creates a new NDEFMessage object, initialized with the given NDEF records."

Let's follow the MDN, and we can add the Constructor method later
```scala
class NDEFMessage extends js.Object {
  /** Returns the list of NDEF records contained in the message. */
  var records: js.Array[NDEFRecord] = js.native
}
````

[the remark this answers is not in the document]
Comment (armanbilge) [reply dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala]: It looks like that should be a `FrozenArray`.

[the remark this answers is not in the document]
Comment (FabioPinheiro) [reply dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala]: updated

File: dom/src/main/scala/org/scalajs/dom/NDEFReader.scala [hunk 7]
```diff
@@ -0,0 +1,58 @@
...
+    * @param options
+    *   An object with the following properties:
+    *
+    * @return
+    *   a Promise that either resolves when a message has been written to the tag or rejects if a hardware or permission
+    *   error is encountered.
+    */
+  def write(message: String, options: NDEFWriteOptions): js.Promise[Unit] = js.native
+  def write(message: js.typedarray.ArrayBuffer, options: NDEFWriteOptions): js.Promise[Unit] = js.native
+  // def write(message:js.typedarray.TypedArray[NDEFRecord, ???] , options: NDEFWriteOptions = js.native): js.Promise[Unit] = js.native
```
zetashift, line 39:
```suggestion
```

Sorry for the nits :P!

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  /** Returns the record identifier, which is an absolute or relative URL used to identify the record.
+    *
+    * Note: The uniqueness of the identifier is enforced only by the generator of the record.
+    */
+  var id: js.UndefOr[String] = js.native
+
+  /** Returns a DataView containing the raw bytes of the record's payload. */
+  var data: js.typedarray.DataView = js.native
+  // var data: DataView = js.native
```
zetashift, line 30:
```suggestion
```

I believe this one can be deleted as well?

dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
FabioPinheiro: But `TypedArray` is a possible return type
I just don't know what todo with the yype parameters in `TypedArray[T, Repr]`

dom/src/main/scala/org/scalajs/dom/NDEFScanOptions.scala
```diff
@@ -0,0 +1,10 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+@js.native
+trait NDEFScanOptions extends js.Object {
```
armanbilge, line 6: Sorry, this encoding is not quite right. It should not have a `@js.native` annotation and the implementations should be `js.undefined`.

Example:
https://github.com/scala-js/scala-js-dom/blob/55205332bb87d30bc2701f3d68876534dc859fe5/dom/src/main/scala/org/scalajs/dom/ClipboardItemOptions.scala#L7-L13

dom/src/main/scala/org/scalajs/dom/NDEFWriteOptions.scala
```diff
@@ -0,0 +1,13 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+@js.native
+trait NDEFWriteOptions extends js.Object {
```
armanbilge, line 6: Same here.

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+/** The [[NDEFRecord]] interface of the Web NFC API provides data that can be read from, or written to, compatible NFC
+  * devices, e.g. NFC tags supporting NDEF.
+  */
+@js.native
+@JSGlobal
+class NDEFRecord extends js.Object {
```
armanbilge, line 11: Constructor missing arguments.
https://developer.mozilla.org/en-US/docs/Web/API/NDEFRecord/NDEFRecord

dom/src/main/scala/org/scalajs/dom/NDEFReader.scala
armanbilge:
```suggestion
  def write(message:js.typedarray.TypedArray[_, _] , options: NDEFWriteOptions = js.native): js.Promise[Unit] = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFWriteOptions.scala
FabioPinheiro: If the field `overwrite` is not optional should we use js.native or nothing?
- `def overwrite: Boolean = js.native`
- `def overwrite: Boolean`

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
FabioPinheiro: Thanks I create a new trait for the argument NDEFRecordOptions. Can you have a look?

FabioPinheiro:
@armanbilge Thanks for the review
I only saw it today when I was updating the version of the lib.
If you could review it again would be much appreciated I have a couple of questions

dom/src/main/scala/org/scalajs/dom/NDEFScanOptions.scala
armanbilge: Gah, that was a bad example! It should be using a `var` there, not a `def`. Sorry. Here's a better one.
https://github.com/scala-js/scala-js-dom/blob/55205332bb87d30bc2701f3d68876534dc859fe5/dom/src/main/scala/org/scalajs/dom/EventSourceInit.scala#L11-L16

dom/src/main/scala/org/scalajs/dom/NDEFWriteOptions.scala
armanbilge: `def overwrite: Boolean` with no initial value. Are you sure it's not optional? It seems to have a default value of `true`, so what happens if you pass an object without it set?

https://w3c.github.io/web-nfc/#dom-ndefwriteoptions

dom/src/main/scala/org/scalajs/dom/NDEFWriteOptions.scala
```diff
@@ -0,0 +1,14 @@
...
+import scala.scalajs.js
+
+/** @see https://developer.mozilla.org/en-US/docs/Web/API/NDEFReader/write */
+trait NDEFWriteOptions extends js.Object {
+
+  /** A boolean value specifying whether or not existing records should be overwritten, if such exists. */
+  def `overwrite`: Boolean
+
+  /** An AbortSignal that allows the current write operation to be canceled. */
+  def `signal`: js.UndefOr[AbortSignal] = js.undefined
```
armanbilge, line 12:
```suggestion
  var `overwrite`: Boolean

  /** An AbortSignal that allows the current write operation to be canceled. */
  var `signal`: js.UndefOr[AbortSignal] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/NDEFScanOptions.scala
```diff
@@ -0,0 +1,9 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+trait NDEFScanOptions extends js.Object {
+
+  /** An AbortSignal that allows the current write operation to be canceled. */
+  def `signal`: js.UndefOr[AbortSignal] = js.undefined
```
armanbilge, line 8:
```suggestion
  var `signal`: js.UndefOr[AbortSignal] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/NDEFRecordInit.scala
```diff
@@ -0,0 +1,51 @@
...
+  *   https://developer.mozilla.org/en-US/docs/Web/API/NDEFMessage/NDEFMessage
+  * @see
+  *   https://developer.mozilla.org/en-US/docs/Web/API/NDEFRecord/NDEFRecord
+  */
+trait NDEFRecordInit extends js.Object {
+
+  /** Contains the data to be transmitted. It can be a string object or literal, an ArrayBuffer, a TypedArray, a
+    * DataView, or an array of nested records.
+    */
+  def data: js.UndefOr[
```
armanbilge, line 16:
```suggestion
  var data: js.UndefOr[
```

dom/src/main/scala/org/scalajs/dom/NDEFRecordInit.scala
```diff
@@ -0,0 +1,51 @@
...
+      String |
+        js.typedarray.DataView |
+        js.typedarray.ArrayBuffer |
+        js.typedarray.TypedArray[_, _] |
+        js.typedarray.DataView |
+        js.Array[NDEFRecord]
+  ] = js.undefined
+
+  /** A string specifying the record's encoding. */
+  def encoding: js.UndefOr[String] = js.undefined
```
armanbilge, line 26:
```suggestion
  var encoding: js.UndefOr[String] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/NDEFRecordInit.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  /** A developer-defined identifier for the record. */
+  def id: js.UndefOr[String] = js.undefined
```
armanbilge, line 29:
```suggestion
  var id: js.UndefOr[String] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/NDEFRecordInit.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  /** A valid language tag according to [RFC 5646: Tags for Identifying Languages (also known as BCP
+    * 47)](https://datatracker.ietf.org/doc/html/rfc5646).
+    */
+  def lang: js.UndefOr[String] = js.undefined
```
armanbilge, line 34:
```suggestion
  var lang: js.UndefOr[String] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/NDEFRecordInit.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  /** A valid MIME type. */
+  def mediaType: js.UndefOr[String] = js.undefined
```
armanbilge, line 37:
```suggestion
  var mediaType: js.UndefOr[String] = js.undefined
```

dom/src/main/scala/org/scalajs/dom/NDEFRecordInit.scala
```diff
@@ -0,0 +1,51 @@
...
+    * It must be one of the following values:
+    *   - "absolute-url" - An absolute URL to the data.
+    *   - "empty" - An empty NDEFRecord.
+    *   - "mime" - A valid MIME type.
+    *   - "smart-poster" - A smart poster as defined by the NDEF-SMARTPOSTER specification.
+    *   - "text" - Text as defined by the NDEF-TEXT specification.
+    *   - "unknown" - The record type is not known.
+    *   - "URL" - A URL as defined by the NDEF-URI specification.
+    */
+  def recordType: js.UndefOr[String] = js.undefined
```
armanbilge, line 50:
```suggestion
  var recordType: js.UndefOr[String] = js.undefined
```

armanbilge (commented):
I'm sorry, the example I linked was wrong 😕 all these `def`s should be `var`s.

dom/src/main/scala/org/scalajs/dom/NDEFWriteOptions.scala
FabioPinheiro: I expect that it has a default. I think I have tested on Chrome but I can do it again. Either way what is the specs that we should follow?
In https://developer.mozilla.org/en-US/docs/Web/API/NDEFReader/write they don't say anything about being optional as opposite to the other fields. But in the w3c specs you send seems to have the default true.
I think I'll change it to be optional (it makes more sense and we can always fix in the future).

dom/src/main/scala/org/scalajs/dom/NDEFMessage.scala
```diff
@@ -0,0 +1,19 @@
...
+  *
+  * @param records
+  *   The records property of NDEFMessage interface represents a list of NDEFRecords present in the NDEF message.
+  */
+@js.native
+@JSGlobal
+class NDEFMessage(recordsAgr: js.Array[NDEFRecordInit]) extends js.Object {
+
+  /** Returns the list of NDEF records contained in the message. */
+  var records: FrozenArray[NDEFRecord] = js.native
```
armanbilge, line 18:
```suggestion
  def records: FrozenArray[NDEFRecord] = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
```diff
@@ -0,0 +1,22 @@
...
+@js.native
+trait NDEFReadingEvent extends Event {
+
+  /** Returns an NDEFMessage object containing the received message. */
+  var message: NDEFMessage = js.native
+
+  /** Returns the serial number of the device, which is used for anti-collision and identification, or an empty string
+    * if no serial number is available.
+    */
+  var serialNumber: String = js.native
```
armanbilge, line 20:
```suggestion
  def serialNumber: String = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFReadingEvent.scala
```diff
@@ -0,0 +1,22 @@
...
+import scala.scalajs.js
+
+/** The NDEFReadingEvent interface of the Web NFC API represents events dispatched on new NFC readings obtained by
+  * NDEFReader.
+  *
+  * @see
+  *   https://developer.mozilla.org/en-US/docs/Web/API/NDEFReadingEvent
+  */
+@js.native
+trait NDEFReadingEvent extends Event {
```
armanbilge, line 12: This should be a `class` and it needs a constructor.

https://developer.mozilla.org/en-US/docs/Web/API/NDEFReadingEvent/NDEFReadingEvent

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+  */
+@js.native
+@JSGlobal
+class NDEFRecord(options: NDEFRecordInit) extends js.Object {
+
+  /** Returns the record type of the record. Records must have either a standardized well-known type name such as
+    * "empty", "text", "url", "smart-poster", "absolute-url", "mime", or "unknown" or else an external type name, which
+    * consists of a domain name and custom type name separated by a colon (":").
+    */
+  var recordType: String = js.native
```
armanbilge, line 17:
```suggestion
  def recordType: String = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Returns the MIME type of the record. This value will be null if recordType is not equal to "mime". */
+  var mediaType: js.UndefOr[String] = js.native
```
armanbilge, line 20:
```suggestion
  def mediaType: js.UndefOr[String] = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Returns the record identifier, which is an absolute or relative URL used to identify the record.
+    *
+    * Note: The uniqueness of the identifier is enforced only by the generator of the record.
+    */
+  var id: js.UndefOr[String] = js.native
```
armanbilge, line 26:
```suggestion
  def id: js.UndefOr[String] = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Returns a DataView containing the raw bytes of the record's payload. */
+  var data: js.typedarray.DataView = js.native
```
armanbilge, line 29:
```suggestion
  def data: js.typedarray.DataView = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Returns the encoding of a textual payload, or null otherwise. */
+  var encoding: js.UndefOr[String] = js.native
```
armanbilge, line 32:
```suggestion
  def encoding: js.UndefOr[String] = js.native
```

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Returns the language of a textual payload, or null if one was not supplied. */
+  var lang: js.UndefOr[String] = js.native
```
armanbilge, line 35:
```suggestion
  def lang: js.UndefOr[String] = js.native
```

armanbilge (commented):
Hmm, did you change those `def`s to `var`s? Some of them should be `def`s. If it is "read-only", then it is a `def`.

dom/src/main/scala/org/scalajs/dom/NDEFRecord.scala
FabioPinheiro: I'm assuming I also need to use `js.undefined` (every time I have a `js.UndefOr`)

armanbilge: Actually, no :)

We use `js.undefined` for the `trait`s that are _not_ `@js.native` because they are for constructing `js.Object`s without specialized constructors.

Everything else is facades for JS classes, so we use `js.native` for them.

FabioPinheiro: yeah I just realized that after I did the question
(the compiler complain about it =P )

FabioPinheiro:
https://w3c.github.io/web-nfc/ format seems more clear to follow.
I did another review of all fields to seem if they are optional or not.
Hopefully everything is correct now.

zetashift (approved):
LGTM! Thank you for this awesome PR!

FabioPinheiro:
Would be nice to have this merged before the next release

zetashift:
@FabioPinheiro sorry for the wait, and thank you so much for this PR!

PR #808, opened 2023-08-23
Add Navigator's ProtocolHandler methods

FabioPinheiro:
Add to Navigator the two protocol handler methods:
- registerProtocolHandler - https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler
- unregisterProtocolHandler - https://developer.mozilla.org/en-US/docs/Web/API/Navigator/unregisterProtocolHandler

### DIFF
--- a/api-reports/2_12.txt
+++ b/api-reports/2_12.txt
@@ -16477,13 +16477,20 @@ Navigator[JC] def mediaDevices: MediaDevices
 Navigator[JC] def onLine: Boolean
 Navigator[JC] val permissions: Permissions
 Navigator[JC] def platform: String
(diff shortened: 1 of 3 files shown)


FabioPinheiro:
I added the methods to `NavigatorContentUtils` since that is what TypeScript does. As points by @armanbilge
https://github.com/microsoft/TypeScript/blob/fcaa900012252bf2ed90ab31e1a7e4660c88bf28/src/lib/dom.generated.d.ts#L15929C11-L15936/blob/fcaa900012252bf2ed90ab31e1a7e4660c88bf28/src/lib/dom.generated.d.ts#L15834

armanbilge (approved):
Thanks!

PR #817, opened 2023-09-22
Add `BroadcastChannel`

armanbilge:
https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel


zetashift (approved):
LGTM, didn't even know this functionality existed, it sounds great!

dom/src/main/scala/org/scalajs/dom/webgl/BroadcastChannel.scala
zetashift, line 1: Wait, one question, why is this located in the `webgl` folder?

armanbilge: That is a great question 😂

Issue #807, opened 2023-08-23
Confusing type of ImageData.data


OndrejSpanel:
The ImageData.data is defined and documented as follows:

https://github.com/scala-js/scala-js-dom/blob/7d0508cac24b4f2657c3525797fe633feb4849bc/dom/src/main/scala/org/scalajs/dom/ImageData.scala#L23-L26

I am confused: is the type `Uint8ClampedArray`, or  `js.Array[Int]`? Or are those types interchangeable somehow? If they are, perhaps the documentation could clarify how.

armanbilge:
Hmm. I don't _think_ those are interchangeable. Seems like a bug to me.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8ClampedArray

armanbilge:
Yes, we should use this instead.

https://github.com/scala-js/scala-js/blob/7b3092ae64194af7ac34ad6c12f101a382956fdb/library/src/main/scala/scala/scalajs/js/typedarray/Uint8ClampedArray.scala#L24

OndrejSpanel:
I guess the interface of both JS types is similar enough, they both contain `length` and bracket access, therefore one can perform some basic operations on `TypedArray` thinking it is a normal array.

PR #818, opened 2023-09-27
Fix type of `ImageData#data`

armanbilge:
Fixes https://github.com/scala-js/scala-js-dom/issues/807.

https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data


zetashift (approved):
Sweet! :D

PR #822, opened 2023-11-15
Add `hasPointerCapture/releasePointerCapture/setPointerCapture` to `Element`

updraft0:
Looks like these were missing from the `Element` facade:

* https://developer.mozilla.org/en-US/docs/Web/API/Element/hasPointerCapture
* https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture
* https://developer.mozilla.org/en-US/docs/Web/API/Element/releasePointerCapture

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/Element.scala
+++ b/dom/src/main/scala/org/scalajs/dom/Element.scala
@@ -270,6 +270,22 @@ abstract class Element extends Node with NodeSelector with ParentNode with NonDo
     */
   def requestPointerLock(): Unit = js.native
 
+  /** The hasPointerCapture() method of the Element interface checks whether the element on which it is invoked has
+    * pointer capture for the pointer identified by the given pointer ID.
+    */
+  def hasPointerCapture(pointerId: Double): Boolean = js.native
+
+  /** The releasePointerCapture() method of the Element interface releases (stops) pointer capture that was previously
+    * set for a specific (PointerEvent) pointer.
+    */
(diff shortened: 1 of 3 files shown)


dom/src/main/scala/org/scalajs/dom/Element.scala
```diff
@@ -270,6 +270,22 @@ abstract class Element extends Node with NodeSelector with ParentNode with NonDo
...
 
+  /** The hasPointerCapture() method of the Element interface checks whether the element on which it is invoked has
+    * pointer capture for the pointer identified by the given pointer ID.
+    */
+  def hasPointerCapture(pointerId: Double): Boolean = js.native
+
+  /** The releasePointerCapture() method of the Element interface releases (stops) pointer capture that was previously
+    * set for a specific (PointerEvent) pointer.
+    */
+  def releasePointerCapture(pointerId: Double): Boolean = js.native
```
zetashift, line 281:
```suggestion
  def releasePointerCapture(pointerId: Double): Unit = js.native
```
According to the spec `releasePointerCapture` returns undefined: https://w3c.github.io/pointerevents/#dom-element-releasepointercapture

zetashift (changes requested):

dom/src/main/scala/org/scalajs/dom/Element.scala
updraft0: So it does, fixed

zetashift (approved):

zetashift:
Thank you! :D

PR #828, opened 2024-01-05
add roundRect

buntec:
Hi there!

* run `sbt prePR`

Thanks for contributing!

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/CanvasRenderingContext2D.scala
+++ b/dom/src/main/scala/org/scalajs/dom/CanvasRenderingContext2D.scala
@@ -169,6 +169,9 @@ class CanvasRenderingContext2D extends js.Object {
   /** Creates a path for a rectangle at position (x, y) with a size that is determined by width and height. */
   def rect(x: Double, y: Double, w: Double, h: Double): Unit = js.native
 
+  /** Creates a path for a rounded rectangle with a specified position, width, height, and corner radii. */
+  def roundRect(x: Double, y: Double, w: Double, h: Double, radii: js.Any): Unit = js.native
+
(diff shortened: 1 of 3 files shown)


zetashift:
> Alas, that `radii` param is a mess to encode (probably why this method was missing in the first place).

Oh yup, from https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-roundrect

> If radii is an [unrestricted double](https://webidl.spec.whatwg.org/#idl-unrestricted-double) or [DOMPointInit] (https://drafts.fxtf.org/geometry/#dictdef-dompointinit), then set radii to « radii ».
> 
> If radii is not a list of size one, two, three, or four, then throw a [RangeError](https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).

I wonder if we should create a new enum type that encodes this or just stick with `Any`. What do you think @armanbilge?

dom/src/main/scala/org/scalajs/dom/CanvasRenderingContext2D.scala
```diff
@@ -169,6 +169,9 @@ class CanvasRenderingContext2D extends js.Object {
   /** Creates a path for a rectangle at position (x, y) with a size that is determined by width and height. */
   def rect(x: Double, y: Double, w: Double, h: Double): Unit = js.native
 
+  /** Creates a path for a rounded rectangle with a specified position, width, height, and corner radii. */
+  def roundRect(x: Double, y: Double, w: Double, h: Double, radii: js.Any): Unit = js.native
```
armanbilge, line 173: Thanks for the PR! Instead of `js.Any` can we add several overloads of this method for the various types of `radii`?

armanbilge (changes requested):

PR #829, opened 2024-01-12
Added missing constructors to ImageData facade.

dragonfly-ai:
The ImageData spec references several constructors that scala-js-dom has omitted:

https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData

This PR means to include these features.


dragonfly-ai:
Should we introduce an ImageSettings class?

The JS docs just supply them with anonymous argiments:

```javascript

let imageData = new ImageData(200, 100, { colorSpace: "display-p3" });

```

armanbilge:
> Should we introduce an ImageSettings class?

Yes. Here's an example of how it should be defined.

https://github.com/scala-js/scala-js-dom/blob/efe1b87772b114c95a3d30fcdddc603f0ebb0b23/dom/src/main/scala/org/scalajs/dom/DOMRectInit.scala

dragonfly-ai:
OK, I'll try to add that, too.

dragonfly-ai:
> Yes. Here's an example of how it should be defined.

Should we also have a companion object with the two supported values?

"colorSpace: Specifies the color space of the image data. Can be set to `"srgb"` ... or `"display-p3"` ..."

```scala
object ImageSettings {
  val sRGB:String = "srgb"
  val DisplayP3:String = "display-p3"
}
```

dragonfly-ai:
What about adding a companion object to `ImageSettings` since its only field supports exactly two values: "srgb" and "display-p3"?

zetashift:
> What about adding a companion object to ImageSettings since its only field supports exactly two values: "srgb" and "display-p3"?

Wouldn't these be better off as an `enum` then? E.g. `ImageSettingsColorSpace`?
@dragonfly-ai This PR LGTM, if you want to add the enum or not, let me know, it can always be a separate ticket/PR

zetashift (approved):

dragonfly-ai:
Thanks for approving!
I've never tried passing a Scala enum to a native JS function; do Scala enumerations interoperate well with plain JavaScript?

sjrd:
No, they don't.

dragonfly-ai:
Well, in that case, companion object with constants for `display-p3` and `srgb`, or should we just leave it out?

dragonfly-ai:
Added the `ImageSettings` object as discussed.

dom/src/main/scala/org/scalajs/dom/ImageSettings.scala
```diff
@@ -0,0 +1,12 @@
...
+import scala.scalajs.js
+
+trait ImageSettings extends js.Object {
+  var colorSpace: js.UndefOr[String] = js.undefined
+}
+
+object ImageSettings {
+  lazy val sRGB: ImageSettings = js.Dynamic.literal(colorSpace = "srgb").asInstanceOf[ImageSettings]
+  lazy val displayP3: ImageSettings = js.Dynamic.literal(colorSpace = "display-p3").asInstanceOf[ImageSettings]
+}
```
zetashift, line 12: Enums are done a bit differently to support Scala 2 and Scala 3. You'll need 2 files, one in `scala-2` and one in `scala-3

For example the `CompressionFormat` enum is made like this:
Scala 3: https://github.com/scala-js/scala-js-dom/blob/efe1b87772b114c95a3d30fcdddc603f0ebb0b23/dom/src/main/scala-3/org/scalajs/dom/CompressionFormat.scala#L5-L13
Scala 2: https://github.com/scala-js/scala-js-dom/blob/efe1b87772b114c95a3d30fcdddc603f0ebb0b23/dom/src/main/scala-2/org/scalajs/dom/CompressionFormat.scala#L5-L14

zetashift (changes requested):

dom/src/main/scala/org/scalajs/dom/ImageSettings.scala
```diff
@@ -0,0 +1,12 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+trait ImageSettings extends js.Object {
+  var colorSpace: js.UndefOr[String] = js.undefined
+}
+
+object ImageSettings {
```
armanbilge, line 9: 1. This should be called [`ImageDataSettings`](https://html.spec.whatwg.org/multipage/canvas.html#imagedatasettings)
2. Remove the companion object
3. As @zetashift points out we should add an enum [`PredefinedColorSpace`](https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace)

armanbilge (changes requested):

dom/src/main/scala/org/scalajs/dom/ImageSettings.scala
dragonfly-ai: I've tried to do all of these things, but can't get the enum to work.

Can you see the problem?
https://scastie.scala-lang.org/C3O0MaFoQpW8dhGT7QIJDw

dragonfly-ai: I would like to also add this test, but can't figure out where to put it:

```scala

  @Test
  final def ImageDataTest(): Unit = {
    import org.scalajs.dom.{ImageData, ImageDataSettings, PredefinedColorSpace}
    import PredefinedColorSpace._

    val defaultImageData: ImageData = new ImageData(200, 100, new ImageDataSettings { colorSpace = `display-p3` })
    assertEquals(defaultImageData.width, 200)
    assertEquals(defaultImageData.height, 100)
  }
  ```

armanbilge: > I've tried to do all of these things, but can't get the enum to work with string literals.

Can you push the code and I'll review it? Thanks :)



> I would like to also add this test, but can't figure out where to put it:

How about here?

https://github.com/scala-js/scala-js-dom/blob/009226a920c4bca457ef8c5c8a57c0f1283c193d/tests-shared/src/main/scala/org/scalajs/dom/tests/shared/SharedTests.scala

dragonfly-ai: I tried that first, but `ImageData` isn't defined in node.

`[error] Test org.scalajs.dom.tests.node.jsdom.NodeJsdomTests.ImageDataTest failed: scala.scalajs.js.JavaScriptException: ReferenceError: ImageData is not defined, took 0.000862245 sec`

Maybe it ought to be?  (I sure would like to have it there.)

dragonfly-ai: pushed!  ready for review.

armanbilge: > I tried that first, but `ImageData` isn't defined in node.

I see, that's too bad. In that case move it to the `BrowserTests` file.

dragonfly-ai: Awesome!  Adding that test may have exposed a bug in Firefox:

image

dragonfly-ai: all of the other constructors seem to work, so I'll write tests for all of those and comment this one out.

armanbilge (approved):
nice work, thanks!

PR #831, opened 2024-01-14
add upgrade whenDefined to CustomElementRegistry

cyz1901:
Hi there!

* run `sbt prePR`

Thanks for contributing!

add upgrade whenDefined to CustomElementRegistry

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
+++ b/dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
@@ -18,4 +18,8 @@ abstract class CustomElementRegistry extends js.Object {
 
   /** Defines a new custom element. */
   def define(name: String, constructor: js.Dynamic, options: ElementDefinitionOptions = js.native): Unit
+
+  def upgrade(root: Node): Unit
+
+  def whenDefined(name: String): js.Promise[Any]
 }
--- a/api-reports/2_12.txt
+++ b/api-reports/2_12.txt
@@ -1442,6 +1442,8 @@ CryptoKey[JC] val usages: js.Array[KeyUsage]
 CryptoKeyPair[JT] val privateKey: CryptoKey
 CryptoKeyPair[JT] val publicKey: CryptoKey
 CustomElementRegistry[JC] def define(name: String, constructor: js.Dynamic, options: ElementDefinitionOptions?): Unit
+CustomElementRegistry[JC] def upgrade(root: Node): Unit
+CustomElementRegistry[JC] def whenDefined(name: String): js.Promise[Any]
 CustomEvent[JC] def bubbles: Boolean
 CustomEvent[JC] def cancelBubble: Boolean
 CustomEvent[JC] def cancelable: Boolean
--- a/api-reports/2_13.txt
+++ b/api-reports/2_13.txt
@@ -1442,6 +1442,8 @@ CryptoKey[JC] val usages: js.Array[KeyUsage]
 CryptoKeyPair[JT] val privateKey: CryptoKey
 CryptoKeyPair[JT] val publicKey: CryptoKey
 CustomElementRegistry[JC] def define(name: String, constructor: js.Dynamic, options: ElementDefinitionOptions?): Unit
+CustomElementRegistry[JC] def upgrade(root: Node): Unit
+CustomElementRegistry[JC] def whenDefined(name: String): js.Promise[Any]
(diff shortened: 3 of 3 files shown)


cyz1901:
Hello team. I am not sure should we also add the [get()](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/get) and [getName()](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/getName) functions, which seems to involve the type being a constructor...

armanbilge:
Hmm. I think the type for a constructor should be `js.Dynamic`. For example see `val constructor` in this API:
https://www.scala-js.org/api/scalajs-library/latest/scala/scalajs/js/ConstructorTag.html

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -11,11 +11,20 @@ import scala.scalajs.js.annotation._
 
 /** The CustomElementRegistry interface provides methods for registering custom elements and querying registered
   * elements. To get an instance of it, use the window.customElements property.
+  * https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry
   */
 @js.native
 @JSGlobal
 abstract class CustomElementRegistry extends js.Object {
 
+  def get(name: String): js.Dynamic
```
zetashift, line 20:
```suggestion
  /** Returns the constructor for a previously-defined custom element.  */
  def get(name: String): js.Dynamic
```

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -11,11 +11,20 @@ import scala.scalajs.js.annotation._
...
+
+  def getName(constructor: js.Dynamic): String
```
zetashift, line 22:
```suggestion
   /** Returns the name for a previously-defined custom element.
     * @param constructor
     *   Constructor for the custom element.
     */
   def getName(constructor: js.Dynamic): String
```

zetashift (changes requested):
Could use some doc comments, but otherwise looks good to me! (You might have to format my example of doc comments).

cyz1901:
fixed

zetashift (approved):
Thank you!

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -11,11 +11,29 @@ import scala.scalajs.js.annotation._
 
 /** The CustomElementRegistry interface provides methods for registering custom elements and querying registered
   * elements. To get an instance of it, use the window.customElements property.
+  * https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry
```
armanbilge, line 14:
```suggestion
```

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -11,11 +11,29 @@ import scala.scalajs.js.annotation._
...
   */
 @js.native
 @JSGlobal
 abstract class CustomElementRegistry extends js.Object {
 
+  /** Returns the constructor for a previously-defined custom element. */
+  def get(name: String): js.Dynamic
```
armanbilge, line 21:
```suggestion
  def get(name: String): js.UndefOr[js.Dynamic]
```

armanbilge (changes requested):

dom/src/main/scala/org/scalajs/dom/CustomElementRegistry.scala
```diff
@@ -11,11 +11,29 @@ import scala.scalajs.js.annotation._
...
   /** Defines a new custom element. */
   def define(name: String, constructor: js.Dynamic, options: ElementDefinitionOptions = js.native): Unit
+
+  /** Upgrades a custom element directly, even before it is connected to its shadow root. */
+  def upgrade(root: Node): Unit
+
+  /** Returns an empty Promise that resolves when a custom element becomes defined with the given name. If such a custom
+    * element is already defined, the returned promise is immediately fulfilled.
+    */
+  def whenDefined(name: String): js.Promise[Any]
```
armanbilge, line 38: Please copy the doc from here.

```suggestion
  def whenDefined(name: String): js.Promise[js.Dynamic]
```

PR #832, opened 2024-02-01
Generic error event for EventSource

cornerman:
According to MDN the error of an EventSource is a generic Event instead of an ErrorEvent: https://developer.mozilla.org/en-US/docs/Web/API/EventSource/error_event

Just caught one of these underspecified errors in the wild :)


zetashift (approved):
Thank you!

armanbilge (approved):
Good catch. Looks like this was a regression introduced in 4de6882aec3512f04bbf29debc25ef493d33bf15. I wonder if any more of those changes are wrong 😅

PR #835, opened 2024-02-21
Fix URLSearchParams constructor parameters

### LINKED ISSUE #834: Constructor parameters for URLSearchParams are not quite right
[URLSearchParams](https://github.com/scala-js/scala-js-dom/blob/efe1b87772b114c95a3d30fcdddc603f0ebb0b23/dom/src/main/scala/org/scalajs/dom/URLSearchParams.scala#L13) has:
```
  def this(init: Sequence[String]) = this()
```

where `Sequence[String]` is an alias for `js.Array[String]`.

However this should be a sequence of string pairs not just strings, i.e. `Sequence[js.Tuple2[String, String]]`.

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams#options:
> A literal sequence of name-value string pairs, or any object — such as a [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object — with an [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_generators#iterators) that produces a sequence of string pairs. Note that [File](https://developer.mozilla.org/en-US/docs/Web/API/File) entries will be serialized as [object File] rather than as their filename (as they would in an application/x-www-form-urlencoded form).

steinybot:
Fixes #834

I ran `sbt prePR` but didn't see any changes.

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/URLSearchParams.scala
+++ b/dom/src/main/scala/org/scalajs/dom/URLSearchParams.scala
@@ -10,7 +10,7 @@ class URLSearchParams extends js.Iterable[js.Tuple2[String, String]] {
 
   /** Leading '?' characters are ignored. */
   def this(init: String) = this()
-  def this(init: Sequence[String]) = this()
+  def this(init: Sequence[js.Tuple2[String, String]]) = this()
   def this(init: js.Dictionary[String]) = this()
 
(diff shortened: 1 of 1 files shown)


zetashift (approved):
Great catch, that subtle! Thank you.

PR #838, opened 2024-03-10
Add support for WebGL2

tsnee:
This PR is an attempt to add complete support for WebGL 2 (as defined [at MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext), _not_ [the official WebGL2 spec](https://registry.khronos.org/webgl/specs/latest/2.0/)). In the course of preparing this PR I have uncovered a few inconsistencies between the two and [alerted MDN](https://github.com/mdn/content/issues/32629).

This is my first scala-js related PR and I could use some help. For one thing, there are many ScalaDoc warnings, mostly about references to `scala.scalajs.js` classes. IntelliJ thinks references like `[[js.Array]]` in my ScalaDocs are fine but sbt does not, and I don't know what to do about it. Another issue is numeric data types. I was advised to use `Int` for everything that would fit in 32 bits and `Double` for everything that would not, and I did my best to follow that advice, but I don't know if there is a way to verify that I did it perfectly.

Any and all comments and criticism are welcome.


zetashift:
Thank you very much!!! This is amazing work. The reviewing process might be a bit slow, so please bear with us!

dom/src/main/scala/org/scalajs/dom/WebGLQuery.scala
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (C) 2014 Matt Seddon.  This source is donated in full to the scala-js-dom project.
```
armanbilge, line 2: Sorry, who is Matt Seddon? What is the origin and license of this code?

tsnee: I created this file and and several others by copying WebGLBuffer.scala and changing only the class name and comment. I didn't feel good about copying the code without the copyright statement. Did I err?

mseddon: Oh sorry, that is me, it appears to have crept in from a template years ago- please remove the license, it is unnecessary. All code committed here is licensed to the scala.js project

mseddon: @armanbilge Confirming via ping since there hasn't been much movement.

armanbilge: @mseddon Thanks for the ping, I had missed this (although haven't had much time recently for sjs-dom maintenance). I appreciate the clarification! In retrospect, it should have been obvious that `mseddon` is Matt Seddon 😅

PR #839, opened 2024-03-12
Add composition related handlers and field

nineclue:
Keyboard Event is missing isComposing field.
HTMLElment is missing composition related event handlers.

Both are listed in MDN documents. 
* [KeyboardEvent.isComposing](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing)
* [compositionstart Event](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event)

This is my first pull request. Hope it works out.


dom/src/main/scala/org/scalajs/dom/HTMLElement.scala
armanbilge, line 1: Thanks for the PR. If I understand correctly, these APIs should be added to `Element`, not `HTMLElement`.

https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event

armanbilge (changes requested):

dom/src/main/scala/org/scalajs/dom/HTMLElement.scala
nineclue: You're correct.  Moved handlers to Element.
Thanks for pointing out.

armanbilge (approved):
Excellent, thank you!

PR #840, opened 2024-03-15
Added Canvas 2D filter property

daniel-ciocirlan:
Added the [`filter` property](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/filter) to the CanvasRenderingContext2D JavaScript class.


zetashift (approved):
Thank you! 🚀

Issue #754, opened 2022-12-29
Typo in "WriteableStream" trait name?


raquo:
It seems that the native interface is actually called [WritableStream](https://streams.spec.whatwg.org/#writablestream), however Scala trait name has an extra `e` character in it.

I'm not familiar with this API, the trait is not JSGlobal, and I don't actually see how one could get instances of this trait in scala-js-dom, but nevertheless it should probably match the native name?

armanbilge:
Thanks, this indeed seems like a mistake :( there are actually many issues with those facades, see https://github.com/scala-js/scala-js-dom/issues/629.

I don't think we can fix this binary-compatibly. Maybe we can add a type alias with the correct name.

Issue #629, opened 2021-11-11
Revisit `ReadableStream` et al. `class`/`trait` designations


armanbilge:
I don't think we can change these binary-compatibly, but the current designations are causing all sorts of friction in #628.

PR #841, opened 2024-03-17
[WIP] Support for OPFS

DanielMoss:
Hi!

I was hoping to add support for the [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system). I've run into two areas where I was hoping to get advice - I'll highlight these in PR comments.

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/FileSystemDirectoryHandle.scala
+++ b/dom/src/main/scala/org/scalajs/dom/FileSystemDirectoryHandle.scala
@@ -1,0 +1,67 @@
+/** All documentation for facades is thanks to Mozilla Contributors at https://developer.mozilla.org/en-US/docs/Web/API
+  * and available under the Creative Commons Attribution-ShareAlike v2.5 or later.
+  * http://creativecommons.org/licenses/by-sa/2.5/
+  *
+  * Everything else is under the MIT License http://opensource.org/licenses/MIT
+  */
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** The FileSystemDirectoryHandle interface of the File System API provides a handle to a file system directory. */
+@js.native
+trait FileSystemDirectoryHandle extends FileSystemHandle {
+
+  /** The getDirectoryHandle() method of the FileSystemDirectoryHandle interface returns a FileSystemDirectoryHandle for
+    * a subdirectory with the specified name within the directory handle on which the method is called.
+    *
+    * @param name
+    *   A string representing the FileSystemHandle.name of the file you wish to retrieve.
+    * @param options
+    *   An optional object containing options for the retrieved subdirectory.
+    *
+    * @return
+    *   A Promise which resolves with a FileSystemDirectoryHandle.
+    */
+  def getDirectoryHandle(name: String,
+      options: js.UndefOr[FileSystemGetDirectoryOptions] = js.native): js.Promise[FileSystemDirectoryHandle] = js.native
+
+  /** The getFileHandle() method of the FileSystemDirectoryHandle interface returns a FileSystemFileHandle for a file
+    * with the specified name, within the directory the method is called.
+    *
+    * @param name
+    *   A string representing the FileSystemHandle.name of the subdirectory you wish to retrieve.
+    * @param options
+    *   An object.
+    *
+    * @return
+    *   A Promise which resolves with a FileSystemFileHandle.
+    */
+  def getFileHandle(name: String,
+      options: js.UndefOr[FileSystemGetFileOptions] = js.native): js.Promise[FileSystemFileHandle] = js.native
+
+  /** The removeEntry() method of the FileSystemDirectoryHandle interface attempts to remove an entry if the directory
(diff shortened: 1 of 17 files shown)


dom/src/main/scala/org/scalajs/dom/FileSystemDirectoryHandle.scala
DanielMoss, line 1: This type [also defines](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle#instance_methods) methods that return [async iterators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols), which don't appear to be supported at the moment in scala-js. It makes sense to me to tackle those methods in a separate piece of work. I've asked about potentially adding support over in the [scala-js discord](https://discord.com/channels/632150470000902164/635668814956068864/1218993100530126848).

dom/src/main/scala/org/scalajs/dom/FileSystemWritableFileStream.scala
```diff
@@ -0,0 +1,71 @@
...
+package org.scalajs.dom
+
+import scala.scalajs.js
+
+/** The FileSystemWritableFileStream interface of the File System API is a WritableStream object with additional
+  * convenience methods, which operates on a single file on disk. The interface is accessed through the
+  * FileSystemFileHandle.createWritable() method.
+  */
+@js.native
+trait FileSystemWritableFileStream extends js.Object {
```
DanielMoss, line 16: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream)

I'm a little lost here at the moment. This should actually extend a `WritableStream` type, and I started implementing that before realising that there's already a [`WriteableStream`](https://github.com/scala-js/scala-js-dom/blob/aca3aa2ca0b445c7bed401d784871948f9e2694e/dom/src/main/scala/org/scalajs/dom/WriteableStream.scala) (sp) defined in scala-js-dom. However what's already defined looks quite different to what I'd have expected from the whatwg/MDN spec. I came across #754 which looks related.

Ultimately I'm not sure how to proceed with the facade for `FileSystemWritableFileStream` when it comes to whether or not to integrate with the existing `WriteableStream`. I'd appreciate any thoughts or advice.

dom/src/main/scala/org/scalajs/dom/FileSystemDirectoryHandle.scala
armanbilge: - https://github.com/scala-js/scala-js/pull/4967

dom/src/main/scala/org/scalajs/dom/FileSystemWritableFileStream.scala
DanielMoss: My bad - definitely should have elaborated in the original comment.

My confusion in primarily that [the existing interface](https://github.com/scala-js/scala-js-dom/blob/53f9a1a2e56fddae60395b63f0ae313abaa3a827/dom/src/main/scala/org/scalajs/dom/WriteableStream.scala) in scala-js-dom seems to resemble the spec for [`WritableStreamDefaultWriter`](https://streams.spec.whatwg.org/#writablestreamdefaultwriter) rather than [`WritableStream`](https://streams.spec.whatwg.org/#writablestream). However the current scala-js-dom version also includes a `state` method that doesn't exist in either public interface (though `state` is an internal slot for the `WritableStream` spec, albeit with different potential values).

If we treat `WriteableStream` as a misnamed `WritableStreamDefaultWriter`, then it seems clear that an implementation of `FileSystemWritableFileStream` should not extend `WriteableStream`. However we have methods like [`CompressionStream#writable`](https://github.com/scala-js/scala-js-dom/blob/53f9a1a2e56fddae60395b63f0ae313abaa3a827/dom/src/main/scala/org/scalajs/dom/CompressionStream.scala#L18) that would then be returning `WritableStreamDefaultWriter` rather than an actual `WritableStream` [like the spec suggests](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream).

Maybe the spec has evolved over time, or I'm missing a trick with `WritableStream` = `WritableStreamDefaultWriter`.

armanbilge: Hmm, I see. That's a quite a bit more than a couple mistakes 😅 thanks for bringing that to our attention.

Since the name doesn't correspond to anything in the actual DOM maybe we should just deprecate it and start over, with the right names, `class`/`trait` designations, and APIs 🤔

DanielMoss: Thanks, glad to know I'm interpreting the code correctly. I'm still familiarising myself with a lot of this stuff. :)

Open to any ideas you guys have on how to handle this. Maybe it's best if I remove `FileSystemWritableFileStream` from this PR to simplify things, and we can move discussion of what to do about `WriteableStream` to a separate issue - maybe #629?

PR #858, opened 2024-09-22
Proposal for MathML base class

Quafadas:
* run `sbt prePR` [x]
* commit changes to `api-reports` [x]
 
This proposes a base class for MathML elements. 

If it gets merged / agreed, I'd propose to follow it up with concrete elements following the style of the SVG elements, and the documentation here;

https://developer.mozilla.org/en-US/docs/Web/MathML

Related;
- https://github.com/raquo/scala-dom-types/pulls
- https://github.com/raquo/Laminar/issues/172

cc @raquo

### DIFF
--- a/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
@@ -3,8 +3,7 @@ package org.scalajs.dom
 import scala.scalajs.js
 import scala.scalajs.js.annotation._
 
-/** The MathMLElement interface represents any MathML element.
-  */
+/** The MathMLElement interface represents any MathML element. */
 @js.native
 @JSGlobal
 abstract class MathMLElement extends Element {
@@ -21,7 +20,6 @@ abstract class MathMLElement extends Element {
   /** Corresponds to attribute xml:base on the given element. */
   var xmlbase: String = js.native
 
-
   var autofocus: Boolean = js.native
   var `class`: String = js.native
   var dir: String = js.native
@@ -33,7 +31,6 @@ abstract class MathMLElement extends Element {
   var scriptlevel: Int = js.native
   var style: String = js.native
   var tabindex: Int = js.native
-
 
   var intent: String = js.native
   var arg: String = js.native
--- a/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
@@ -21,7 +21,6 @@ abstract class MathMLElement extends Element {
   /** Corresponds to attribute xml:base on the given element. */
   var xmlbase: String = js.native
 
-  /** https://w3c.github.io/mathml-core/#global-attributes */
 
   var autofocus: Boolean = js.native
   var `class`: String = js.native
--- a/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
@@ -1,0 +1,26 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** Represents a Scala.js facade for the MathML element. Documentation for MathML elements can be found at:
+  * https://developer.mozilla.org/en-US/docs/Web/MathML/Element
+  *
+  * All MathMl elements implement this interface.
+  */
+@js.native
+@JSGlobal
+abstract class MathMLElement extends Element {
+  var onclick: js.Function1[MouseEvent, _] = js.native
+  var onmouseover: js.Function1[MouseEvent, _] = js.native
+  var onmouseout: js.Function1[MouseEvent, _] = js.native
+  var onmousedown: js.Function1[MouseEvent, _] = js.native
+  var onmouseup: js.Function1[MouseEvent, _] = js.native
+  var onmousemove: js.Function1[MouseEvent, _] = js.native
+  var onload: js.Function1[Event, _] = js.native
+  var onfocusin: js.Function1[FocusEvent, _] = js.native
+  var onfocusout: js.Function1[FocusEvent, _] = js.native
+
+  /** Corresponds to attribute xml:base on the given element. */
+  var `xmlbase`: String = js.native
+}
--- a/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
@@ -10,7 +10,7 @@ import scala.scalajs.js.annotation._
   */
 @js.native
 @JSGlobal
-abstract class MathMLElement extends Element {
+abstract class MathMLElement extends org.scalajs.dom.Element {
   var onclick: js.Function1[MouseEvent, _] = js.native
   var onmouseover: js.Function1[MouseEvent, _] = js.native
   var onmouseout: js.Function1[MouseEvent, _] = js.native
--- a/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
+++ b/dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
@@ -23,4 +23,24 @@ abstract class MathMLElement extends Element {
 
   /** Corresponds to attribute xml:base on the given element. */
   var xmlbase: String = js.native
+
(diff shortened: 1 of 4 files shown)


dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
```diff
@@ -0,0 +1,26 @@
...
+import scala.scalajs.js.annotation._
+
+/** Represents a Scala.js facade for the MathML element. Documentation for MathML elements can be found at:
+  * https://developer.mozilla.org/en-US/docs/Web/MathML/Element
+  *
+  * All MathMl elements implement this interface.
+  */
+@js.native
+@JSGlobal
+abstract class MathMLElement extends org.scalajs.dom.Element {
```
raquo, line 13: The Element type is in the same package, so you don't need the `org.scalajs.dom.` prefix.

dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
```diff
@@ -0,0 +1,26 @@
...
+  var onmouseout: js.Function1[MouseEvent, _] = js.native
+  var onmousedown: js.Function1[MouseEvent, _] = js.native
+  var onmouseup: js.Function1[MouseEvent, _] = js.native
+  var onmousemove: js.Function1[MouseEvent, _] = js.native
+  var onload: js.Function1[Event, _] = js.native
+  var onfocusin: js.Function1[FocusEvent, _] = js.native
+  var onfocusout: js.Function1[FocusEvent, _] = js.native
+
+  /** Corresponds to attribute xml:base on the given element. */
+  var `xmlbase`: String = js.native
```
raquo, line 25: Backticks are not needed for `xmlbase`.

raquo (commented):
Thanks!

A couple questions, perhaps more to the scalajs-dom maintainers:

MDN says the `style` property is defined for both [HTML](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style), [SVG](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement/style), and [MathML](https://developer.mozilla.org/en-US/docs/Web/API/MathMLElement/style) elements, but in scalajs-dom interfaces, it's only defined for HTML elements. Unless there's a known reason for that, I think it should be defined for SVG (and now MathML) elements as well.

I'm also wondering why event props like onclick / onmouseover / etc. are defined in HTML and SVG element classes in scalajs-dom – MDN does not mention anything HTML or SVG specific about most of those events ([example](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event)). I guess it's ok for MathML to follow this pattern for now, but shouldn't most of those event properties (that apply to all elements) be moved into the parent Element class?

dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
Quafadas: Oof, good point, I originally had `xml:base:, which apparently isn't needed.

Quafadas:
I've updated the two direct pieces of feedback - thankyou 🙏

Quafadas:
Just checking in - anything else I could do to nudge this forward?

zetashift:
>  but in scalajs-dom interfaces, it's only defined for HTML elements. Unless there's a known reason for that, I think it should be defined for SVG (and now MathML) elements as well.

I do not know what it has been done like that, I'm not against this change however.

As for this PR, thank you! TIL about the MathML spec.
I am wondering how faithful this implementation is to the MathML spec: https://w3c.github.io/mathml-core/#the-top-level-math-element, shouldn't this at least have the global attributes in the class: https://w3c.github.io/mathml-core/#global-attributes ?

Quafadas:
> I am wondering how faithful this implementation is to the MathML spec: https://w3c.github.io/mathml-core/#the-top-level-math-element, shouldn't this at least have the global attributes in the class: https://w3c.github.io/mathml-core/#global-attributes ?

I think you are absolutely right. I'll try to add it. I think I got the `on*` properties through sheer luck parroting the SVG element. Thanks for the pointer, BRB...

Quafadas:
@zetashift done I believe. Thanks for the pointer 🙏

dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
```diff
@@ -0,0 +1,46 @@
+package org.scalajs.dom
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+/** Represents a Scala.js facade for the MathML element. Documentation for MathML elements can be found at:
+  * https://developer.mozilla.org/en-US/docs/Web/MathML/Element
+  *
+  * All MathMl elements implement this interface.
+  */
```
zetashift, line 10:
```suggestion
/** The MathMLElement interface represents any MathML element.
  */
```

dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
```diff
@@ -0,0 +1,46 @@
...
+  var onmouseup: js.Function1[MouseEvent, _] = js.native
+  var onmousemove: js.Function1[MouseEvent, _] = js.native
+  var onload: js.Function1[Event, _] = js.native
+  var onfocusin: js.Function1[FocusEvent, _] = js.native
+  var onfocusout: js.Function1[FocusEvent, _] = js.native
+
+  /** Corresponds to attribute xml:base on the given element. */
+  var xmlbase: String = js.native
+
+  /** https://w3c.github.io/mathml-core/#global-attributes */
```
zetashift, line 27:
```suggestion
```

dom/src/main/scala/org/scalajs/dom/MathMLElement.scala
```diff
@@ -0,0 +1,46 @@
...
+  var displaystyle: Boolean = js.native
+  var mathbackground: String = js.native
+  var mathcolor: String = js.native
+  var mathsize: String = js.native
+  var nonce: String = js.native
+  var scriptlevel: Int = js.native
+  var style: String = js.native
+  var tabindex: Int = js.native
+
+  /** https://w3c.github.io/mathml-core/#other-valid-attributes */
```
zetashift, line 41:
```suggestion
```

Quafadas:
Just checking in? Anything else I can do?

zetashift (approved):

zetashift:
Sorry! This looks good to me!

Quafadas:
@raquo I assume, that I also need to add all the mathML instances, before we can take advantage of it downstream? I'm planning to do that in a separate PR to keep this one small...

raquo:
@Quafadas Well, there are levels to this.

If we just have a single `MathMLElement` class that is shared by all MathML elements in the DOM, then in Laminar we can still type all MathML elements as instances of that one class, ignoring the subclasses. It will work, but you wouldn't be able to _read_ any subclass-specific properties on such instances using `el.ref.<propName>`. You would still be able to _set_ those properties via Laminar's usual syntax though, if you define those properties in Scala DOM Types.

FTR, I don't actually know how many of such subclass-specific properties MathML elements have.

Quafadas:
@raquo Thanks for the note. 

I think what I'm proposing is that once this gets merged, I'll follow it straight up with another PR that puts up a complete list of the concrete elements. 

Based on what you've said II think my proposal would be to be quite diligent on the top level `math` element, but I would prefer to avoid a big bike shedding exercise on the correctness of the sub-elements (that are rather niche). If they turned out to have meaningful properties, it should be easier to add them at a later date...

raquo:
@Quafadas Thanks, sounds good to me, just keep in mind that you want to look at the `MathMLElement` superclass that is common to all MathML elements. The outer `<math>` element will likely have its own subclass of `MathMLElement` (not sure the exact name), so it may have properties on it that are not available on other types of MathML elements.

From what I see [on MDN](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/math), it seems that the root `<math>` element only has `display` as the additional attribute. I assume it also [reflects](https://github.com/raquo/scala-dom-types?tab=readme-ov-file#reflected-attributes) as a property, so should be present in the subclass definition, but I'm not sure.

Quafadas:
@zetashift Is there a remaining barrier to merging?

zetashift:
Thank you!

Quafadas:
@zetashift Mostly thank you in fact :-)!

Issue #296, opened 2017-08-23
Extension proposition dom.document.createElementX


MasseGuillaume:
I find myself doing this a lot:

```scala
dom.document.createElement("div").asInstanceOf[dom.raw.HTMLDivElement]
```

It would make sense to have extensions for most of the `dom.raw` package.

ritschwumm:
i have code for this lying around somewhere.

way too soon afterwards though, code like this began annoying me:
```scala
import tags._
val x = div()
x.id = "foo"
x.className = "message"
x.textContent = "hello"
```
so that preceived gain was not as big as i imagined.

Grogs:
If you import `org.scalajs.dom.html._` then there are shorter type aliases for most elements. You example would become `dom.document.createElement("div").asInstanceOf[Div]`.

I just live with that, although I'm not sure how users would know to import that package.

Grogs:
Thinking about it, in this case, it would be nice to be able to do: `document.createElement[Div]`. HTMLDivElement doesn't currently know the corresponding element name is 'div' though.

ritschwumm:
that could easily be done with a simple type class

ramnivas:
Another approach: https://blog.ramnivas.com/technology/2020/02/05/literal-types-to-simplify-code.html.

raquo:
Wow, the literal types approach is super clean:

```
def createElement(tagName: "script"): HTMLScriptElement = js.native
def createElement(tagName: "a"): HTMLLinkElement = js.native
...
```

I understand why this repo wouldn't want custom `createElementX` methods – those wouldn't match the underlying API. But it seems that literal types address this concern, the API would match the JS API, and would not require any implementation to be written, just the interfaces.

These new methods would be available in 2.13+, but maybe that's good enough?

armanbilge:
I like the idea of literal types too, let's keep this on our radar :)

PR #860, opened 2024-10-01
literal types

Quafadas:
* run `sbt prePR` - done
* commit changes to `api-reports` - done

DISCLAIMER: 

I created an object called HTML Literals i what looked like the right place, then made an AI generate the code according to the design sketched out in the PR. It matches my understanding of the ask.

It's scala 3 only, because doing it also for scala 2.13 means seperating 2.13 and 2.12, which would introduce (at least for me) a headspace problem. Scala 2 & 3 are already seperate, making that leg of the task easy - I would propose to do scala 3 only to ease the maintainer burden. 

If there is interest in merging, then I'll put my time behind the PR and try and help it over the line. 

I did not spend a lot of time here, so I shall not feel offended, if there is little appetite, either :-). I wish to help, not burden.

### DIFF
--- a/dom/src/main/scala-3/org/scalajs/dom/HtmlLiterals.scala
+++ b/dom/src/main/scala-3/org/scalajs/dom/HtmlLiterals.scala
@@ -1,7 +1,12 @@
 package org.scalajs.dom
 
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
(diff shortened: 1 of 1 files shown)


Quafadas:
https://github.com/scala-js/scala-js-dom/issues/296

PR #861, opened 2024-10-01
Update contrib

Quafadas:
Pure documentation - this was my biggest pain point getting started.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,15 @@ Contents:
 * Binary Compatibility
 * Submitting a PR
 
+Getting started
+=======
+Scala js dom currently targets java 17 for maximum compatibility. 
+
+```sh
+cs launch sbt --jvm 17
+```
+May would select a jvm that is known to work with the project
+
 
 Packages
 ========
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,6 +22,15 @@ Contents:
 * Binary Compatibility
 * Submitting a PR
 
+Getting started
+=======
+Scala js dom currently targets java 17 for maximum compatibility. 
+
+```sh
+cs launch sbt --jvm 17
```
zetashift, line 30: Is there a coursier-less way to do this? I don't think we should include it if it is not possible.

Quafadas: Hmmm... not that I know of, although there are many things I don't know :-). I would note, that that I bled some time figuring out the JVM 17 requirement, and then plenty of time futzing with env variables in shells before hitting on this as a QuickStart. 

I'm not precious on it, so will bow to superior wisdom - it would have helped me.

zetashift: > I'm not precious on it, so will bow to superior wisdom - it would have helped me.

I understand that it helps local dev, so how about a small paragraph explaining what `cs` is with a link to the coursier site?

Quafadas: That sounds quite reasonable... BRB.

PR #900, opened 2025-12-03
PerformanceEntry: Change duration type from Int to Double

nafg:
I'm getting ClassCastExceptions about a number not being convertible to a java.lang.Integer. The stack trace is unclear but I think it's coming from code that's reading the duration field of a PerformanceMeaure. And MDN says it's a DOMHighResTimeStamp which it says is a double.


sjrd (approved):

nafg:
I tried to run `sbt prePR` and I got a compiler crash. And why is CI failing?

nafg:
What's the story with this?

